PR #134, opened 2026-04-02
Variant Eq derive, regression tests, error types roadmap revision

O6lvl4:
## Summary
- Add auto-derive Eq for variant types (`type Color: Eq = | Red | Green | Blue` now generates `Color_eq`)
- Document effect fn test pattern in CHEATSHEET (use `-> Result[T, String]` for error case testing)
- Revise flexible-error-types roadmap: remove `throws` keyword, add test Result visibility as Phase 0
- Add `regression_v0_11_test.almd` with 13 regression tests covering all v0.11 fixes

## Test plan
- [x] `almide test` — 170/170 files pass
- [x] `cargo test` — all pass
- [x] Regression test covers: variant Eq, recursive generics, default fields, list patterns (literal/guard/tuple), generic structs, empty for, 3-param generics, effect fn ok/err, quicksort

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,7 +22,7 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "almide"
-version = "0.11.3"
+version = "0.11.4"
 dependencies = [
  "almide-base",
  "almide-codegen",
@@ -44,7 +44,7 @@ dependencies = [
 
 [[package]]
 name = "almide-base"
-version = "0.11.3"
+version = "0.11.4"
 dependencies = [
  "lasso",
  "serde",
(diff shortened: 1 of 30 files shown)


O6lvl4:
## Updated: v0.11.4

### New in this push
- **Test block result semantics**: Effect fn calls in test blocks now return `Result[T, String]` instead of auto-unwrapping. Use `!` for explicit unwrap or assert on `ok()`/`err()` directly.
- **Type inference**: Fix tuple destructure in lambda parameters, enable `zip_with` 3-param generic
- **Roadmap**: Move test-block-result-semantics to done

### Test results
- `almide test spec/lang/` — 90/90 (Rust), 89/89 (WASM)
- `almide test spec/stdlib/` — 58/58 (Rust), 57/57 (WASM)
- `almide test spec/integration/` — 23/23 (Rust), 22/22 (WASM)
- `cargo test` — all passed

PR #315, opened 2026-05-31
WASM engine v2 + value/json self-host cut-over

O6lvl4:
Integrates the WASM engine v2 redesign and the value/json self-hosting cut-over into develop. `json-self-host` is a superset of `wasm-engine-v2` (already merges origin/develop), so this single PR brings everything.

## WASM engine v2 (engine/)
- break/continue branch-depth resolved structurally (TCO'd tail recursion now terminates)
- `__alloc` grows linear memory (fix OOB trap past initial pages)
- lexicographic string ordering via `__str_cmp`
- self-host milestone 1: a JSON parser in Almide running on v2 (differential-gate regression)
- self-host milestone 2: `json_almide.almd` — full Almide JSON (objects/strings/escapes) on v2, 0 fallback

## Rust target
- string ordering compares via `AsRef<str>` (fix `&str` vs `String` E0308)

## value/json self-host cut-over (道A — unified recreate)
`stdlib/value.almd` + `stdlib/json.almd` are now pure Almide over a `Value` ADT — one source for both the Rust target and the WASM engine. The Codec derive emits real `value.*` module calls with lambda element encoders/decoders (so monomorphization instantiates the bundled generics); native `value.rs`/`json.rs` emptied; `sse.rs` self-contained with a private parser; recursive-ADT boxing qualified to `::std::boxed::Box`.

## Verification
- All 240 spec test files pass (incl. 17 json/value/codec + http_response)
- Engine suite 113 green
- Differential gate `scripts/wasm-v2-diff.sh`: ran-under-v2=13, 0 bugs

## Follow-ups (not in this PR)
- v2-native production json (needs v2 `string.replace`/`float.parse`/`map` intrinsics)
- finish sse migration + delete native value/json
- module-var name-shadow codegen fix


O6lvl4:
Parking this draft: the branch is now 84 ahead / **376 behind** develop, and the intervening releases (v0.26.16–v0.26.20: per-package type/ctor namespacing, the string position API codepoint unification, record-construction validation) touch the same emit_wasm/stdlib surfaces — a 12k-line rebase here would be a dedicated session of its own, not an update-branch.

The engine-v2 + value/json self-host direction itself stays active (docs/roadmap/active/wasm-engine-redesign.md); the `json-self-host` branch is preserved. When the effort resumes, re-cut a fresh PR from a rebase so review starts from a diff that reflects current develop.

PR #409, opened 2026-06-08
fix(codegen): streaming callback closure + Value AlmideRepr (#408)

### LINKED ISSUE #408: codegen: closure passed to FnMut-bound runtime fn (streaming callbacks) produces invalid Rust
## Summary

On almide v0.24.0 and v0.25.0, generating code that passes a closure to a
runtime function bound as `impl FnMut(String)` (the streaming-callback shape)
produces invalid Rust. This breaks all streaming LLM calls, including the
stdlib's own `http.anthropic_streaming_call` / `http.openai_streaming_call`,
and makes `almide-ai/almai` impossible to compile.

## Repro (stdlib, no external deps)

`src/main.almd`:

```almide
import http
import env

effect fn main() -> Unit = {
  let key = env.get("ANTHROPIC_API_KEY") ?? ""
  var acc = ""
  let body = "{\"model\":\"claude-haiku-4-5\",\"max_tokens\":64,\"messages\":[{\"role\":\"user\",\"content\":\"hi\"}]}"
  let r = http.anthropic_streaming_call(key, body, (delta) => { acc = acc + delta })!
  println(acc)
}
```

```
$ almide build src/main.almd -o build/llm
error[E0277]: expected a `FnMut(String)` closure, found `Rc<dyn Fn(String)>`
  ... required by a bound in `almide_rt_sse_anthropic_messages`
  help: you might have meant to create the closure instead of a block
        let r: String = (almide_rt_sse_anthropic_messages(&*key, &*body, |_| {
```

The generated code passes the callback as `Rc<dyn Fn(String)>` (or emits a
block instead of a closure) while the runtime helper expects `impl FnMut(String)`.

## Impact

- Non-streaming `http.request` works fine on v0.25.
- Any streaming-callback path fails to compile (native + wasm).
- `almide-ai/almai` (last updated 2026-05-01) fails to compile on v0.24/v0.25
  for this reason, plus a secondary `error[E0599]: no method named almide_repr
  found for enum Value` from `Tool` deriving `Repr` with a `Value` field.

## Environment

- almide 0.25.0 (also reproduced on 0.24.0), prebuilt macOS aarch64 release
- rustc 1.96.0 (native fallback)

atzurody:
Fixes #408.

Two codegen/runtime bugs blocked **all HTTP streaming** and **any user type
with a `Value` field deriving `Repr`** (which is why `almide-ai/almai` won't
compile on current almide).

## 1. Streaming callback closure type mismatch

The SSE runtime helpers took `on_text_delta: impl FnMut(String)`:

```rust
pub fn almide_rt_sse_anthropic_messages(api_key, body_json, mut on_text_delta: impl FnMut(String)) -> ...
```

But the uniform closure repr lowers every `(String) -> Unit` to
`Rc<dyn Fn(String) -> ()>`. So every streaming call failed with:

```
error[E0277]: expected a `FnMut(String)` closure, found `Rc<dyn Fn(String)>`
  required by a bound in `almide_rt_sse_anthropic_messages`
```

— native and wasm both. Aligned the SSE signatures with the
`Rc<dyn Fn(...) -> ...>` convention already used by `list.map` / `list.fold`
in `runtime/rs/src/list.rs`.

## 2. `Value` does not implement `AlmideRepr`

A record/variant with a `Value` field deriving `Repr` generates
`self.<field>.almide_repr()`, but `Value` had no `AlmideRepr` impl:

```
error[E0599]: no method named `almide_repr` found for enum `Value`
```

(e.g. almai's `type Tool: Repr = { …, parameters: Value }`). Added
`impl AlmideRepr for Value` in `runtime/rs/src/value.rs`, delegating to the
existing JSON stringify — consistent with how `Value` renders elsewhere.

## Verification

- Repro from #408 (`http.anthropic_streaming_call`) now builds.
- `almide-ai/almai` now compiles as a dependency.
- Spec suite green with the patched compiler: **stdlib 100 / lang 121 /
  integration 27** files, 0 failures.
- `crates/almide-codegen/src/generated/rust_runtime.rs` regenerated by
  `build.rs` from the edited `runtime/rs/src/*.rs` and committed.

(The one pre-existing `cargo test` failure, `to_fixed_cross_target_boundary`,
reproduces on a clean `develop` checkout and is unrelated to this change.)


O6lvl4:
Thanks @atzurody! 🙏 Landed in two parts:

**`sse.rs` (the #408 streaming-callback fix) — landed** on `develop` as 3c0a81e7 (rebased onto v0.26.4, your authorship preserved). Changing the SSE callback params from `impl FnMut(String)` to `Rc<dyn Fn(String)>` matches the codegen's uniform closure repr — verified by building a program that passes a *capturing* closure to `http.openai_streaming_call(...)`: it now compiles to valid Rust where it previously didn't. Full `spec/` suite green on native + wasm. Closes #408. Ships in the next release.

**`value.rs` (`impl AlmideRepr for Value`) — held back**, tracked in #428. The impl itself is correct, but on its own it doesn't fully fix a `Value` field in a `Repr`-deriving type: native string interpolation `"${v}"` and `.repr()` go through `Display` (which `Value` still doesn't implement), and the native `almide_repr` output (JSON) would diverge from the wasm repr (`Value {  }`) — a cross-target divergence with no `spec/wasm_cross/` fixture to catch it. #428 captures the complete cross-target fix (consistent `Value` repr across all native paths + wasm, byte-identical, with a fixture). Happy to have you pick that up there if you'd like!

Closing this PR since the #408 fix is in; the repr work continues in #428.

PR #410, opened 2026-06-08
fix: build -o creates output dir; don't auto-? the Result arg of result/option fns

atzurody:
Two independent papercuts that surface when building real multi-file programs
(both pass `almide check` / `almide test` but break `almide build`).

## 1. `almide build -o <dir>/<bin>` doesn't create the output directory

```
$ almide build src/main.almd -o build/app
Compile error: …                 # actually compiles fine
Failed to copy binary to build/app: No such file or directory (os error 2)
```

The compile succeeds; only the final copy fails because `build/` doesn't
exist. `-o <dir>/<bin>` is the natural way to place a binary, so every caller
otherwise needs a manual `mkdir -p`. Fix: create the output's parent dir
before copying (`src/cli/build.rs`).

## 2. auto-? unwraps the Result argument of `result.*` / `option.*` functions

The auto-? pass wraps every Result-typed call in a `Try`, including a call
passed as the **first argument** to a `result.*` / `option.*` function — which
consumes it *as* a Result (UFCS: `r.unwrap_or(d)` == `result.unwrap_or(r, d)`):

```almide
effect fn main() -> Unit = {
  let cap = result.unwrap_or(int.parse(env.get("CAP") ?? ""), 512)  // lowered to (int.parse(...))?
  ...
}
```

lowered to `result.unwrap_or((int.parse(...))?, 512)` — unwrapping the very
Result the callee needs → `error[E0308]: expected Result<i64,_>, found i64`
at `almide build` (passes check/test). It also bites effect-fn call sites
(typed `Result[T,String]`) consumed by `result.unwrap_or`, e.g.
`result.unwrap_or(run_once(), seq)`.

Fix: same shape as the existing `??` (UnwrapOr) and match-subject exceptions —
suppress the top-level auto-? on the first arg of `result.*` / `option.*`
calls (`crates/almide-frontend/src/lower/auto_try.rs`).

## Verification

- Repro (both patterns + `-o build/x` into a missing dir) builds and runs.
- Spec suite green with the patch: **stdlib 100 / lang 121 / integration 27**, 0 failures.

Found while building the Fizz component repos (multi-package CLIs) on v0.25.

### DIFF
--- a/crates/almide-frontend/src/lower/auto_try.rs
+++ b/crates/almide-frontend/src/lower/auto_try.rs
@@ -236,10 +236,24 @@ fn insert_try(expr: IrExpr, in_match_subject: bool) -> IrExpr {
             op,
             operand: Box::new(insert_try(*operand, false)),
         },
-        IrExprKind::Call { target, args, type_args } => IrExprKind::Call {
-            target,
-            args: args.into_iter().map(|a| insert_try(a, false)).collect(),
-            type_args,
+        IrExprKind::Call { target, args, type_args } => {
+            // `result.*` / `option.*` functions consume their FIRST argument AS a
+            // Result/Option (UFCS: `r.unwrap_or(d)` == `result.unwrap_or(r, d)`).
+            // That arg must NOT be auto-?'d, exactly like the `??` (UnwrapOr) and
+            // match-subject exceptions — otherwise `result.unwrap_or(int.parse(s), d)`
+            // becomes `result.unwrap_or((int.parse(s))?, d)`, unwrapping the Result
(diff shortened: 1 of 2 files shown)


O6lvl4:
Landed on `develop` as 854c0862 (cherry-picked and rebased onto v0.26.4) — thanks @atzurody! 🙏

Both changes verified locally: `result.unwrap_or(int.parse("x"), 42)` correctly yields `42` (the `Result`/`Option` first argument is no longer auto-`?`'d), and `build -o build/app` now creates the parent directory. Full `almide test spec/` suite green on native + wasm, no regression. Will go out in the next release.

PR #510, opened 2026-06-11
True Perceus Stage C: per-iteration TCO reclamation (2M-deep recursion 4.27GB to 7.3MB)

O6lvl4:
Stage C of the true-Perceus roadmap: TCO self-call loops now reclaim heap per iteration. The headline: a 2M-deep constant-size TCO recursion went from **4.27 GB to 7.3 MB flat**.

## What landed

- **The parked M2 re-applied** (loop-exit flattening + managed-param entry-Inc / per-iteration Dec). Its old free-list corruption was never M2's fault — the per-iteration loop REGION reset rolled back the bump pointer without clearing the free list (fixed in the v0.27.0 flip); M2's Decs were simply the first feeder of that list inside region loops. The churn gate built for exactly this decision confirmed the re-land in minutes.
- **Call-valued self-call args are adoption-safe** — `is_fresh_alloc` extended beyond literals: user fns return OWNED (mechanism #6) and alias-shaped temps receive the bind-level Inc once the TCO rewrite turns args into VDecls. The literal-only test left `spin(n-1, string.take(acc + "x", 8))` unmanaged (unbounded growth).
- **The inner-hoist alias-Inc is now gated on the tail temp's own value aliasing** — a fresh tail moved out of its block owns itself; the unconditional +1 was a measured 16 B/iteration leak.
- **A real classifier hole found by that refinement** (the predicted load-bearing-over-inc class): `?? default` can reach Perceus as `Call { option.unwrap_or }`, which the alias classifier (RuntimeCall-spelling only) treated as FRESH — removing the masking +1 exposed a map-owned payload double-free at teardown, caught by the byte gate within minutes. The Call{Module} spelling of `unwrap_or`/`get_or` is now alias-classified.

## Gates

Full bar green: native corpus 264/264 · wasm default 264/264 · wasm opt-out 264/264 · byte gate 67/67 · churn gate (record 2M, TCO-loop 200k, NEW `tco_deep_recursion_churn` 2M-deep) · cargo full suite 0 failures.

## Documented limit (not a bug)

A monotonically GROWING accumulator still peaks at the sum of intermediates: every freed block is one unit smaller than the next request, so first-fit can never reuse it. That is an allocator-policy frontier (capacity growth / size classes), recorded in the roadmap; C-066's bounded-exception list shrinks to the construct-from-temp over-count outside reclaimed regions.

### DIFF
--- a/crates/almide-codegen/src/pass_perceus.rs
+++ b/crates/almide-codegen/src/pass_perceus.rs
@@ -405,6 +405,18 @@ fn insert_decs_before_ret(fb: FnBody, heap_vars: &[VarId], ret_vars: &HashSet<Va
                     for var in vars_to_dec.iter().rev() {
                         res = FnBody::Dec { var: *var, body: Box::new(res) };
                     }
+                    // Rule 1 applies HERE too: this hand-built VDecl bypasses
+                    // the ChainHead::VDecl arm, so a ret expr that ALIASES one
+                    // of the locals being Dec'd below (e.g. `if filled then
+                    // base + "…" else base` with `Dec base` following) needs
+                    // its own reference or the function returns a freed
+                    // pointer. This under-count was masked for years by the
+                    // chain-temp over-incs the alias-gated hoist removed
+                    // (caught by the byte gate as a resurrection trap,
+                    // default_fields_test).
+                    if is_heap_type(&ret_ty) && yields_borrowed_alias(&expr) {
+                        res = FnBody::Inc { var: ret_var, body: Box::new(res) };
+                    }
                     FnBody::VDecl { var: ret_var, ty: ret_ty, mutability: Mutability::Let, expr, body: Box::new(res) }
                 } else {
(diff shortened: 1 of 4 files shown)


O6lvl4:
CI's Test WASM caught a second masked under-count the alias-gated hoist exposed: the Ret-arm tail-lift hand-builds its `__perceus_ret` VDecl, bypassing the ChainHead::VDecl alias rule — a ret expr aliasing a to-be-Dec'd local (`if filled then base + \"…\" else base` with `Dec base` below) returned a freed pointer, surfaced as a resurrection trap in default_fields_test. Rule 1 now applies at the lift too. Lesson burned in: the local bar now runs `almide test spec/ --target wasm` EXPLICITLY (the default mode silently falls back to native for some files — this one fell back and passed). Full bar re-run green: native 264, wasm-explicit 254/0/10skip, byte gate 67/67, churn ×3 (deep 7.3MB), cargo 0.

PR #614, opened 2026-06-11
Dispatch cross-module Codec on wasm + surface wasm-only test failures (completeness mechanism)

### LINKED ISSUE #609: Cross-module derived Codec method ICEs on wasm (native ok) and almide test masks it via native fallback
**Danger class: wasm compile ICE + TEST MASKING — hole-hunt round 4, Lens K. Adversarially verified; latent in the shipped spec corpus.**

A cross-module derived Codec method (`dep.Type.encode/decode`) works on NATIVE but ICEs on WASM at compile time. The wasm call dispatch reconstructs the runtime fn name as `almide_rt_{TYPENAME}_{method}` instead of `almide_rt_{OWNING-MODULE}_{method}`.

Repro (the shipped spec dep): copy `spec/integration/modules/varlib/src/mod.almd` as a dep `varlib`; main calls `varlib.Pigment.decode(varlib.Pigment.encode(p))` — the EXACT call in `spec/integration/modules/cross_module_codec_test.almd:19`.
- native: prints `3`, exit 0.
- wasm: `[ICE] emit_wasm: no WASM dispatch for \`Pigment.decode\`` (`emit_wasm/calls.rs:732`), exit 101.

Root: `emit_wasm/calls.rs:711-737` builds the name from the TYPE name (`Pigment`), but the fn was registered as `almide_rt_varlib_Pigment_decode` (owning module `varlib`) — native `--target rust` emit confirms the correct name. The bare/qualified fallbacks also miss → panic. Intra-module Codec on wasm is FINE.

**Test-masking (separate, important):** running `almide test` on `cross_module_codec_test.almd` shows the wasm path DOES ICE in stderr, but the native fallback swallows it and reports "All 1 test file(s) passed" exit 0 — so the bug is reachable from the standard test command yet hidden (same family as the documented wasm-test-false-pass trap). `tests/crossmod_matrix_test.rs` (the both-target gate) has ZERO Codec cell.

Fix: dispatch by the type's OWNING module, not the type name; add a Codec cell to `crossmod_matrix_test.rs` as the regression lock; and make `almide test` not let a native pass mask a wasm build ICE.

O6lvl4:
Closes #609 — cross-module derived Codec ICE'd on wasm AND `almide test` masked it. This PR fixes the bug and installs the COMPLETENESS-MAINTAINING MECHANISM that makes the whole class detectable (per the directive: every hit ships the mechanism that keeps the class detectable, not just the instance).

## The bug (#609)

A derived Codec method on a type in a DEPENDENCY module — `dep.Type.encode/decode`, the exact call in `spec/integration/modules/cross_module_codec_test.almd:19` — works on native but ICE'd the wasm emitter: the call arrives with `module` = the TYPE name (`Pigment`), so the dispatch built `almide_rt_Pigment_decode` and missed the registered `almide_rt_varlib_Pigment_decode`. The native resolution (`MODULE_METHOD_FNS`) is a Rust-only pass.

Two wasm dispatch fixes:
1. The scalar `Type.method`: when the prefix lookup misses, find the unique `func_map` key ending in `_{Type}_{method}` under the `almide_rt_` namespace.
2. The synthesized list helper `__decode_list_varlib.Pigment` (for a `List[dep.Type]` field) was SPLIT on the dot into module/func; reassemble the full name and route it to the value-codec dispatch (the Named arm already did this for same-module types).

## The mechanism (the part the directive asked for)

The bug survived because **`almide test` hid the wasm ICE behind the native fallback** — `compile_and_run_wasm_test` turned a WASM CODEGEN PANIC into a `Skip` (treated like a benign `// wasm:skip`), so a wasm-only crash on a type-checking program reported PASS.

- A wasm codegen PANIC is now a `Fail` (it carries the panic message) — a panic on a program that type-checks is a compiler bug, not a skip. (The author had already learned this for parse errors; codegen panics were the remaining hole.)
- `cmd_test_fast` now tracks wasm `Fail`s distinctly from `Skip`s and reports any file that FAILED on wasm but PASSED on native as a `WASM-FAILED (cross-target bug)` — exit 1. A genuine `// wasm:skip` is unaffected.

Running the full corpus with the detector surfaced exactly one previously-hidden bug (this one); after the fix, `0 wasm-only-failed`, and `spec/` wasm-skips drop 10 → 9. A `cross_module_codec_roundtrip` cell is added to `tests/crossmod_matrix_test.rs` (the both-target gate) as the permanent regression lock — covering the scalar method AND the `__decode_list` helper.

## Gates

native 268/268 · wasm explicit 259/0/9-skip · default `almide test` 0 wasm-only-failed · byte gate 67/67 · crossmod matrix incl. new Codec cell · cargo full 0 failures.

### DIFF
--- a/crates/almide-codegen/src/emit_wasm/calls.rs
+++ b/crates/almide-codegen/src/emit_wasm/calls.rs
@@ -709,6 +709,18 @@ impl FuncCompiler<'_> {
                         self.emit_call(&target, args, _ret_ty);
                     }
                     _ => {
+                        // #609: a synthesized codec helper for a DEPENDENCY-module
+                        // element type — `__decode_list_varlib.Pigment` — gets
+                        // SPLIT on the dot into module=`__decode_list_varlib`,
+                        // func=`Pigment`. Reassemble and route to the value-codec
+                        // dispatch (the Named arm does this for same-module types).
+                        let full = format!("{}.{}", module.as_str(), func.as_str());
+                        if full.starts_with("__encode_option_") || full.starts_with("__decode_option_")
+                            || full.starts_with("__decode_default_") || full.starts_with("__encode_list_")
+                            || full.starts_with("__decode_list_") {
+                            self.emit_codec_helper(&full, args);
+                            return;
+                        }
                         // Try user module function: almide_rt_{module}_{func}
                         let mod_ident = module.as_str().replace('.', "_");
                         let func_ident = func.as_str().replace('.', "_");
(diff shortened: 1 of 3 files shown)


O6lvl4:
Closing as superseded + premature.

**Codec fix (the main substance) already shipped.** #609 is closed and develop's `emit_wasm/calls.rs` already carries the equivalent cross-module Codec dispatch — `emit_codec_helper`, the `__decode_list_`/`__encode_option_`… prefix routing, and the `Cross-module derived Codec method (#609)` handling. That's why this branch is now CONFLICTING against develop.

**The remaining novel part — panic→Fail instead of Skip for a WASM codegen crash — is good but premature.** develop still treats a wasm codegen panic as a Skip (`commands.rs`), and flipping it to Fail now would redden CI on the wasm codegen panics already in the corpus (the nightly fuzz surfaced ~478 native↔wasm divergences, triaged in #727). It should land as its own small ratchet once that burndown clears the wasm panic class, not bundled with an already-shipped codec fix on an 11-day-stale base.

No action lost: the codec bug is fixed on develop; the harness-strictness idea is captured here and tied to #727.

Issue #611, opened 2026-06-11
type X: Codec with a non-Codec Named field passes the checker, then native E0425 / wasm ICE (missing derivability gate)


O6lvl4:
**Danger class: missing checker gate (checker accepts, both backends fail) — hole-hunt round 4, Lens K Finding 2. Adversarially verified.**

A `Codec`-deriving type with a field whose Named type does NOT derive Codec passes the checker, then crashes codegen on BOTH targets with divergent failure modes.

Repro (single file) — `/tmp/lensk/f2_single.almd`:
```almide
type Inner = { a: Int }          // no Codec
type Outer: Codec = { inner: Inner }
```
- `almide check`: "No errors found".
- native: E0425 `cannot find function \`Inner_encode\``.
- wasm: `[ICE] no WASM dispatch for \`Inner.decode\`` (`emit_wasm/calls.rs:732`).

Root: `lower/derive_codec.rs:107-117` emits `Type.encode(field)` unconditionally for any `Ty::Named` field (and the symmetric decode) with NO derivability check; no field-derivability gate exists in `check/`. Correct behavior = a frontend E-error ("field `inner: Inner` requires `Inner` to derive Codec"), not an E0425/ICE. Distinct from the cross-module-Codec dispatch bug (filed separately): that is a VALID program broken only by wasm dispatch; this is an INVALID program the checker should reject.

(Round-4 Lens L additionally claimed Repr-on-variants missing, Codec.decode-variant-stub, and sort+Ord issues — the adversarial verifier REFUTED Lens L's lead findings as stale-binary artifacts, so those are NOT filed; this Lens-K field-derivability gate is the verified one.)

Issue #608, opened 2026-06-11
 unwrap in a non-effect fn is checker-accepted then mis-lowered (wasm silent-wrong Some(ptr) / native E0277)


O6lvl4:
**Danger class: SILENT WRONG on wasm + native ICE — hole-hunt round 4, Lens J. Adversarially verified.**

The `!` (unwrap) operator used in a NON-effect, NON-test fn whose return type is not Result/Option is accepted by `almide check` but mis-lowered by BOTH backends. The checker's Unwrap rule (`check/infer.rs:890-905`) only type-checks the inner Option/Result; it never enforces the documented "effect fn only" contract (CHEATSHEET.md:273) — it ignores the effect-context flags already in `self.env` (`mod.rs:577-578`) and the enclosing return type.

Repro (silent wrong on wasm) — `/tmp/lensJ/ret_opt.almd`:
```almide
fn first(xs: List[Int]) -> Option[Int] = Some(list.get(xs, 0)!)
fn main() -> Unit = {
  match first([7,8,9]) { Some(v) => println("some " + int.to_string(v)), None => println("none") }
  match first([])      { Some(v) => println("some " + int.to_string(v)), None => println("none") }
}
```
- `almide check`: "No errors found".
- native `almide run`: REJECTED — generated-Rust E0277.
- wasm: builds, exit 0, prints `some 7` then `some 17596481011713` — the empty-list case must print `none`; wasm `return_`'s an err-pointer the caller mis-reads as `Some(<ptr>)`.

Also (wasm [COMPILER BUG]): `fn first(xs) -> Int = list.get(xs,0)!` → check ok, native E0277, wasm "structural validation: expected i64, found i32". Control: the SAME ops inside `effect fn` have perfect parity (both abort `Error: none` exit 1). Distinct from closed #485 (implicit auto-unwrap-on-reassignment in a valid effect context). Fix: `check/infer.rs:890` should reject `!` when NOT in effect context AND the enclosing return type is not Result/Option — a clean check error, identical on both targets.

Issue #489, opened 2026-06-10
check/lower: lambda bodies skip the effect auto-? rule (checker clean, rustc E0369)


O6lvl4:
## Summary

Inside a lambda in an effect fn, `let v = step(n)` (step = effect fn) passes the checker but the generated Rust fails (E0369/E0308): the checker keeps `auto_unwrap` enabled inside lambda bodies, but `auto_try` never descends into `Lambda` bodies, so no `?` is inserted.

## Repro

```almide
effect fn step(n: Int) -> Int = n + 1
effect fn main() -> Unit = {
  let f = (n: Int) => {
    let v = step(n)
    v + 1
  }
  println(int.to_string(f(1)))
}
```

`almide check`: clean. Native build: rustc error on `v + 1` (v is Result).

## Root cause

- checker: lambda inference saves/restores only `current_ret` + `lambda_depth` (`infer.rs:769-785`) — `auto_unwrap` stays true inside the lambda, so the binding unwraps in the checker's view.
- lowering: `auto_try.rs` passes Lambda bodies through untouched (`IrExprKind::Lambda` arm returns params/body unchanged).

The two phases disagree; either both should apply the rule inside lambdas (and define what `?` propagates TO — the lambda would need a Result return), or the checker should disable `auto_unwrap` inside lambda bodies and report a proper E-code telling the user to handle the Result. Shared gap for let AND assign — fix both together (do not re-diverge the binding positions unified in #485).

PR #616, opened 2026-06-12
Reject bang unwrap outside an error-propagating context (E022) (#608)

O6lvl4:
### Root cause
`expr!` propagates the unwrapped error: lowering renders it as `?` (effect fn body) or `.unwrap()` (test block). The checker's `Unwrap` rule only required the operand to be `Option`/`Result` — it never checked whether the **enclosing context can propagate the error**. So `int.parse(s)!` in a pure `fn -> Int` type-checked ("No errors found") and then emitted `(int.parse(s))?` inside a function that does not return `Result`:

```
error: codegen produced invalid Rust — this is an Almide bug.
```

(native ICE; wasm build failure). Pure checker-accepts-but-codegen-rejects soundness hole. The same class also hit `!` used **inside a lambda** within an effect fn — `?` can't propagate out of the closure (#489), and that too produced invalid Rust.

### Fix
A single propagation-context gate (`check_unwrap_propagation_context`), called from both the direct and the piped `Unwrap` arms. `!` is valid exactly where error propagation is possible:
- `auto_unwrap` live — an `effect fn` body, outside any lambda, AND
- inside a `test` block (`in_test_block`).

Everywhere else → **E022** with a context-aware hint (lambda vs. mark-effect-fn).

### Completeness mechanism (kill the class, not the instance)
- The gate itself closes the whole class structurally: it tracks *propagation possibility*, not the one `int.parse` shape, so every fallible call in a non-propagating context is caught.
- **Diagnostic fixture** `tests/diagnostics/e022-bang-outside-effect/` — `broken.almd` asserts the E022 code/message/hint; `fixed.almd` (the `effect fn` form **and** the `??` form) is compiled clean by the harness. The harness pins *both* "the error fires" and "the suggested fix actually compiles" — the latter is exactly the meta-gap this class lived in.
- The **hard** coverage gate (`diagnostic_coverage_test.rs`, `SOFT_GATE=false`) forced this PR to also ship `docs/diagnostics/E022.md`; a new `with_code` without its fixture + doc now fails CI.

### Verification
- native `spec/`: 268/268 · wasm `spec/`: 258 passed, 0 failed · detector: 268/268
- `cargo test --release`: 0 FAILED suites (diagnostic harness 4/4, coverage 2/2)
- the heavy `!` user `spec/lang/result_option_matrix_test.almd` still passes (all uses are effect-fn / test)

### Sibling
Same meta-class (checker accepts → codegen ICE) as the Codec-derive field gate (#611), tracked separately — different root cause (derive field-type derivability), shipped on its own branch.

### DIFF
--- a/crates/almide-frontend/src/check/infer.rs
+++ b/crates/almide-frontend/src/check/infer.rs
@@ -890,6 +890,7 @@ impl Checker {
             ExprKind::Unwrap { expr: inner, .. } => {
                 let t = self.infer_expr(inner);
                 let resolved = resolve_ty(&t, &self.uf);
+                self.check_unwrap_propagation_context();
                 if let Some(inner_ty) = resolved.option_inner().or_else(|| resolved.result_ok_ty()) {
                     inner_ty
                 } else if matches!(&resolved, Ty::Unknown | Ty::TypeVar(_)) {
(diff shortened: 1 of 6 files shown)


O6lvl4:
Superseded by https://github.com/almide/almide/pull/619 (rebased onto develop to resolve the llms.txt E022/E023 conflict after #618 merged). Same change, conflict-free.

Issue #621, opened 2026-06-12
Phantom/unused generic type param emits invalid Rust natively (no PhantomData); wasm correct


O6lvl4:
> Found by the round-5 multi-lens hole-hunt (12 lenses, adversarial-verified; all 18 candidates upheld, 0 dismissed). Binary: almide 0.27.4 (develop + #607/#608/#611).

**Repro:**
```almide
type Tagged[T] = { raw: Int }
fn main() -> Unit = {
  let a: Tagged[String] = { raw: 7 }
  println("${int.to_string(a.raw)}")
}
```

- **native:** check: "No errors found" (exit 0). run: "codegen produced invalid Rust — this is an Almide bug" then rustc error[E0392]: type parameter `T` is never used (generated `pub struct Tagged<T: Clone + PartialEq> { raw: i64 }` with no PhantomData<T>), plus two error[E0283]: type annotations needed (`#[derive(...)]` on the struct, and `Tagged { raw: n }` cannot infer T). exit 1.
- **wasm:** check passes; `build --target wasm` succeeds (8722→798 bytes); wasmtime prints `7` (exit 0).
- **expected:** 7  (Tagged is just a tagged Int wrapper; the phantom param T should be dropped or carried as PhantomData). wasm is correct; native must either inject PhantomData<T> into the struct or erase the unused param.

<details><summary>verification notes</summary>

CONFIRMED (adversarial verification upheld the candidate). Binary asserted almide 0.27.4. macOS has no `timeout`; used a perl alarm-based wrapper to enforce the 15s cap.

Evidence:
- /tmp/h5_generics_adv1.almd (original candidate, with `tag[T]` helper): native E0392/E0283 (exit 1), wasm prints `7` (exit 0). Reproduces verbatim.
- /tmp/h5_generics_adv2.almd (MINIMAL — helper fn removed, just a phantom-param type + annotated record literal): native E0392 + 2×E0283 (invalid Rust), wasm prints `7`. This is the reduced repro reported above.
- /tmp/h5_generics_adv3.almd (CONTROL — T actually used in a `used: T` field): BOTH targets print `7 hi`, exit 0. Proves the defect is specifically the UNUSED/phantom generic param, not generics-with-records in general.

[...]

PR #637, opened 2026-06-12
Register anonymous-record variant payload types from declarations (#628)

### LINKED ISSUE #628: Native invalid Rust (E0425) when 2+ variants carry anonymous-record payloads with different field-name sets; wasm fine
> Found by the round-5 multi-lens hole-hunt (12 lenses, adversarial-verified; all 18 candidates upheld, 0 dismissed). Binary: almide 0.27.4 (develop + #607/#608/#611).

**Repro:**
```almide
type Shape = | Circle({ r: Int }) | Square({ s: Int })
fn main() -> Unit = {
  let s = Circle({ r: 9 })
  println("${s}")
}
```

- **native:** Compile error: codegen produced invalid Rust — this is an Almide bug. error[E0425]: cannot find type `s` in this scope / 3144 | Square(s) (for the minimal repro). Original report's repro yields the same class: error[E0425]: cannot find type `h_w` in this scope / 3144 | Box(h_w). Native exit 1.
- **wasm:** Built successfully (8844 -> 1189 bytes), runs and prints: Circle({ r: 9 }) (exit 0). `almide check` also passes ("No errors found").
- **expected:** `almide check` reports 'No errors found' and wasm builds + runs (prints `Circle({ r: 9 })`), so the program is well-typed. Native codegen is wrong: it concatenates the anonymous-record payload's field names (alphabetically: h, w) into `h_w` and emits `Box(h_w)` treating that as a type, yielding rust

<details><summary>verification notes</summary>

CONFIRMED, with a sharper trigger than the report claimed. `check` passes and wasm builds+runs (prints "Circle({ r: 9 })"), but native codegen emits invalid Rust: the anonymous-record payload's field name(s) are concatenated (alphabetically) into an identifier that is then used as if it were the variant's payload TYPE, e.g. Box payload {w,h} -> `Box(h_w)`, Square payload {s} -> `Square(s)`. That identifier is not a defined type -> rustc E0425. This is a well-typed program (idiomatic Almide: variants leading |, anonymous-record payloads) so it is a genuine native ICE, not user error.

REFINED TRIGGER (report said ">=2 variants whose payloads are anonymous records, multi-field required for h_w" — partly wrong): the real trigger is 2+ variants carrying anonymous-record payloads whose FIELD-NAME SETS DIFFER. Evidence from probes:
 - Single anon-record variant (type Shape = | Box({w,h})): compiles fine native, prints Box({h:4,w:3}). (report agreed)
 - B: Circle({r}) | Square({s}) -> FAILS `Square(s)` E0425. So single-field records suffice; multi-field NOT required (contradicts report's "alphabetically h,w into h_w" being the essence).
 - F: Circle({r}) | Ring({r}) (SAME field name) -> WORKS on BOTH targets (prints Circle({ r: 9 })). Structurally-identical anon records do not trigger it.
 - C/D: Circle({r}) | Box(Int) (one record + one non-record) -> WORKS native. So a non-record sibling variant is fine; it needs >=2 record-bearing variants with differing shapes.
 - E: Circle({r}) 

</details>

O6lvl4:
## #628 — variant case with an anon-record payload of a distinct field set → native E0425

`type Shape = | Circle({ r: Int }) | Square({ s: Int })`. `Circle({ r })` is constructed so its anon record registers; `Square`'s `{ s }` is reachable only from the type declaration, so the collector never saw it. `render_type`'s fallback then emitted the bare field name as a type:

```rust
pub enum Shape { Circle(AlmdRec_r<i64>), Square(s) }   // error[E0425]: cannot find type `s`
```

wasm (type-erased) was fine → a cross-target divergence (round-5 generics cluster).

### Fix
`collect_anon_records` now also scans the field / variant-payload types of every type **declaration** (`collect_anon_from_type_decl`), so an anon-record payload that is never constructed still registers and lowers to its `AlmdRec_<fields>` struct.

### Completeness mechanism
`spec/wasm_cross/anon_record_variant_payloads.almd` (contract **C-079**) — three cases with distinct anon-record payloads (one never matched destructively), constructed + area-matched + repr'd, byte-identical native == wasm.

### Verification
- repro: `Square(AlmdRec_s<i64>)` now emitted (was `Square(s)`); native == wasm `Circle({ r: 9 })`
- fixture: native == wasm `12 / 25 / 12 / Square({ s: 9 })`
- native `spec/` 269/269 · wasm 259/0/10

> Stacks on #636 (#621, C-078); rebase after that merges so C-079 is contiguous.

Closes #628

### DIFF
--- a/crates/almide-codegen/src/walker/declarations.rs
+++ b/crates/almide-codegen/src/walker/declarations.rs
@@ -340,6 +340,15 @@ pub fn collect_anon_records(program: &IrProgram, named: &HashMap<Vec<String>, St
     let named_set: HashSet<Vec<String>> = named.keys().cloned().collect();
     let mut seen: HashSet<Vec<String>> = HashSet::new();
 
+    // Collect from TYPE DECLARATIONS — a variant case payload or record field
+    // whose type is an anonymous record (e.g. `| Square({ s: Int })`) may NEVER
+    // be constructed in the program, so it is reachable only here; without this
+    // its struct goes unregistered and `render_type` falls back to emitting the
+    // bare field name as a type → `Square(s)` → rustc E0425 (#628).
+    for td in program.type_decls.iter().chain(program.modules.iter().flat_map(|m| m.type_decls.iter())) {
+        collect_anon_from_type_decl(td, &named_set, &mut seen);
+    }
+
     // Collect from all types AND expressions in the program
     for func in &program.functions {
         for p in &func.params { collect_anon_from_ty(&p.ty, &named_set, &mut seen); }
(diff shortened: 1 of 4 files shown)


O6lvl4:
Superseded by the rebased branch (onto develop with C-078) so the contract is contiguous.

PR #724, opened 2026-06-17
Skip the copy-on-write clone for a uniquely-owned object on wasm (fix #696, fix #695)

### LINKED ISSUE #695: wasm perf: building a List[String] of freshly-allocated strings is O(n²) — push/map/set all quadratic; List[Int] is fine
**On the wasm target, accumulating heap-allocated strings into a List is O(n²). At 152k elements (a GGUF tokenizer vocab) this turns a <2s native load into >5 minutes. Native is unaffected. Found building the nn browser-tokenizer (M2 of the WebGPU port).**

## Measured scaling (node 22 WASI, v0.27.7 / develop 058f33ff)

Inner body `let s = string.from_bytes(bytes.to_list(bytes.slice(data,8,12))); list.push(out, s)`:

| n | time |
|---|---|
| 50,000 | 0.01 s |
| 100,000 | 2.63 s |
| 150,000 | 18.49 s |

100k→150k is 1.5× the work for 7× the time — quadratic. `list.range(0,n) |> list.map(...)` producing strings is identical (50k 0.01s / 100k 2.45s / 150k 17.12s). `list.repeat("",n)` + `list.set(out,i,s)` is even worse (50k = 30s, then `unreachable` trap).

## The smoking gun — element TYPE decides it

Same push loop, three element sources:

```almide
// (a) List[Int], push integers           — 150k = 0.00s   O(1) amortized ✓
var out: List[Int] = []; while i<n { list.push(out, i); i=i+1 }

// (b) List[String] from a LITERAL list    — 150k = 0.02s   O(1) amortized ✓
var out: List[String] = []; while i<n { list.push(out, string.from_bytes([71,71,85,70])); i=i+1 }

// (c) List[String] from to_list(slice())  — 150k = 18.5s   O(n²) ✗
var out: List[String] = []; while i<n { list.push(out, string.from_bytes(bytes.to_list(bytes.slice(data,8,12)))); i=i+1 }
```

`bytes.slice`, `bytes.to_list`, `string.from_bytes` in isolation (no push, or pushing-but-discarding) are all O(1)/fast at 300k. Only **(heap string) × (push into a growing List)** is quadratic, and a *literal* string in the same position is fine.

## Likely mechanism

The List grow/realloc deep-clones its elements, and for non-Copy element types (String, and presumably any heap type / List-of-List) the clone is O(content) AND the backing array appears to grow by a constant rather than doubling — so n pushes × O(n) copy = O(n²). Literal strings dodge it (static/interned backing → shallow copy); `List[Int]` dodges it (Copy elements, real doubling). This smells like the value-semantics/RcCow element-clone-on-grow path on the wasm emitter not amortizing for heap elements. Native uses Vec growth + Rc, so it's O(n) there.

## Impact

Blocks any wasm program that builds a large `List[String]` (or List of heap values) element-by-element — parsers, tokenizers, CSV/JSON readers. Concretely it's the last blocker for the nn in-browser tokenizer: the encode/decode logic is already verified correct on wasm (20-prompt pretokenize parity + all unit tests green after the #643 fix); only the 152k-vocab *load* is unusable.

## Workarounds I tried (all fail)

`list.set` into a preallocated `list.repeat` → worse (value-semantic copy-on-set). `list.map` → same O(n²). The only path that stays O(n) is to avoid `List[String]` entirely: store the vocab as one packed `Bytes` blob + a `List[Int]` offset table (List[Int] builds fine) and binary-search byte ranges. I'll likely refactor the tokenizer that way regardless, but the underlying List-of-heap-elements quadratic is worth fixing for everyone.

### LINKED ISSUE #696: wasm: bottom-up merge sort of List[Int] traps "memory access out of bounds" at n≳8000 (native fine at 152k)
**A textbook bottom-up merge sort over `List[Int]` runs correctly on native at 152k elements but traps with a raw `memory access out of bounds` on the wasm target once n is ≳ 8000. Found building the nn browser tokenizer (the vocab index sort).**

## Minimal repro (single file, no imports)

```almide
fn msort(xs: List[Int]) -> List[Int] = {
  let n = list.len(xs)
  var cur = list.slice(xs, 0, n)
  var width = 1
  while width < n {
    var merged: List[Int] = []
    var lo = 0
    while lo < n {
      let mid = if lo + width < n then lo + width else n
      let hi = if lo + 2 * width < n then lo + 2 * width else n
      var i = lo
      var j = mid
      while i < mid and j < hi {
        if (list.get(cur, i) ?? 0) <= (list.get(cur, j) ?? 0) then { list.push(merged, list.get(cur, i) ?? 0); i = i + 1 }
        else { list.push(merged, list.get(cur, j) ?? 0); j = j + 1 }
      }
      while i < mid { list.push(merged, list.get(cur, i) ?? 0); i = i + 1 }
      while j < hi { list.push(merged, list.get(cur, j) ?? 0); j = j + 1 }
      lo = lo + 2 * width
    }
    cur = merged
    width = width * 2
  }
  cur
}
fn sort_first(n: Int) -> Int = {
  var xs: List[Int] = []
  var i = 0
  while i < n { list.push(xs, n - i); i = i + 1 }   // [n, n-1, ..., 1]
  list.get(msort(xs), 0) ?? -1                       // expect 1
}
```

`almide build --target wasm`, then call `sort_first(n)`:

| n | wasm | native |
|---|---|---|
| 5000 | `1` ✓ | `1` ✓ |
| 8000 | **`memory access out of bounds`** | `1` ✓ |
| 10000 | **trap** | `1` ✓ |
| 152000 | trap | `1` ✓ (nn tokenizer uses this — works on native) |

The algorithm is correct (sorts at n≤5000 on wasm, and at every n on native). It's purely a wasm codegen/runtime fault at scale.

## Notes that may localize it

- The threshold (~5000–8000) sits near a 64 KiB wasm page boundary (8000 × 8 B = 64 KB), which smells like the bump allocator's `memory.grow` not firing (or a pointer computed past the current page) for the allocation pattern here.
- It is NOT plain list growth: a flat `while i<n { list.push(out, i) }` at 150k is instant and fine; `list.slice` and read-via-`list.get`+`push` at 100k are both fine. The fault needs the merge structure — repeated `cur = merged` (a push-built list) reassignment across the outer loop, then reading `cur` via `list.get` while building the next `merged`.
- Forcing an owned copy (`cur = list.slice(merged, 0, list.len(merged))`) does NOT help — still traps at 10k.
- `list.get`/`list.push` are bounds-checked (Option / grow), so a *raw* OOB cannot originate from the Almide indices — it must be in the emitted list/alloc code.

## Impact

Blocks sorting any large array on wasm — and specifically the nn in-browser tokenizer's 152k-vocab index sort. (Same area as #695: the wasm List backend struggles at scale. #695 is List[String] build O(n²); this is List[Int] sort OOB.) The tokenizer's encode/decode logic itself is verified correct on wasm; only the load-time sort is blocked.

O6lvl4:
## Problem (#696, #695)

A bottom-up merge sort over `List[Int]` ran correctly on native at 152k but trapped on wasm with a raw `memory access out of bounds` once n ≳ 8000 (#696). Separately, building a `List[String]` by repeated `push` was O(n²) on wasm (#695).

## Root cause

`__cow_check` — the copy-on-write guard emitted at the mutation sites of `AliasCowPass`-marked vars — **cloned the whole heap object unconditionally**. That was correct under the pre-0.27.0 bump-allocate-and-leak model, where the refcount was a no-op so a `rc>1` test could never fire — but **frees (a real refcount) became the default in 0.27.0** and the guard was never updated.

So a `needs_cow` list mutated in a loop — `while … { list.push(merged, …) }` where `cur = merged` aliases it — was cloned in **full on every push**: O(n²) work + memory churn. On wasm the churn trapped at scale (#696); the quadratic copy made the `List[String]` build slow (#695).

## Fix

`__cow_check` now returns the pointer unchanged when the object is **uniquely owned** (`rc ≤ 1`) — no alias to protect, so the in-place mutation is safe — and only copies a genuinely **shared** object (`rc > 1`):

```rust
if frees {
    // rc <= 1 (rc < 2) → unique → skip the clone
    w.get(Local(0)).i32c(Imm32(rc_neg)).sub().emit_load(0, rc_ty);
    w.i32c(Imm32(2)).lt_u();
    w.if_void(|w| { w.get(Local(0)).ret(); }, |_| {});
}
// shared (or ALMIDE_WASM_FREES=0): unchanged unconditional clone
```

With `ALMIDE_WASM_FREES=0` the rc is a no-op, so the unconditional clone is kept (a `rc≤1` test would always pass and break value semantics under the leak model).

## Verification

- **#696** merge sort and **#695** `List[String]` build run in ~0.15 s at 50k / 20k, **byte-identical native ⇄ wasm** (was: trap / O(n²)).
- **Value semantics preserved**: `var b = a; list.push(a, 4)` → `a` len 4, `b` len 3 on both targets (the `rc > 1` path still clones).
- `almide test spec/`: **277/277** files green; the frees churn gate (record / TCO-deep / TCO-loop) green; `cargo test -p almide-codegen` 107/0.
- New regression `spec/lang/cow_unique_no_clone_test.almd` (value semantics + an 8000-element merge sort + a 4000 `List[String]` build, both targets).

Closes #696. Closes #695.

### DIFF
--- a/crates/almide-codegen/src/emit_wasm/runtime.rs
+++ b/crates/almide-codegen/src/emit_wasm/runtime.rs
@@ -830,21 +830,29 @@ fn compile_rc_dec(emitter: &mut WasmEmitter) {
 /// body carries no hardcoded element-size — it works uniformly for List/String/
 /// Map/Record/Bytes/variant blocks, all of which __alloc stamps with their size.
 ///
-/// This clones UNCONDITIONALLY (a data-section pointer, which has no header, is the
-/// only pass-through). It does NOT branch on the refcount: in the current WASM
-/// runtime the rc header guard (rc_inc/rc_dec) is a no-op (a bump-allocate-and-leak
-/// model — see `HEAP_START_GLOBAL_IDX`), so the rc never reflects aliasing and a
-/// `rc>1` test would never fire. Unconditional clone matches the Rust target's
-/// eager `.clone()` at the bind: correct, and the extra copy when the alias is
-/// already dead is the accepted, conservative cost of `needs_cow` marking. The
-/// original block is left untouched (it leaks like every other block today), so no
-/// refcount bookkeeping is needed.
+/// A data-section pointer (no header) is the one pass-through. Otherwise: when
+/// frees are ENABLED (the 0.27.0+ default — a real refcount), a UNIQUELY-owned
+/// object (`rc <= 1`) has no alias to protect, so the mutation is safe in place
+/// and we return `ptr` unchanged; only a genuinely SHARED object (`rc > 1`) is
+/// cloned. This was unconditional under the original bump-allocate-and-leak model
+/// (where rc was a no-op and a `rc>1` test could never fire) — but that left a
+/// `needs_cow` var mutated in a loop (`while … { list.push(merged, …) }` where
+/// `cur = merged` aliases it) cloning the WHOLE list on EVERY push: O(n²) work and
+/// memory churn that trapped on wasm at scale (#696, and the List[String] O(n²)
+/// of #695). With `ALMIDE_WASM_FREES=0` the rc is a no-op, so we keep the
+/// unconditional clone (a `rc<=1` test would always pass and break value
+/// semantics). The shared (`rc>1`) path is unchanged — the original block is left
+/// untouched (it leaks like every other block under the leak model), matching the
+/// Rust target's eager `.clone()` at the bind.
 fn compile_cow_check(emitter: &mut WasmEmitter) {
     use super::engine::{WasmBuilder, layout::*};
     let type_idx = emitter.func_type_indices[&emitter.rt.cow_check];
     let size_neg = emitter.layout_reg.alloc_header_neg_offset(alloc::SIZE) as i32;
     let size_ty = emitter.layout_reg.field(ALLOC_HEADER, alloc::SIZE).ty;
+    let rc_neg = emitter.layout_reg.alloc_header_neg_offset(alloc::RC) as i32;
+    let rc_ty = emitter.layout_reg.field(ALLOC_HEADER, alloc::RC).ty;
     let alloc_fn = emitter.rt.alloc;
+    let frees = wasm_frees_enabled();
 
     // locals: 1 = $size (data byte count), 2 = $new_ptr
     let mut f = Function::new([(2, ValType::I32)]);
@@ -855,6 +863,13 @@ fn compile_cow_check(emitter: &mut WasmEmitter) {
         // global directly (the rt field is still 0 at this compile point).
         w.get(Local(0)).gget(HEAP_START_GLOBAL_IDX).lt_u();
         w.if_void(|w| { w.get(Local(0)).ret(); }, |_| {});
+        if frees {
+            // Uniquely owned (rc <= 1, i.e. rc < 2) → no alias to protect, skip the
+            // clone. (#696/#695) Only a shared object falls through to the copy.
+            w.get(Local(0)).i32c(Imm32(rc_neg)).sub().emit_load(0, rc_ty);
+            w.i32c(Imm32(2)).lt_u();
+            w.if_void(|w| { w.get(Local(0)).ret(); }, |_| {});
+        }
         // size = header.SIZE; new = alloc(size); memcpy(new, ptr, size); return new.
         w.get(Local(0)).i32c(Imm32(size_neg)).sub().emit_load(0, size_ty).set(Local(1));
         w.get(Local(1)).call(alloc_fn).set(Local(2));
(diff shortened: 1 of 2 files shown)


O6lvl4:
Closing: the rc-gating approach is unsafe. `__cow_check` runs in BOTH rc-maintained code and test blocks, and `insert_rc_ops` (pass_perceus.rs:811) skips test functions — so the refcount is never maintained inside a `test` block (it stays at the alloc value 1). Gating the COW clone on `rc<=1` therefore always skips the clone in tests, breaking value semantics (`spec/lang/alias_cow_test.almd` double-frees → wasm trap; the local `almide test spec/` masked it via native fallback, CI `Test WASM` caught it). Correct for all real code, wrong in the test harness. Root cause + fix directions posted to #696.

PR #726, opened 2026-06-21
Reap child process groups and bound output in the nightly fuzz harness

O6lvl4:
## Why

The nightly differential fuzz [run #723](https://github.com/almide/almide/actions/runs/27897273202) went red with **exit 143 (SIGTERM)** at ~25 min — not because of the 478 native↔wasm divergences it found (those don't fail the step), but because the **fuzz harness itself wedged**, and the crash then **skipped `Upload findings` / `Open issue`, discarding all 478 repros**.

Root cause: `almide run` executes the generated native binary with `Command::status()`, so that binary **inherits the fuzzer's captured stdout pipe**. When a generated program hangs, the per-program timeout kills only the direct `almide` child — the grandchild survives holding the pipe, so the reader thread blocks forever in `read_to_end`. Unbounded capture of multi-MB wasm memory dumps is a second OOM vector.

## What

- **`oracle/runner.rs`** — spawn each child in its own process group (`process_group(0)`) and `kill(-pgid, SIGKILL)` the whole tree on timeout, so hung grandchildren die and the readers drain to EOF. Cap each stream at **4 MiB**, flag overflow, and tear the child down (bounds memory + the findings artifact). Adds 3 regression tests (pipe-holding grandchild / runaway output / clean child).
- **`oracle/ladder.rs`** — classify overflow (native runaway → `Hang`, wasm dump → `OutputDivergence`) and clip divergence summaries so a pathological line can't bloat the log, dedup key, or artifact.
- **`.github/workflows/fuzz-nightly.yml`** — count findings + upload + open-issue moved to `if: always()` steps, so a campaign crash/timeout no longer silently throws away the night's findings.

## Validation

- `cargo test --release` in `tools/xtarget-fuzz`: 16/16 pass (incl. the 3 new regression tests).
- A real 1-minute campaign completes cleanly (dedup works, findings recorded, summary printed — no wedge).
- Replaying captured findings is fast (~2.8s) with bounded summaries.
- Diff is scoped to the fuzzer harness + workflow only — no codegen/spec changes.

The 478 codegen divergences are tracked separately for the wasm-codegen burndown (dominated by wasm string-representation bugs).

### DIFF
--- a/.github/workflows/fuzz-nightly.yml
+++ b/.github/workflows/fuzz-nightly.yml
@@ -96,15 +96,27 @@ jobs:
           STATUS=${PIPESTATUS[0]}
           set -e
           echo "status=$STATUS" >> "$GITHUB_OUTPUT"
+
+      # Count findings in a SEPARATE always()-step. The campaign writes
+      # each finding to disk as it is discovered, so its findings survive
+      # even if the campaign step itself is killed mid-run (an OOM, or the
+      # job timeout). Computing the count here — not inside the campaign
+      # step — is what lets the upload/issue steps still fire after a
+      # crash, instead of silently discarding a night's worth of repros.
+      - name: Collect findings
+        id: collect
+        if: always()
+        run: |
           FINDINGS_DIR="tools/xtarget-fuzz/findings"
           COUNT=0
           if [ -d "$FINDINGS_DIR" ]; then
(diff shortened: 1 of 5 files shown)


O6lvl4:
Triage of the 478 findings this unblocks is tracked in #727 (clustered to ~8 root causes, dominated by wasm string representation). Once this lands, the next nightly attaches the full minimized-repro artifact and updates #727 automatically.

Issue #939, opened 2026-07-27
json.parse is superlinear on the wasm leg: a 5ms VRM load became >11min between 0.15.8 and 0.36.0


O6lvl4:
`json.parse` on the wasm leg is superlinear: doubling the document multiplies parse time by 4–15x. A real workload that loaded in **5 ms** under 0.15.8 does not finish in **11 minutes** under 0.36.0.

This is not a micro-regression. It takes a working application to unusable.

## Minimal reproduction

```almide
import json

var g_in = bytes.new(4194304)

@export(wasm, "in_ptr")
fn in_ptr() -> Int = bytes.data_ptr(g_in)

@export(wasm, "parse")
fn parse(len: Int) -> Int = {
  let s = bytes.to_string_lossy(bytes.slice(g_in, 0, len))
  match json.parse(s) {
    ok(v) => list.len(option.unwrap_or(json.get_array(v, "a"), [])),
    err(_) => 0 - 1,
  }
}

// Control: the slice + string conversion without the parse.
@export(wasm, "slice_only")
fn slice_only(len: Int) -> Int = string.len(bytes.to_string_lossy(bytes.slice(g_in, 0, len)))
```

The host writes `{"a":[{"i":0,"name":"n0","v":[0,1,2]}, …]}` into `in_ptr()` and calls `parse(len)`.

`almide 0.36.0`, `almide build src/mod.almd --target wasm`, run under node 22:

|  JSON size | elements | `parse` ms | `slice_only` ms | ratio vs previous row |
|-----------:|---------:|-----------:|----------------:|----------------------:|
|     10 KiB |      250 |          5 |               0 |                     — |
|     20 KiB |      500 |         18 |               0 |                 3.94x |
|     40 KiB |     1000 |        268 |               0 |            **15.02x** |
|     86 KiB |     2000 |       2205 |               0 |                 8.24x |
|    178 KiB |     4000 |      14063 |               0 |                 6.38x |

Doubling the input should roughly double the time. It multiplies it by 4–15x. The `slice_only` control stays at 0 ms across the whole range, so the slicing and the `Bytes`→`String` conversion are not involved — the cost is inside `json.parse`.

## The real workload

`almide-graphics/nendo` parses glTF/VRM entirely in Almide. `sample.vrm` is a 7.5 MiB VRM whose JSON chunk is 103 KiB — right on the cliff above.

Same model, same host harness, two binaries built from the same source:

| binary | `load_vrm_data(sample.vrm)` |
|---|---|
[...]

Issue #910, opened 2026-07-27
Re-measure the monotonic-accumulator reuse gap now that amortized doubling landed; size-class binning only if it still leaks


O6lvl4:
Supersedes #519. The top candidate policy from that issue has since shipped: `crates/almide-mir/src/concat_to_append.rs` rewrites the self-append window (`x = x + [e]`, `list.push` desugar) to the self-hosted `__list_append1` with `cap = 2·len + 8` amortized doubling (commits b245c780 → a59c9086), targeting exactly #519's canonical case (the accumulator loop; spectralnorm's ~4.8 GB of memcpy). The string self-append shape is handled on the same v1 leg (`lower/binds_b.rs`).

## What to do

1. Re-run #519's repro: a monotonically growing accumulator in a TCO loop (`acc + "x"` kept whole, and the list equivalent), outside any region reset. Measure peak memory.
2. **If bounded**: close this issue and delete the allocator-policy-frontier paragraph from docs/roadmap/active/wasm-frees-ownership-discipline.md — the doubling made freed blocks reusable and the first-fit starvation argument no longer applies.
3. **If it still peaks at the sum of intermediates** (e.g. a shape the append window doesn't catch keeps whole-copy concat): the remaining levers from #519 are size-class binning for the free list, then block coalescing — each requires re-running the byte gate and the churn family, since allocation sizes change.

The point of this issue is the measurement, not the policy work: don't build binning until the repro proves doubling didn't already close the gap.

O6lvl4:
Closed in `3423738d` — the measurement the issue asked for came back SPLIT, and the open half is now fixed.

## The measurement (wasmtime max RSS, monotonic accumulator, no region reset)

| shape | 100k | 400k | 4M |
|---|---:|---:|---:|
| **list** `acc + [i]` (TCO) | 21.8 MB | 20.2 MB | — |
| **string** `acc + "x"` (TCO), before | **1.27 GB** | **died** | — |
| **string** `acc = acc + "x"` (while), before | **1.27 GB** | **died** | — |
| both string shapes, after | 19.6 MB | 11.8 MB | **20.8 MB** |

[...]

PR #940, opened 2026-07-27
Size each parsed JSON string from its own span, not the document remainder

O6lvl4:
Fixes the dominant term in #939. A 103 KiB glTF document parses in **30 ms / 9 MiB** instead of **708 ms / 326 MiB** — 24x faster, 38x less memory — with byte-identical results on every case tested.

## The change

`__jp_string` sized its output buffer from `len - p - 1` — the rest of the **document** — for every string it parsed:

```almide
let buf = prim.alloc_str(if len > p + 1 then len - p - 1 else 0)
```

`__jp_str_fill` then wrote the decoded bytes and `prim.store32` patched the length header down, but the allocation had already been made at full size. Parsing therefore committed O(strings × document length) of memory. glTF is string-dense, so a mid-size model degrades sharply and a large one exhausts the heap.

This sizes the buffer from the string's own span instead, via a new `__jp_str_span` pass. The over-allocate-then-patch discipline is unchanged — the span is still an upper bound, just a tight one.

## Keeping the two passes in step

`__jp_str_span` must mirror `__jp_str_fill`'s advance and stop conditions exactly. Advancing **less** would under-allocate and let the fill write past the end of the buffer, so the span mirrors the fill branch for branch:

| input | fill advances | fill writes | span counts |
|---|---|---|---|
| plain byte | 1 | 1 | 1 |
| `\n` `\t` `\r` `\b` `\f` `\"` `\\` `\/` | 2 | 1 | 1 |
| unknown escape | 2 | 0 | 1 |
| `\uXXXX` (BMP or invalid hex) | 6 | ≤3 | 3 |
| `\uXXXX\uXXXX` valid surrogate pair | 12 | 4 | 4 |

The subtle one is invalid hex: `__jp_str_u` advances 6 unconditionally when `__jp_hex4` fails, which can step **over a closing quote**. A span pass that stopped at that quote would under-allocate, so it has to reproduce the same skip.

## Tests

`spec/wasm_cross/json_string_span.almd` pins every branch where the two passes could disagree — invalid `\u` hex before a quote, lone and paired surrogates, unknown escapes, truncation mid-string and mid-escape, empty strings, multi-byte source bytes, and object keys (which take the same path). As a `wasm_cross` fixture it is pinned against the native oracle, not against itself.

I could not build the compiler here to run the suite — the machine has 13 GiB free against a 24 GiB `target/`, and rebuilding would have contended with the running codopsy worktrees. Instead I compiled **both** parser versions — `develop`'s and this one — into a single module with the shipped 0.36.0 compiler and diffed them on identical input. Full suite runs still need doing.

Correctness, 21 cases including all of the above plus the real 103 KiB glTF: **every one agrees**, compared on `json.stringify` of the parse result so a dropped escape or truncated string would show up.

## Numbers

Fresh instance per measurement — running both parsers in one instance contaminates the second, since the first leaves a heap grown to hundreds of MiB:

| document | develop | this branch | |
|---|---|---|---|
| real glTF, 103 KiB | 708 ms / 326 MiB | **30 ms / 9 MiB** | 24x faster, 38x less memory |
| synthetic 40 KiB | 264 ms / 104 MiB | 49 ms / 25 MiB | 5x |
| synthetic 86 KiB | 1858 ms / 436 MiB | 347 ms / 103 MiB | 5x |
| synthetic 178 KiB | 10004 ms / 1817 MiB | 1978 ms / 425 MiB | 5x |

## What this does not fix

Parsing is **still superlinear** after this change — roughly 4x per doubling at the top of that table. The remaining term is `acc + [v]` in `__jp_array` / `__jp_object`, which accumulate `List[Value]` and `List[(String, Value)]`.

That is a known gap rather than a new finding: `spec/wasm_cross/append_accumulator_heap.almd` (C-105) says so outright — *"the concat is the rc-incrementing `__list_concat_rc` … (O(n^2) copy like v0; n kept small.)"*. Scalar element lists got amortized doubling; heap-element lists did not. Measured on this compiler:

|    n | `List[Int]` | `List[String]` | `List[(String,Int)]` |
|-----:|------------:|---------------:|---------------------:|
|  500 |      0.1 ms |         0.9 ms |               0.9 ms |
| 4000 |      0.4 ms |        52.6 ms |              97.4 ms |

~2x per doubling for scalars, ~4-5x for heap elements. Closing that would take the parser the rest of the way, and belongs with #910 rather than here.

### DIFF
--- a/crates/almide-interp/interp-abstain-ledger.txt
+++ b/crates/almide-interp/interp-abstain-ledger.txt
@@ -62,6 +62,7 @@ io_write_ordering  out-of-interp-scope capability: io.write
 json_gltf_walk  out-of-interp-scope capability: json.parse
 json_number_unicode  out-of-interp-scope capability: json.parse
 json_path_edges  out-of-interp-scope capability: json.parse
+json_string_span  out-of-interp-scope capability: json.parse
 json_stringify_pretty  out-of-interp-scope capability: json.from_string
 json_value  out-of-interp-scope capability: json.parse
 let_unwrap_propagation  out-of-interp-scope capability: value.str
--- a/docs/contracts/README.md
+++ b/docs/contracts/README.md
@@ -114,7 +114,7 @@ Evidence classes (weakest → strongest): `doc-only` < `by-construction` <
 | C-084 | Codec/value decode error messages are byte-identical across targets |  | active | fixture | 1 |
 | C-085 | Float decode widens an integral JSON number to f64 |  | active | fixture | 1 |
 | C-086 | Pass-through stdlib combinators give their result its own reference |  | active | fixture | 1 |
-| C-087 | JSON number and \\u string decoding are byte-identical across targets |  | active | fixture | 1 |
+| C-087 | JSON number and \\u string decoding are byte-identical across targets |  | active | fixture | 2 |
 | C-088 | A Rust-keyword function name compiles on both targets |  | active | fixture | 1 |
 | C-089 | A default parameter referencing an earlier parameter is filled with its argument |  | active | fixture | 1 |
 | C-090 | bytes.from_list on a List[Int] parameter compiles on both targets |  | active | fixture | 1 |
--- a/docs/contracts/conformance.md
+++ b/docs/contracts/conformance.md
@@ -11,7 +11,7 @@
 > (spec-coverage + evidence-class >= fixture for every active contract), so this
 > page cannot legitimately contain an empty Fixtures cell.
 
-64 normative sections; 324 distinct executable fixtures.
+64 normative sections; 325 distinct executable fixtures.
 
 | Section | Contracts | Fixtures (how CI runs each) |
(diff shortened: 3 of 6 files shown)


O6lvl4:
Pushed `acaffc54` fixing all three red checks — one root cause: the new fixture `json_string_span.almd` was missing its two ledger registrations.

- **Emit & Format**: C-087's evidence now lists `json_string_span.almd` (the fixture already declared `// @contract: C-087`; the link must be symmetric). README/conformance regenerated.
- **Test Rust / Build & Test (Linux)**: both failed on the same `interp_abstain_ledger` gate — the fixture uses `json.parse`, which the interp abstains on; added `json_string_span  out-of-interp-scope capability: json.parse` in sorted position. Gate test re-run locally: green.

Note to the session driving this branch: I committed from a detached worktree at `55c4e02a` — pull before your next push.

Issue #943, opened 2026-07-27
import self.X stops linking once the consumer has one more declaration — adding an unrelated let walls the submodule call


O6lvl4:
A call into a `self.` submodule stops linking as soon as the consuming module has more than the barest shape. Adding a single unrelated `let` to the consumer turns a working build into `unlinked stdlib/runtime call(s) with no wasm definition`.

This makes `import self.X` unusable for any real module, so a package cannot be split into files.

## Minimal reproduction

Two files, nothing else in the package.

`src/lib.almd`:
```almide
pub fn write_one(m: Bytes, o: Int, v: Float) -> Unit = bytes.set_f32_le(m, o, v)
```

`src/mod.almd`:
```almide
import self.lib as lib

var g = bytes.new(64)

@export(wasm, "u")
fn u(v: Float) -> Unit = lib.write_one(g, 0, v)
```

`almide build src/mod.almd --target wasm` → **Built.**

Now add one module-level constant that nothing else touches:

```almide
import self.lib as lib

let A = 1                      // <- the only change

var g = bytes.new(64)

@export(wasm, "u")
fn u(v: Float) -> Unit = lib.write_one(g, 0, v)
```

```
error: this program shape is not yet supported by the verified wasm renderer
  wall: Unsupported("unlinked stdlib/runtime call(s) with no wasm definition: almide_rt_lib_write_one
  — rendering them would emit a dangling `(call $…)` (invalid wasm). Add the callee to the self-host
  registry or wall the using function.")
```

| consumer | result |
|---|---|
| no extra declarations | Built |
| `let A = 1` | wall `almide_rt_lib_write_one` |
| `let A = 1` `let B = 2` | wall `almide_rt_lib_write_one` |
| three constants | wall `almide_rt_lib_write_one` |

`var A = 1` instead of `let A = 1` walls identically, as does declaring two ordinary functions instead of the constants, so it is not about `let` or about constants specifically. Declaration *order* moves the boundary in a larger package — in one case hoisting the constants above the `var` made it build again — which is why I would not describe this as a clean "N declarations" threshold so much as the linker losing the submodule once the consumer grows.

Replacing the imported call with an equivalent local function always builds, whatever else the module contains. It is specifically the cross-submodule call that fails to resolve.

## On real code

`Aid-On/aivarium` builds today. Adding one line to its entry module:

```almide
 import self.app as app
+import self.mesh as mesh
 import self.render as render
 import self.dom as dom
```

```
[...]

Issue #906, opened 2026-07-27
bytes.set_f32_le on a module-level buffer walls the wasm renderer — var receiver does not help, and the exported-fn wall drops the inner reason


O6lvl4:
## Summary

Writing into a module-level `bytes` buffer with `bytes.set_f32_le` walls the wasm renderer. Making the receiver a `var` does not help, despite the diagnostic naming "a non-var receiver" as the cause.

This blocks `almide-graphics/nendo`'s VRM viewer (`src/viewer_webgl.almd`), which keeps its whole scratch arena in one module-level buffer and writes matrices into it.

## Repro (5 lines)

```almide
// src/mod.almd
let g_pool = bytes.new(64)

@export(wasm, "probe")
fn probe() -> Int = {
  bytes.set_f32_le(g_pool, 0, 1.5)
  bytes.len(g_pool)
}
```

```
$ almide check
No errors found

$ almide build src/mod.almd --target wasm
error: this program shape is not yet supported by the verified wasm renderer
  wall: Unsupported("exported `pub fn probe` is outside the MIR-lowering subset
  (the wasm module must carry its export)")
```

## A/B

| Variant | `almide build --target wasm` |
|---|---|
| `let g_pool` + no mutator (`fn probe() -> Int = bytes.len(g_pool)`) | ✅ Built, 625 bytes |
| `let g_pool` + `bytes.set_f32_le(g_pool, 0, 1.5)` | ❌ wall |
| **`var g_pool`** + `bytes.set_f32_le(g_pool, 0, 1.5)` | ❌ wall (same message) |

So the mutator call is the trigger, and `var` does not lift it.

## Where the real diagnosis appears

The minimal repro's message names only the exported function. Building nendo, where the same shape is reached from `main`, the nested reason does surface:

```
$ cd nendo/almide && almide build src/viewer_webgl.almd --target wasm
  wall: Unsupported("main is outside the MIR-lowering subset: Unsupported(
  \"in-place mutator bytes.set_f32_le over a non-var receiver (a shared record
  field would alias the write — the two-level record+field COW) not in this
  brick\")")
```

Two things follow:

1. **The message is misleading**: it attributes the wall to a *non-var* receiver, but the A/B above shows a `var` receiver walls identically. Whatever the real condition is, "make it a `var`" is not the fix a reader would infer.
[...]

Issue #881, opened 2026-07-26
v1 wasm renderer walls on package-qualified type references (got Ident 'snaidhm') — ceangal has no wasm path since #782


O6lvl4:
## Summary

`almide check` passes ceangal cleanly, but `almide build --target wasm` walls with ~50 parse errors of the form `Expected type name (got Ident 'snaidhm')`. Since #782 retired the unverified v0 emitter, this is a hard block: **ceangal (and anything using cross-package type references) currently has no wasm path.**

## Environment

- almide **0.36.0** (`almide self-update` on 2026-07-26)
- repro repo: `almide-graphics/ceangal` @ `455d79b`
- `almide.toml` deps: `snaidhm = { git = ... }` (single dep line, lock re-resolved)

## Repro

```
cd ceangal
almide check                    # → No errors found
almide build --target wasm      # → wall
```

Wall reason (representative lines out of ~50, spread over lines 469–2225 of the unit the renderer re-parses):

```
error: this program shape is not yet supported by the verified wasm renderer
  wall: Unsupported("parse error: error: Expected type name at line 469:56 (got Ident 'snaidhm')
  hint: Type names must start with an uppercase letter, e.g. Int, String, MyType
  ...
  error: Assignments return Unit and can't appear here
    at line 1184
    in assignment-in-expr
  error: Expected top-level declaration (fn, effect fn, type, let, var, protocol, test) at line 1185:51 (got Else 'else')
  ...")
```

## Source-side facts (why this looks like a flattener/re-parse issue)

The error count doesn't match the source. In ceangal `src/*.almd` there are only **13** `snaidhm.` references in total:

- 1 import: `import snaidhm.web.gpu as gpu`
- ~8 value-position calls: `snaidhm.upload_items(...)`, `snaidhm.upload_tile_assignment(...)`, …
- **4 type-position references** (all in `mod.almd`):

```almide
var cached_scene: snaidhm.SceneHandle = { ... }   // mod.almd:20
  cubics: List[snaidhm.Cubic],                    // mod.almd:178
  paints: List[snaidhm.Paint],
  shadows: List[snaidhm.Shadow],
```

[...]

Issue #812, opened 2026-07-20
v1 subset burn-down: wall-rate ~30% on fuzz-generated programs, dominated by _hacc folds and map.to_string walls


O6lvl4:
The fuzzer's wall histogram (new taxonomy, campaign 2026-07-20, 5,224 programs) quantifies the v1 MIR-subset debt: **1,540 walls / 5,224 generated (~30%)**. Top reasons:

```
519× main is outside the MIR-lowering subset (unattributed — improve the wall reason first)
131× unlinked: map.fold_hacc
124× unlinked: map.to_string_x_skv_wall
 98× unlinked: map.to_string_ss_str_wall
 85× unlinked: list.fold_hacc
 62× unlinked: result.unwrap_or_hx
 47× unlinked: option.unwrap_or_hx
 29× unlinked: map.get_or_str_wall
```

These are honest walls, not bugs — but every one is a program an LLM could legitimately write. Burn-down order by frequency:
1. **`*_hacc` (heap-accumulator folds, 216×)**: `list.fold`/`map.fold` with a heap accumulator — the single largest brick
2. **`map.to_string_*` (222×)**: Map interpolation/repr for the missing instantiations
3. **`*.unwrap_or_hx` (109×)**: unwrap_or at heap-payload instantiations
4. The 519 unattributed `main is outside the MIR-lowering subset` walls need finer reasons before they can be bucketed

The histogram prints in every campaign summary, so progress is directly measurable. Related: #805 (fasta wall).

O6lvl4:
Progress on develop today. Item 4 of the burn-down order — "the 519 unattributed `main is outside the MIR-lowering subset` walls need finer reasons before they can be bucketed" — is **done**, and three wall CLASSES are closed.

**Wall attribution (item 4).** `ALMIDE_WALL_REASON=1` on the test harness now names which of the three stages declined (render / wat assemble / validate) and threads `verbose` through, so the per-function reason arrives with it:

```
$ ALMIDE_WALL_REASON=1 almide test spec/ --target wasm
[v1-wall] 3 of 23 function(s) outside the lowering subset (NOT rendered):
  __almd_test_12: Unsupported("match over an UNTRACKED subject with a call-bearing arm …")
[wall] spec/…: render: Unsupported("exported `pub fn __almd_test_12` is outside …")
```

The `[render_program]` prefix on the per-function report was also misleading — it appears from the library, not the example — so it now reads `[v1-wall]`.

**Classes closed:**

[...]

PR #944, opened 2026-07-27
Resolve an in-place mutator receiver locally before consulting the globals map (#943)

O6lvl4:
Fixes the reported half of #943 and makes the rest diagnosable.

## The bug

`cow_inplace_receiver` checked `self.globals` **before** resolving the receiver locally. `value_or_global` documents the opposite precedence, and the invariant it relies on:

> A function-local var is in `value_of`. A miss is a global IFF it is in the DECLARED global set — the frontend guarantees every non-global reference is bound by a preceding local form.

The two are not in one numbering space. A module sibling lowers against whichever globals map its region resolves to, while its own VarIds are numbered independently — so `VarId(0)` can be the sibling's first **parameter** and the consumer's first top-let simultaneously.

```almide
// src/lib.almd
pub fn write_many(m: Bytes, o: Int, v: Float) -> Unit = {
  var i = 0
  while i < 16 { bytes.set_f32_le(m, o + i * 4, v + int.to_float(i))  i = i + 1 }
}
```

`m` is a parameter. It was reported as:

```
in-place mutator bytes.set_f32_le over the IMMUTABLE module-level `let` VarId(0)
— a module-level `let` has no storage slot … Declare the buffer `var` to give it one
```

Advice that cannot be followed: there is no `let` to change.

The sibling then walled, and a walled sibling is deliberately non-fatal, so it was silently dropped. The only thing reaching the user was the caller's:

```
unlinked stdlib/runtime call(s) with no wasm definition: almide_rt_lib_write_many
… Add the callee to the self-host registry or wall the using function.
```

— naming a mangled symbol and giving stdlib-gap advice for a user module sitting in the same package. Adding **one unrelated `let`** to the consumer was the whole trigger; without it the identical program built.

## The fix

Resolve locally first; only a `value_of` miss may be a global. The immutable-top-let decline is unchanged for the case it was written for.

## Second change: attribute the unlinked-call gate

The sibling's own wall reason was already recorded — but only printed under `verbose`, which the CLI never sets. The comment at that site says outright that it *must not* wall silently (#881), and it was walling silently anyway.

Sibling reasons now go into `fn_walls`, and the unlinked-call error names them:

```
unlinked call(s) with no wasm definition — the callee is present in the package but
did not survive lowering: `almide_rt_lib_write_many` walled while lowering:
Unsupported("in-place mutator bytes.set_f32_le over the IMMUTABLE module-level `let` …")
```

Same treatment `main` has had since #812 and exports since #906. This is what made the bug above findable at all, and it is what surfaces the remaining cause below.

## Tests

- `a_module_siblings_parameter_is_not_the_consumers_top_let` — the two-module program above must link. **Verified falsifiable**: with the `calls.rs` change reverted it fails, and the failure message is the full attributed chain.
- `a_real_immutable_top_let_receiver_still_walls` — the other side. Resolving locally first must not disarm #906: a genuine module-level `let` buffer written in place must still decline *and still name the fix*.

`cargo test -p almide-mir --lib` → **599 passed, 0 failed.**

One note on the suite: `lower::tests::scalar_loop_with_break_walls` is **flaky independent of this change** — on the unmodified base commit it failed 1 of 2 full-suite runs and passed in isolation, so it carries state or order dependence. It passed in every run with this branch applied. Flagging it rather than folding it in here.

## What this does NOT fix

`import self.X` is still not usable for every module. With this branch, adding a fourth submodule import to `Aid-On/aivarium` now reports its actual cause instead of a bare unlinked symbol:

```
`almide_rt_app_app` walled while lowering:
  List argument cannot be faithfully materialized in this brick
  (would borrow an empty deferred heap value)
```

`app.app()` builds `v.col([...])` — a List literal in call-argument position. It lowers when it is inlined into main and walls on the sibling path, so adding an unrelated module changes an inlining decision and moves it onto the weaker path. That is a lowering-subset gap in `calls_arg.rs`, a different and deeper thing than the precedence bug here, and I have left it for its own change. #943 should stay open for it.


O6lvl4:
Merged `develop` in (40 commits since I branched, and one of them touched `pipeline_c.rs` too — no textual conflict, but that is not the same as still being correct).

Verified on the merged state:

- `cargo test -p almide-mir --lib` → **601 passed, 0 failed**
- the fix is still load-bearing: reverting the precedence change in `calls.rs` makes `a_module_siblings_parameter_is_not_the_consumers_top_let` fail again, so `develop` has not fixed this independently
- `a_real_immutable_top_let_receiver_still_walls` passes, so the #906 diagnostic is intact

Merging.

Issue #941, opened 2026-07-27
Contract ledger and 13 cross-target fixtures cite the deleted `emit_wasm` runtime tree


O6lvl4:
## Summary

`crates/almide-codegen/src/emit_wasm/` no longer exists (the v1 / `almide-mir` trust-spine became the only wasm path), but 16 files still point readers at files inside it: 3 citations in `docs/contracts/contracts.toml` and 13 in `spec/wasm_cross/*.almd` headers.

Dead paths cited:

```
crates/almide-codegen/src/emit_wasm/rt_dragon.rs
crates/almide-codegen/src/emit_wasm/rt_dec2flt.rs
crates/almide-codegen/src/emit_wasm/rt_string_case.rs
crates/almide-codegen/src/emit_wasm/rt_string_extra.rs
```

Neither the directory nor any `*dragon*` file survives anywhere under `crates/`:

```bash
ls -d crates/almide-codegen/src/emit_wasm   # No such file or directory
find crates -name '*dragon*'                # (nothing)
```

## Example

C-023's statement says the wasm runtime "implements Dragon4 with exact big-integer arithmetic (rt_dragon.rs)", and `spec/wasm_cross/float_shortest_roundtrip.almd` opens with:

```
// The WASM runtime implements Dragon4 with exact
// big-integer arithmetic (crates/almide-codegen/src/emit_wasm/rt_dragon.rs).
```

The contract itself is fine and the fixture passes — only the pointer to the implementation is stale.

## Why the gate misses it

`scripts/check-contracts.sh` validates that every `evidence.path` exists on disk and that the fixture↔contract link is bidirectional. It does not look at paths mentioned inside `statement` prose or fixture comment headers, so these rot silently.

## Suggested fix

1. Repoint the 4 filenames at their current homes under `crates/almide-mir/src/render_wasm/`.
2. Optionally extend the gate to flag `crates/...` -shaped strings in contract statements and fixture headers that do not resolve — the same class of hole as #938 (`since` declared REQUIRED but never checked).

Found while verifying the wasm guide's cross-target claims for the docs site; the guide now cites contract IDs rather than source paths for this reason.

PR #952, opened 2026-07-28
Build each cross-target fixture once instead of six times

O6lvl4:
Three gates ask three different questions about the same 318 fixtures in `spec/wasm_cross/`, and each walked the corpus itself:

| gate | question | builds per fixture |
|---|---|---|
| `wasm_cross_target_spec` | native == wasm | native + wasm |
| `interp_cross_target_spec` | interp == native == wasm | native + wasm ← both duplicates |
| `wasm_opt_parity_spec` | wasm == wasm-opt | wasm + wasm-opt ← plain wasm duplicate |

**1908 builds, of which 954 were redundant.** They now share one lazily-built table of legs, so the corpus costs native + wasm + wasm-opt — half the builds.

**384s → 205s locally (47%)**, matching the predicted halving.

## Why they had to move into one binary

Each `tests/*.rs` compiles to a **separate process**, so a cache in one cannot serve another. That is the whole reason the wasm-opt and interp gates were relocated rather than left where they were. `almide-interp` was already a regular dependency of the root package, so there was no crate-boundary obstacle.

Relocating also **deleted duplicated code**: the interp harness carried its own copies of `run_native_capture` / `run_wasm_capture` / `build_scratch` — ~90 lines mirroring the root helpers. Net **−138 lines**.

## What was deliberately preserved

These are the project's core correctness gates. A refactor that quietly weakened one would cost far more than the minutes it saved, so:

- **Every classification arm was moved verbatim.** Only the source of the legs changed.
- **`interp_abstain_ledger` does not touch `corpus()`.** Its value comes from being backend-free — it never self-skips on a missing `almide` binary or wasmtime. Reading the shared table would trigger 318 native+wasm builds and destroy exactly that property. Enforced in code and stated in the comment.
- **`wasm-opt` stays optional.** Absent, only the parity gate skips; the other two still have everything they need.
- **A wasm build/run panic is still surfaced**, via a sentinel leg each gate reports in its own words rather than swallowing.

## Verification — verdicts, not just timings

Captured each gate's summary before the change and compared after. All four identical:

```
wasm_cross_target_spec    318 equal, 0 tracked-divergence, 0 stale-allow, 0 unexpected
wasm_opt_parity_spec      318 equal, 0 mismatch
interp_cross_target_spec  318 fixtures | 176 interp==native==wasm | 142 skipped | 0 split | 0 dissent
interp_abstain_ledger     318 fixtures | 176 evaluated | 142 abstained (ledgered)
```

Regenerating the abstain ledger moved **zero entries** — the only diff was the two header lines documenting the new command. Independent confirmation that the interp leg's verdicts are unchanged.

`cargo test --workspace` (the CI command): 92 suites, 1823 tests, 0 failures.

## References repointed

Nine places still named the deleted files — `docs/ARCHITECTURE.md`, three roadmap docs, `src/cli/build.rs`, `.github/workflows/ci.yml`, `crates/almide-interp/CLAUDE.md` and `Cargo.toml`, plus the ledger header. Leaving them would have reproduced the exact dead-citation problem #941 just fixed.

The load-bearing one: the ledger's regeneration command was `cargo test -p almide-interp interp_abstain_ledger`, which after the move **runs nothing**. Now `--test wasm_runtime_test`, and verified by actually running it.

`almide-interp`'s `tempfile` dev-dependency went unused once the duplicated capture helpers were deleted, so it is gone too.

## Sizing note

An earlier estimate in #951 put this at "~10 minutes". That was wrong — it extrapolated a local ratio to CI. The measured CI breakdown after #951 showed the interp gate already down to 1.70 min, and the real duplication was the third gate (`wasm_opt_parity_spec`, 3.07 min) which that estimate had not accounted for at all. The 47% here is measured, not projected; what it converts to in CI will be visible in this PR's own `Test Rust` timing.


O6lvl4:
Rebased onto develop (post-v0.40.1) with the semantics that landed since this PR was written carried INTO the unified architecture — a naive rebase would have reintroduced the #991 bug class this PR predates:

- **#998 semantics ported to the shared leg builder** (`p4_corpus.rs`): a guest exit 127 is a comparable observable (`Ok(o) => Some`, `Err => None` — spawn failure only), and a mid-run wasmtime spawn failure becomes a `<wasmtime-spawn-failed>` SENTINEL LEG like the panic sentinel — never the old `return None` that silently skipped every gate over every fixture (the mid-corpus green-return, which this PR had centralized).
- The equality gate reports both sentinels as that fixture's failure and continues; the parity/interp gates surface them as visible divergences.
- ci.yml composed cleanly with the re-armed gates from #995/#1009 (delta vs develop: one comment line).
- interp CLAUDE.md conflict resolved to the relocated-gate wording.

Re-verified on the rebase: **75 passed / 0 failed in 225s** (all three gates, one corpus walk, real wasmtime + wasm-opt) vs ~380s for the pre-unification split on the same machine — the halving claim holds on the current 324-fixture corpus.

Issue #956, opened 2026-07-29
Numeric conversion surface: complete the lossy-pair trio matrix, add bounds, and gate the shape


O6lvl4:
## Motivation

The conversion trio (`to_T` / `to_T_saturating` / `to_T_checked` — Swift's `truncatingIfNeeded` / `clamping` / `exactly`) is complete **only from canonical `Int`** (`stdlib/int.almd`). The eight sized modules (`int8`..`uint64`) expose only the plain truncating form for sized→sized pairs: `uint8.to_int8_checked` does not exist, and neither do bounds constants (`int8` min/max appear as bare literals even inside the stdlib's own saturating bodies).

This is the point-wise-growth failure mode: each function was added where a caller needed it, and nothing asserted the *shape* of the family. Per the new CLAUDE.md rule (**API families are extended by matrix, never point-wise**), this issue completes the matrix AND lands the executable gate that keeps it complete.

## Completeness rule

Over the 9 integer carriers {Int, Int8, Int16, Int32, Int64, UInt8, UInt16, UInt32, UInt64}, for every ordered pair (src, dst), src ≠ dst:

- the plain conversion `src_mod.to_dst` exists (already true today);
- if the pair is **lossy** (src's value range ⊄ dst's range): `to_dst_checked -> Option[Dst]` and `to_dst_saturating -> Dst` MUST exist;
- if the pair is **lossless**: the `_checked`/`_saturating` forms MUST NOT exist (an always-`Some` checked variant is noise in a monomorphic enumerated API — same reasoning as the existing omission of `int.to_int64_checked`).

Lossy sized→sized pairs: 38 (I8:4, I16:5, I32:6, I64:7, U8:1, U16:3, U32:5, U64:7) → **76 new one-line fns**, each a pivot composition `int.to_dst_checked(int.from_src(x))` — no new logic, the Int-side trio is the single source of semantics.

## Bounds

`min_value()` / `max_value()` on all 8 sized modules (16 fns). Named `*_value` because `int.min`/`int.max` are already the binary ops — reusing the bare names with different arity across sibling modules is exactly the kind of naming trap this campaign exists to remove.

## The gate

An in-process matrix test in the `check_stdlib_fn_count` style (module-interface scan): encode each carrier's (min, max) once, compute lossiness per pair, generate the expected name set, and diff it against the actual module interfaces — both directions (missing cell = red, unexpected always-lossless variant = red). A future sized type (or a new conversion family) cannot land with a partial surface.

[...]

PR #969, opened 2026-07-29
Admit the checked numeric conversion family by shape so value-position matches over runtime scrutinees execute

### LINKED ISSUE #958: Value-position variant match with a runtime scrutinee walls the v1 renderer (heap-result match outside the executable subset)
## The gap

An Option/Result `match` whose **result is used as a value** (bound to a `let`, returned, or fed to a call argument) and whose **scrutinee is a runtime value** is an unshipped v1 brick. Two wall spellings of the same gap:

```almd
// (a) let-bound unwrap — "variant (Option/Result) match bound to a let/var
//     outside the executable subset cannot be faithfully computed"
effect fn main() -> Unit = {
  let small: UInt64 = 42
  let v = int.from_uint64_checked(small)!
  println(int.to_string(v))
}
```

```almd
// (b) unwrap chained in call-arg position inside a test fn — "heap-result
//     `match` outside the executable subset cannot be faithfully returned in
//     this brick (would move out an empty deferred heap value)"
test "unwrap a runtime checked result" {
  let u200 = int.to_uint8(200)
  assert_eq(uint8.to_int8_checked(int.to_uint8(5))!.to_int64(), 5)
}
```

Match in **statement** position (arms perform effects) lowers fine, as does `?? fallback` on the same scrutinees — those are the current workarounds.

## Why it surfaced now (#956)

`spec/stdlib/sized_conversion_safety_test.almd` has used `int.to_int8_checked(42)!.to_int64()` for months without walling — but only because its arguments are **literals**, so the whole Option folds at compile time and no runtime match survives to lowering. The #956 matrix test was the first corpus content to unwrap checked conversions of *runtime* values; the walled-real ratchet correctly rejected the five new walls (`2e224741`'s Trust Spine red). The test now uses `(checked(...) ?? sentinel).to_int64()` with a header note; this issue is the brick that would let `!` be written there naturally.

## Relation

Same residue family as the #810 terminal default-deny and the arm-discipline bricks (#907/#945 `lower_scalar_arm`/`lower_heap_result_arm`): what is missing is faithfully *returning* the matched value out of a runtime variant match in value position. For `Option[Int]`/`Option[sized]` the payload is a scalar lane, so a first brick could special-case scalar-payload Option before the general heap-payload case.

O6lvl4:
Fixes #958.

## The gap was an admission drift, not a missing mechanism

The value-position variant match machinery (`try_lower_variant_value_match` — tag-read, run only the taken arm, bind the scalar payload) already existed and is proven. What walled the issue's repros was the SUBJECT admission: `is_self_host_option_module_fn` tracked materialized-Option call results through a hand-enumerated per-fn list, and that list drifted the moment #956 completed the conversion matrix. The matrix spans **10 numeric carrier modules × 58 checked cells**; the list covered 17 (`int` and `float` only) — `int.from_uint64_checked` (repro a) and every sized-module cell (`uint8.to_int8_checked`, repro b) fell to the wall. Exactly the "API families are extended by matrix, never point-wise" failure mode CLAUDE.md warns about, in the compiler's own admission table.

## The fix, applied at the family level

- **Shape rule** replaces the enumerated int/float arms: numeric carrier module × `to_/from_*_checked` name ⇒ admitted. Sound because every cell is self-hosted through the ordinary `some()`/`none` ctor rails and returns a materialized Option over a scalar lane — the invariant the tag-read lowering requires.
- **Derived matrix gate** (`checked_conversion_family_is_admitted`): reads the cell list from the stdlib sources themselves and asserts the predicate admits each (58 cells; floor of 50 guards against the scan going blind). A new cell cannot drift out of the admission again — the point-wise failure mode is structurally closed.
- **C-195** + `spec/wasm_cross/checked_conversion_value_match.almd`: runtime scrutinees (env.args-derived, nothing folds) through let-bound both-arms, call-argument, return-position, the `!` unwrap, and int/float/sized carriers — byte-identical native == wasm (`42 / -1 / 200 / 5 / 7 / fits 12 / overflows`). Interp abstains on `prim.args_get_list`, ledgered.

Both issue repros now run: the (a) let-`!` spelling and the (b) sized-module call-arg spelling, plus the explicit value-position `match` over a runtime scrutinee that reproduced the wall verbatim.

## What stays outside this brick

Heap-payload variant subjects (`Option[String]` arms borrowing the subject slot) remain the documented Camp-4 frontier of `try_lower_variant_value_match` — unchanged by this PR, still an honest wall.

## Verification

| gate | result |
|---|---|
| `cargo test --workspace --no-fail-fast` | green |
| `almide test` (full spec) | 323/323 |
| new fixture, native vs wasm | byte-identical |
| `scripts/check-contracts.sh` | OK — 195 contracts, 324/324 fixtures linked |
| matrix gate | 58 cells admitted |


O6lvl4:
The Coq/PCC gate failure was a REAL find by the caps gate, now fixed in `3fcf94f4`:

**`classify` (the new fixture's fn) breached `mir_calls <= ir_calls` (mir 6 > ir 5).** The counter skips identity int-widening nodes (`int.to_int64`, `int8.to_int64`) by predicate on the assumption the lowering never emits a call for them — true in operand position (`lower_scalar_call_form` elides), but the value-match brick exercised two producers that funneled through `lower_pure_module_value_call` and emitted real `CallFn`s: a widening in ARGUMENT position (`to_int8_checked(int.to_int64(v))`) and one inside a heap-result arm's interpolation part (`"${x.to_int64()}"`).

Fix, per the same-predicate discipline:
- The predicates now have shared `(module, func, args)` cores (`is_identity_int_widening` / `is_float_from_int_prim`) that both the expr-based counter predicate and the lowering consult — one implementation, structurally unable to drift.
- `lower_pure_module_value_call` (the choke point every non-operand producer funnels through) elides both shapes at entry: widening forwards the operand, `float.from_int` emits the single `F64FromInt` prim.
- `ALMIDE_DEBUG_CALL_OPS=1` on `classify_corpus` now dumps per-fn call ops — the instrument this diagnosis needed.

Re-verified: fixture breach gone, **whole-corpus wall gate `WALL OK` over 5402 functions**, fixture still byte-identical, `almide test` 323/323, `cargo test --workspace --no-fail-fast` green.

Issue #1008, opened 2026-07-29
Windows leg hangs 4h+ in lsp_didchange_never_fetches_or_writes_lock — possible fetch-on-didChange via GCM prompt; LspClient reads are unbounded; build-cross has no timeout


O6lvl4:
Surfaced by #983's arming of the Windows leg (`cargo test --workspace` on build-cross): run 30478897403's Windows job hung for **4h20m** inside `tests/lsp_test.rs::lsp_didchange_never_fetches_or_writes_lock` (started 18:36Z, still "running" at 22:33Z when cancelled; macOS and Linux legs green in minutes). Without a job timeout it would have burned the full 6h default.

## Evidence

```
2026-07-29T18:35:14  Running tests\lsp_test.rs (...)
2026-07-29T18:35:14  running 20 tests
2026-07-29T18:36:14  test lsp_didchange_never_fetches_or_writes_lock has been running for over 60 seconds
(no further test output until cancellation at 22:33)
```

## Hypotheses (unverified — no Windows box here)

1. **The test may have caught the real thing it guards**: the fixture's manifest declares `nonexistent_pkg = { git = "https://invalid.invalid/nowhere" }`. If the Windows path/URI handling makes the server MISS the no-fetch guarantee and shell out to git, Windows git commonly blocks on a **Git Credential Manager interactive prompt** instead of failing fast like Linux — an infinite hang is exactly what a fetch-on-didChange looks like on a Windows runner. If so this is a product bug (the #961 guarantee broken on Windows), not a flaky test.
2. **Malformed file URI on Windows**: the suite builds URIs as `format!("file://{}", path.display())` → `file://D:\a\...` (backslashes, no third slash). Depending on the server's URI parsing, requests may target a path that never resolves and some code path may wait.
3. Harness-level: `LspClient::recv` is an UNBOUNDED blocking read — ANY server non-response is an infinite test hang rather than a fast red.

## Mitigations landing now (separate PR)

- `timeout-minutes: 60` on build-cross (the trust-spine #921 lesson: a hang must be a fast red, not a 6h grind).
- `GIT_TERMINAL_PROMPT=0` workflow-wide — no CI job should ever wait on a credential prompt; if hypothesis 1 is right this converts the hang into the intended fast failure.
- `LspClient` reads through a reader thread + `recv_timeout(60s)` with a kill-on-drop child, so any future LSP hang fails in ~1 minute naming the request.

## Closure condition

[...]

PR #1010, opened 2026-07-29
DO NOT MERGE — #1008 evidence run: Windows LSP hang instrumentation

O6lvl4:
Throwaway PR to run build-cross (Windows) on instrumented code — **never merge; will be closed after the log is read** (#1008).

Contents:
- **Fix candidate under test**: every LSP request now gets a response — `handle_request`'s `None` (bad params / no analyzed doc) replies null and LOGS instead of silently dropping (a JSON-RPC violation that turned any client await into an infinite hang; the Windows leg has burned 6h runs inside lsp_test on every develop→main PR since #961 — verified on run 30443590068: 10:25→16:25, killed at the GitHub default limit).
- Server-side `ALMIDE_LSP_TRACE` phase tracing (didChange: uri/path, dep-resolution time, analyze time).
- Test client captures server stderr; the didchange test dumps uri/response/server-trace unconditionally.
- PROBE-ONLY: build-cross runs just `cargo test --test lsp_test -- --nocapture --test-threads=1`.

Expected outcomes on the Windows leg:
- **Green** → the always-respond fix resolves the hang; the trace names which `None` used to fire. Fix + client/server keepable parts get promoted to develop and shipped as v0.40.1.
- **Red in ~1 min** (deadline panic with server stderr) → the hang is deeper than the response protocol; the trace shows the last phase entered.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -543,10 +543,13 @@ jobs:
       # DELIBERATELY unarmed (#983 decision): wasmtime is not installed there,
       # so the tool-gated wasm suites skip on Windows by design — Linux
       # (test-rust) and macOS are the armed legs.
-      - name: Cargo tests (workspace)
+      # PROBE ONLY (#1008 evidence run — this branch never merges): run just
+      # the LSP suite, serial and uncaptured, so the Windows hang's server-side
+      # trace lands in this log verbatim.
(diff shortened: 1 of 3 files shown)


O6lvl4:
Evidence collected — root cause proven (see #1008 / the fix PR). Windows leg: 20/20 in 2.17s with the always-respond dispatch. Closing the probe.

Issue #1001, opened 2026-07-29
Dogfood a 10k-line real Almide project — nothing at scale exists, so every scale claim and scale-triggered optimization is theoretical


O6lvl4:
## Background

The compiler pipeline is measured linear (~1ms/line for a native debug build: 4,000-fn file = 4.2s full build), but no real Almide program anywhere near that size exists. Consequences:

- Scale claims ("Go-class builds") are only demonstrated up to a few thousand lines
- The per-module compilation cache (#1003) has a trigger condition — full build > 2–3s in a real project — that nothing can currently trip
- There is no answer to "what was built in Almide that couldn't have been?"

## Task

Build one substantial (~10k-line) real program in Almide. Candidate shapes: a tool the almide/almide-dojo ecosystem actually needs (task-bank tooling, a docs-site generator, an .almd-aware code-search/index tool), so the effort is not throwaway.

Value beyond the artifact itself:
- Generates the real-world pain signal that arms or dismisses #1003 (and the http-rlib item #1002) with data instead of extrapolation
- Exercises the module system, package system, and diagnostics at a scale the spec suite doesn't
- Becomes the first "written in Almide" proof point

This is the driver item for the trigger-gated backlog: the deferred optimizations stay deferred until this exists.

O6lvl4:
## Done — `tools/almide-gates`, seven byte-identical subcommands

Unit 0.46 is closed. The program is this repository's own quality gates, rewritten in Almide,
with byte-identity against the bash originals as the acceptance check throughout.

| subcommand | bash original | evidence |
|---|---|---|
| `roadmap-readme` | `docs/roadmap/generate-readme.sh` | 390 lines / 59,262 bytes identical |
| `contracts-readme` | `docs/contracts/generate-readme.sh` | 232 lines / 25,450 bytes identical |
| `conformance` | `docs/contracts/generate-conformance.sh` | 81 lines identical |
| `stamp` | `proofs/lib/stamp.sh` | identical — and it has aborted a real measurement |
| `output-parity` | `proofs/output-parity.sh` | 10 lines / 607 bytes, both exit 0 |
| `fuzz-track-record` | `scripts/fuzz-track-record.sh` | 12 lines / 759 bytes identical |
| `check-contracts` | `scripts/check-contracts.sh` | 13 lines / 542 bytes + **15/15 mutations** |

~2,100 lines across thirteen modules, 43 tests, gated in CI.

### Eight defects found by writing it

[...]

Issue #924, opened 2026-07-27
fuzz-nightly full-budget ratchet: 14 consecutive completed nights (instrument re-measured ONLINE 2026-08-08)


O6lvl4:
> **状態** — 計測器・ゲート・判断待ち
> **一行で** — 終結条件「14 夜連続で予算を完走」は、verdict job が `if: always()` で常に出るようになった後は構成上満たされてしまうため数え直しが必要で、代わりに提案された「14 夜連続で planned fuzz-minutes の 75% 以上を配送（`shards=k/N` から読む）」は 2026-08-08 のコメントに「裁定は maintainer に委ねる」と書かれたまま裁定されていない。
>
> **やること**
> 1. 数える条件を 1 つ選ぶ — 本文の「14 夜連続 full-budget」か、下のコメントの「14 夜連続で planned fuzz-minutes の ≥75%」か。これが決まらないと何を数えているのか言えない。
> 2. 決まった条件で `bash scripts/fuzz-track-record.sh 20` を走らせ、現在値をこの issue に記録する（この issue に残る最後の実測は 2026-08-17 の「1 夜」）。
> 3. #2390（verdict 行の数え落ち: preempt された shard の finding が総数から抜け、本文の列挙は 20 件で打ち切られる）を閉じる。coverage と findings を同じ `fuzz-night:` 行から読むので、その行の数が信用できない限り streak の値も信用できない。
> 4. bot の finding はもうこの issue に積まれない（#2379 で `fuzz-findings` label へ routed）。この issue が持つのは ratchet だけになる。
>
> **完了条件** — 裁定された条件で `scripts/fuzz-track-record.sh` が 14 夜連続を報告し、その出力がこの issue に貼られる。
> **根拠** — `scripts/fuzz-track-record.sh:21-29` と `.github/workflows/fuzz-nightly.yml:38,45-48`（#1014 以降 12 夜のうち 10 夜が完走、kill 率 ~1/6、「14 夜連続 full-budget」は (5/6)^14 ≈ 8%）。この issue のコメントの実測: 2026-07-30 に 3/14、2026-08-08 に 4/14、2026-08-17 に 1 夜（corrected metric）。2026-09-21 に `gh run list --workflow=fuzz-nightly.yml --limit 20` は success 8 / failure 10 / cancelled 2 を返すが、赤は finding でも赤になるので完走の判定には使えない。
[...]

Issue #796, opened 2026-07-18
Fuzz nightly to true green: 3 open findings (#807/#808/#809) + two consecutive green nights


O6lvl4:
Fuzz (nightly) has been red since 2026-07-15. All 12 findings from the 2026-07-18 run (seed 1784352208133210990) reproduce on current develop — real native/wasm divergences, none fixed incidentally by the #782 burn-down.

Worst class: float-fixed-format / ZWJ-emoji strings observed as `ok("")` on wasm — a deferred Opaque reaching observed output, i.e. a wall-coverage hole (the #777/F3 motivation and the silent-wrong-value class the v1 discipline exists to prevent). Also: lists collapsing to `[]`, an i32-boundary tuple returning a completely different value, option flips, negative-zero display, and one wasm build + one run failure. The same workflow's coverage-ratchet job is failing too and needs its own diagnosis.

Plan, repro commands, full findings table, and definition of done: docs/roadmap/active/fuzz-findings-triage.md

O6lvl4:
## 2026-07-20 burn-down: taxonomy rebuilt, 8 divergence classes resolved, 3 filed

**Taxonomy** (ca3288b4, 31ffb582): honest walls are no longer findings — `Outcome::Walled` with a per-reason histogram (subset burn-down data), and a wasm hang whose native leg ALSO failed is a skip (failure-speed races are not divergences). The scary "623 findings" resolved to **6 real + 297 walls**.

**Fixed today** (37abd7ca, a15f2227, 6d48398c — each with fixture+contract, all gates green):
1. **C-158**: scalar-call ctor payloads fell to a zeroed Opaque (`ok(chain)` → `ok(0)`) — ANF route
2. invalid_float_literal acceptance divergence — same root, auto-resolved
3. **C-095**: json pretty-print used Rust `{:?}` escaping (`\u{301}` — invalid JSON) — canonical 5-escape on both targets
4. **C-054**: `repeat` negative counts (native capacity panic 101 vs wasm alloc trap 134) — signed clamp to empty, all three repeat families
5. tuple ctor payloads (`ok((0.3, 1000000))` → OOB abort) — ANF route extended to Tuple
6. **C-159 (new)**: `binary_search` duplicate-key index — the wasm self-host replicated the CLASSIC early-exit loop while Rust std (the oracle) is the 1.52+ branchless loop; now instruction-identical
[...]

Issue #983, opened 2026-07-29
build-cross runs bare cargo test (root package only) and never arms the tool-gated suites — the macOS/Windows and feature-branch legs skip-as-pass


O6lvl4:
From the 2026-07-30 gates audit.

Root `tests/` has 18 tool-gated integration suites (pattern: `tool_available()` / wasmtime probe → early-return AS PASS when the tool is missing). Whether they actually run is a property of each JOB's toolbox, and today:

| job | command | wasmtime | wasm-opt | verdict |
|---|---|---|---|---|
| ci.yml test-rust (Linux) | `cargo test --workspace` | installed | installed | fully armed |
| ci.yml build-cross (macOS) | bare `cargo test` | installed | never installed | wasm_opt_parity_test permanent no-op |
| ci.yml build-cross (Windows) | bare `cargo test` | not installed (`if: runner.os != 'Windows'`) | not installed | all 18 tool-gated suites no-op |
| ci-feature.yml | `cargo test --workspace` | not installed | not installed | all 18 no-op on every feature branch |

And the bare `cargo test` at `ci.yml:473` is the exact mistake `ci.yml:139-142` documents fixing for test-rust: "a bare `cargo test` at the root tests only the ROOT package — the member crates' suites silently never ran in CI." So the macOS/Windows legs also never run almide-interp's abstain-ledger gate, almide-mir's ratchets, or any member-crate unit test.

## Closure condition

`build-cross` runs `cargo test --workspace`; each tool-gated suite either has its tool installed in every job that claims its coverage, or is explicitly platform-scoped (cfg/ignore with a reason) instead of skip-as-pass; and a job-level tripwire (e.g. assert the skip count of tool-gated suites is 0 on the fully-armed leg) so a future tool-install regression is loud.

O6lvl4:
[...]

PR #1035, opened 2026-07-31
v0.44.0 — the dogfood grows a spine, and three shell decision rules become testable

O6lvl4:
## v0.44.0

`tools/almide-gates` (Unit 0.46, #1001) reaches **four byte-identical subcommands** — roadmap README (390 lines), contracts README (231), conformance (81), and the shared toolchain stamp (7) — plus **three decision rules extracted from the shell gates as pure, tested functions**.

~740 lines across nine modules, and **gated in CI**. It is written IN Almide, so a compiler regression breaks it — which is the point of dogfooding it. Previously it could grow unverified.

### The extractions

A second kind of value, distinct from the six defects the dogfood surfaced in v0.43.0. In bash these rules are interleaved with process invocation and network calls, so the RULE can only be exercised by running the whole gate:

**The parity verdict** (`output-parity.sh`) — five outcomes whose ORDER is load-bearing:

```
render fails + oracle also failed  → v0fail
render fails + oracle succeeded    → wall
```

Backwards, every broken fixture reports as a v1 wall — and **the wall count is a ratchet**, so a mislabel inflates a number the project manages as debt. The bash has it right; nothing tested that it stayed right.

**The nightly streak fold** (`fuzz-track-record.sh`) — two streaks over one run list that stop **independently**:

| night | green (#796, closes at 2) | full-budget (#924, at 14) |
|---|---|---|
| no findings | continues | continues |
| findings | **ends** | **continues** |
| truncated | ends | ends |

A night with findings ends one and not the other, because #924 measures whether the instrument **ran**, not whether it was quiet. Collapsing them — the obvious simplification — makes #924 unmeasurable on a compiler still being fixed.

**The contract link symmetry** (`check-contracts.sh`) — run against the real ledger:

```
fixture evidence links checked: 346
asymmetric (ledger says, fixture does not): 0
```

An independent implementation agreeing with the bash audit is the strongest check available short of byte-identity, and it arrived before the full subcommand was ported. It reports its two asymmetries separately, because they fail differently: a ledger entry pointing at a fixture that certifies something else looks covered and is not; a fixture nothing lists could be deleted with nothing breaking.

### The toolchain stamp earned its port

`proofs/lib/stamp.sh` aborts every proof gate when the PATH binary is not the workspace build, because the evidence would describe a different compiler. That fired for real during this cycle — a parity run refused to start until `make install` caught up. The Almide version reproduces the abort exactly, including the sha256 comparison.

It is the same one-binary discipline the fuzz campaigns depend on, and it is now expressed in a typed language rather than in shell.


O6lvl4:
## Not merging yet — two cross-platform Build legs are red, and it is not this PR's diff

`Build (macos-latest)` and `Build (windows-latest)` fail in `Cargo tests (workspace)`:

```
---- armed_jobs_actually_have_their_tools stdout ----
panicked at tests/tool_arming_tripwire_test.rs:35:
ALMIDE_EXPECT_TOOLS=1 but `wasmtime` is not runnable — this job claims full tool
coverage, so the tool-gated suites would silently skip-as-pass. Install it or drop
the claim (#983).
```

### Why it is not the diff

- The same commit (`6805e685`) is **green on develop** — CI, Cross-Target and Trust Spine all success. These two legs only run on a PR to `main`.
- PR #1034 (v0.43.0), 40 minutes earlier, had both legs **pass** on the same workflow.
- This PR's only workflow change adds a step to the `test-rust` job (`almide test tools/almide-gates/`). The failing job is `build`, a different one.
- The `Install wasmtime` step in the failing job reports **success** — so the binary was installed and is nonetheless not runnable at test time. That is a PATH/propagation or cache-restore condition, not a missing install.

### Why I am leaving it red rather than merging

The tripwire (#983) exists precisely so a leg that loses its tools turns red instead of skipping 18 suites as pass. Merging past it would defeat the thing it was built for, and "it is green on develop" is exactly the reasoning it was designed to reject.

Needs a fresh look at why an installed `wasmtime` is not runnable in that job — most likely between the install step and the test step's environment. Handing off rather than guessing at it.

Issue #547, opened 2026-06-11
fan.map pure-mapper rejection prints a garbled expected/actual type pair


O6lvl4:
**Danger class: diagnostic quality (no wrong codegen; misleading error text).**

`fan.map([1, 2, 3], (x) => x * 10)` is rejected by DESIGN (the mapper contract is `Fn(A) -> Result[B, String]`; pure mappers are a check-time error). But the E001 text is garbled:

```
expected fn(Result[?1, String]) -> Result[?1, String] but got fn(Int) -> Result[?1, String]
```

The EXPECTED side prints the callback's PARAM position as `Result[?1, String]` (it should be `Int`, the list element type), and the GOT side prints the lambda's return as `Result[?1, String]` (the lambda returns `Int`). Expected/actual halves appear cross-contaminated by unification order before rendering. The hint should also state the actual rule ("fan.map callbacks must return a Result — wrap the value in ok(...)") rather than a raw type mismatch.

Found while probing #533 (the shape itself is correctly rejected on both targets — no cross-target divergence). Note the now-adjacent design asymmetry worth a doc sentence: race/any/settle ACCEPT pure thunks since #546 (Ok-adapted), while fan.map requires an explicit `ok(...)` — intentional (mapper inference is the hard case), but CHEATSHEET should say so.

PR #1145, opened 2026-08-08
Effect fn-typed slots: effect (A) -> B parameters with carrier semantics

### LINKED ISSUE #1055: Effect-typed fn parameters: declare http.serve's handler as an effect lambda slot
Follow-up to #1051 (the option-1 endgame it proposed).

## Current model (pinned by tests in the #1051 fix)

A lambda inherits the enclosing fn's effect capability — one rule for every higher-order callee (`list.map`, `http.serve`'s handler, …). Inside an `effect fn` a lambda may call effect fns, but results stay **explicit Result values** (auto-`?` never crosses a closure boundary, #489); `!` in a lambda is E022. The `"${resp}"` debug-form trap now warns (`validate_result_interpolations`).

## The endgame this issue tracks

Give the type system effect-typed fn parameters so a higher-order fn can *declare* what its lambda may do:

```almide
effect fn serve(port: Int, f: effect (HttpRequest) -> HttpResponse) -> Unit
```

- A lambda checked against an `effect (…) -> …` slot gets effect-fn body ergonomics: `can_call_effect`, auto-unwrap, `!` with propagation. Its closure lowers to `Fn(...) -> Result<T, String>`; for http.serve the runtime contract already exists (`almide_http_serve` maps a handler `Err` to a 500).
- A lambda checked against a plain `(…) -> …` slot would then be *pure by declaration* — rejecting effect calls even inside an effect fn. That is the symmetric enforcement #1051 asked for, and it is a breaking change for the current contextual model (effectful `list.map` lambdas are legal and used today), so it needs its own migration story.

## Why deferred out of the #1051 sweep

The `#489` invariant (no propagation across closure boundaries) is load-bearing in the checker (`auto_unwrap` off in lambdas), the lowering (`insert_auto_try` never descends into `Lambda`), and codegen's ResultPropagation. Flipping it per-parameter-slot is an ADR-scale design change (effect types in the surface type grammar + `Ty::Fn` effect bit or an out-of-band slot table), not a rider on a 6-issue fix batch.

O6lvl4:
Fixes #1055.

## What a user can now write

```almide
http.serve(8080, (req) => {
  let body = fs.read_text("index.html")!   // ← `!` works: the slot is effect-typed
  http.response(200, body)
})!
```

Previously the handler had to open every Result with `match`/`??` — `!` was
E022 inside any lambda. Now a fn type can be declared effect:

```almide
effect fn serve(port: Int, f: effect fn(HttpRequest) -> HttpResponse) -> Unit
```

and a lambda checked against an `effect (…) -> …` slot gets full effect-fn
body ergonomics: effect calls permitted, `!` propagates into the handler's
own failure channel. A failing `http.serve` handler becomes the listener's
500 response (the runtime loop always had that arm — "future error-in-handler
support" — this is that future).

## Design: one carrier, zero new backend machinery

- **Ty::Fn gains `is_effect`** (serde-defaulted; ~80 construction/pattern
  sites swept). The bit lives on DECLARED SLOT TYPES only, where `ret` is the
  unwrapped B; named-fn VALUES keep the carrier baked into `ret` exactly as
  before (the convention is documented at each site — mixing the two produced
  a double-wrapped `Result<Result<..>>` regression the spec suite caught).
- **Checker**: a lambda in an effect slot is typed as the carrier
  `(A) -> Result[B, String]` (a pure value tail rides the existing Phase 1b
  ok-wrap); the slot accepts BOTH the pure and the fallible spelling
  (solver-level carrier acceptance, tried before the plain ret unify so a
  failed attempt cannot contaminate bindings, the #547 discipline); calling an
  effect fn VALUE yields the carrier and is E006 from a pure context.
- **Lowering**: `normalize_effect_fn_types` rewrites every effect fn TYPE to
  its carrier at lower exit — v0 codegen, both v1 renders and almide-interp
  then see exactly the shape the ADR-0009 D1 `(A) -> B!` slots already
  landed. No renderer learns a new type.
- **unify/compatible stay effect-agnostic** (inference plumbing must not fail
  on the bit); policy lives at the call-argument check and the solver's
  carrier arm.

## Ledger (same-PR rules)

- **C-221** + fixture `spec/wasm_cross/effect_slot_carrier.almd` — success,
  pure, and err-path byte-identical native ⇄ wasm (verified: `42 42 -1` on
  both legs).
- **ALS-M15** normative section (docs/specs/als/semantics.md).
- CHEATSHEET section with the http.serve example; conformance/README
  regenerated.
- `spec/lang/effect_slot_test.almd` (the err-path spelling lives in the
  fixture: a test block's unwrap rule reaches into lambda `!`, a pre-existing
  D1 edge noted in the file).

## Not in scope

Bare-slot transparency (`(A) -> B` accepting effectful lambdas without the
annotation) is ADR-0009 D2's second bit and rides #1108 2b-iii's quadrant
machinery, as ratified on the issue.

## Known-inherited red

`every_wasm_skip_is_a_declared_platform_limit` fails on the BASE commit
(spec/stdlib/fs_streaming_test.almd skips wasm without a declared reason —
the C-220 landing's follow-up, not touched here). Everything else: workspace
1983 passed / spec 360 files / contract ledger 221 with 359/359 links.


O6lvl4:
Status: the two failing checks are both INHERITED from the base and tracked in #1146 (A/B: the Trust-Spine UNBACKED breach and the wasm-skip-ledger red reproduce identically on clean `5503c0079` — the fs streaming landing). This branch adds no delta to either gate; it goes green once #1146 is resolved on develop and this rebases.

PR #1165, opened 2026-08-10
Shape 2 third batch: List[Result[Int,String]] element eq (#1134)

O6lvl4:
Third batch of the Shape 2 burn-down. One cell, one wall — walled-real ratchet **8 → 7**.

**The cell** (kills the fallible_hof marker test-fn cond): `long == [ok(1), err("…")]` over `List[Result[Int, String]]` had no element class in the eq engine. Added `list.eq_res_int` to value_core (the `list.eq_opt_int` pattern one row over: len-as-tag element blocks, tag eq + scalar-Ok i64 bit-eq / String-Err byte-eq via the module-local string compare), registered it, routed the eq classify through it (scalar-Ok/String-Err gate — a Float Ok stays out on both sides, the -0.0/NaN rule), and mirrored the classify counter (`mir == ir` by construction).

**A/B evidence**: 5-case matrix byte-identical native ⇄ wasm — ok/err mixed positive, err-message negative, ok-payload negative, length negative, all-ok positive.

**Verification**: corpus totality over 6000 fns + ratchet 7 (baseline pruned same change), `almide test` 364/364 (342 via WASM), 87 cargo suites, value_core fmt clean.

Remaining 7 are all heavier bricks, each named in the baseline: the fs streaming capability trio (`seq_counts`/`par_counts`/`collect_partition`), the Option[record] recursive-drop pair (`Inner_go` + Deep test fn), branch_lift (heap-`??` in a call argument), and custom-E (untracked match subject).


O6lvl4:
The Test WASM red was a REAL latent miscompile this PR exposed, now fixed in the second commit:

- **What CI saw**: `fallible_hof_test` trapped at `each: fallible callback short-circuits` — `wasm trap: indirect call type mismatch`. The local harness had been hiding it: a wasm runtime trap routes to the native fallback silently (reported as "1 via native fallback", not a failure), while CI's Test WASM job counts it honestly.
- **Root cause**: `lift_lambda`'s sub-context left `decl_ret_is_result = false`, so the voided-effect-tail convention (a `Result[Unit,_]` tail of a declared-Unit effect fn produces no value) misfired on a fallible callback's REAL carrier: `(n) => record_seen(n)` (each's `Result[Unit,String]`) lifted as a VOID wasm fn while the `__fallible_each` twin's `call_indirect` expects the carrier — type mismatch. The `map`-family callbacks were asymptomatic because an Int payload and a carrier handle are both i64.
- **Why only now**: the file had walled fns since its creation, so the whole file always ran the native leg; batch 3 removed the last wall, the file flipped to wasm, and the batch-2-era latent trap surfaced — the walls were shielding a miscompile, which is precisely what this burn-down series exists to flush out.
- **Fix**: seed `decl_ret_is_result` from the lambda body's Result/Option type at lift time (a lambda has no synthetic effect-Result, so its carrier type is always real).
- **Re-verified**: the repro passes both targets ok+err paths; `fallible_hof_test` now runs 1-via-WASM; full suite 364/364 (343 via WASM, up from 342); corpus totality + ratchet 7; 87 cargo suites.

PR #1167, opened 2026-08-10
Shape 3: silent custom-E to_result corruption + the branch_lift call-arg unwrap-or wall (#1134)

O6lvl4:
Probing the custom-E baseline entry (the `match r { err(Missing(k)) => … }` wall) uncovered something strictly worse hiding behind it: the **two-step spelling doesn't wall — it silently corrupts**. `let r: Result[Int, LoadError] = option.to_result(o, Missing("cfg")); match r { err(e) => match e { Missing(k) => println(k), … } }` printed **nothing** on wasm (native: `cfg`), rc 0, no error. Reproduces on the released binary — pre-existing, not from this week's batches.

**Root cause — the shim-v1 type-blind link, live**: the self-host registry serves `option.to_result` by NAME with the `(Int?, String)` body, for every scalar-payload instantiation. For `E = LoadError` its `__copy_str` reads the VARIANT error block as string bytes; the match then dispatches over the corrupt copy — the err arm's payload slot holds 0, `print_str(0)` writes nothing, silence. This is the exact failure mode `docs/roadmap/active/selfhost-link-v2.md` names as shim-v1's defect class.

**The fix** (within today's architecture, the `_h`-twin discipline):
- A scalar-A custom-E instantiation routes to a new `option.to_result_ve` twin whose err arm CO-OWNS the caller's error block (payload-type-independent handle physics; the caller's statically-typed drop route frees it recursively — the C-149 share-+1 discipline).
- A heap-A custom-E instantiation routes to an unregistered name and **walls honestly** (it was also silently corrupting; a faithful heap-A twin is the v2 mono-link work).
- String-E instantiations keep their existing routes byte-for-byte (regression-checked).

**Evidence**: new fixture `spec/wasm_cross/to_result_custom_e.almd` (inline ctor, hoisted ctor, ok path — 3 outputs byte-identical native ⇄ wasm), declared on C-149 with the contract statement extended in the same change; contract-ledger gate green (363 fixtures, links symmetric).

**Verification**: corpus totality over 6002 fns, walled-real ratchet 7 (unchanged — this kill is a silent-corruption class, not a wall), `almide test` 364/364 (343 via WASM), 87 cargo suites, interp ledger green.

The nested-pattern spelling (`err(Missing(k))` in one match) keeps its honest wall and stays in the baseline — its blocker is the untracked-subject nested-pattern match, a separate brick. This PR's claim is narrower: the same family can no longer produce wrong bytes.


O6lvl4:
Second cell stacked on this PR — the **branch_lift wall** (`assert_eq(list.len(out ?? []), 2)`, the last `??`-cluster entry): a Result-polarity heap `??` in a CALL-ARGUMENT position now ANFs through the bind machinery's tail-proven `match ok/err` rewrite (a synthetic scope-tracked temp, borrowed as the arg), with the `lifted` rollback on both decline paths per the #1164 lesson. The whole `branch_lift_clone_test` file lowers — including the tail-recursive effect-fn driver with its TCO'd match — and runs the wasm leg.

Ladder: walled-real ratchet **7 → 6** (baseline pruned same change), `almide test` 364/364 (**344 via WASM**), corpus totality over 6002 fns, 87 cargo suites.

PR #1174, opened 2026-08-10
Shape 5: fs streaming self-hosts — seq_counts and par_counts fall, ratchet 3 → 1 (#1134)

O6lvl4:
The fs streaming brick, C1 style. Walled-real ratchet **3 → 1**; `almide test` 366/366 (**348 via WASM**).

**What lands**
- `stdlib/fs_fold_lines.almd`: self-hosted `fs.fold_lines` / `fs.fold_lines_chunked` (`_msi` twins, `Map[String, Int]` acc) over `prim.read_text_file` + a byte-level line walk with `BufRead::read_line`'s EXACT semantics (strip `\n` then `\r`; final unterminated line yielded; empty file folds zero lines), and the native **byte-partition rule transcribed** (`chunk = size/max(w,1)+1`; a line is owned by the chunk containing the byte before its first byte) — the chunked partials are byte-identical to native's, including the chunk COUNT (`chunks=3` parity pinned in the A/B; parallelism is wall-clock, never an observable — C-220). Partials build via prealloc + `store_str` (no concat dependency).
- `map.upsert` gains its skv self-host (the aggregation primitive the count callback needs; v0's exact present→f(old)/absent→init contract).
- The `Result[Map[String, <scalar>], String]` wrappers get tag-aware drops — `$__drop_res_msi` / `$__drop_res_lmsi` (the `res_ilsl` pattern: statics + a program-uses gate + dual-insert seeding at both the bind and match-subject sites, `heap_elem_lists` for the bind-gate admission, `variant_drop_handles` winning the drop).
- Capability admission (FsRead) + typed routing: msi accs link, any other acc routes `_x` and walls cleanly.

**A/B evidence**: counts (x=3/y=2/w=1), `chunks=3`, and the missing-file err — byte-identical native ⇄ wasm.

**What deliberately does NOT land, and why**: `fold_lines_range`'s `_ls` twin is in the tree and proven at parity (partition invariance across 3 cut patterns), but its routing stays unflipped: its consumer (`collect_partition`) carries an in-loop `!`, and landing it trips the corpus gate's `mir>ir` check — the count-side `desugar_all` does not run `desugar_loop_unwrap`'s flag rewrite while the lowering ladder does, so the rewrite's injected `$x ++ ""` owned-copy concat is a MIR op with no counted IR node. That is the classify-chain drift previously noted as rot, now material for the first in-profile loop-`!` fn — the gate did its job; the fn stays honestly walled and the baseline documents the flip-when-fixed. Tracked in the follow-up issue.

**Verification**: corpus totality 6009 fns, ratchet 1 (baseline pruned same change), 87 cargo suites, fmt clean.


O6lvl4:
Superseded by the re-cut from the rebased head — the remote branch was stale (pre-rebase) when this PR was created.

PR #1182, opened 2026-08-10
#1108 Phase 2b-iii Cell 1: user-HOF plain-slot transparency via generated twins (ADR-0009 D2)

O6lvl4:
The first cell of the ratified 2b-iii arc (#1108): a fallible callback in a USER HOF's bare `(A) -> B` slot now compiles — by generating the twin the hand-written `list.__fallible_*` family was standing in for.

**Mechanism** (`crates/almide-frontend/src/check/fallible_user_hof.rs`, called pre-inference from `infer_program`): same name-swap discipline as the list HOFs, with the twin GENERATED as ordinary AST —

```
fn apply1(cb: (String) -> Int, s: String) -> Int = cb(s)
  ⇒ fn __fallible__apply1(cb: (String) -> Int!, s: String) -> Int! = cb(s)!
```

The slot becomes the EXPLICIT fallible form (the proven ADR-0009 D3 path — L7/L8), the return gains the same marker, every direct `cb(...)` gains `!`, the call site renames to the twin, and its signature registers through `register_fn_sig` like any parsed decl. Everything downstream — typing, lowering, both backends — is machinery that already exists; the twin body renumbers its ExprIds above the program max so the checker's id-keyed type map never aliases original and twin. Propagation stays SPELLED at the call site (`!`/`??`/match — ADR-0006), exactly like the list twins. The HOF stays **2-mode**: total callbacks keep the original fn.

**Cell 1 scope** (everything outside keeps today's E005 + the Phase 2b-iii hint): same-program non-effect HOF, not already fallible, EXACTLY ONE fn-typed param, slot bare, callback used only as a direct call's callee, positional fallible-lambda or `-> T!`-fn argument, test blocks untouched (L9).

**Evidence**
- `spec/lang/fallible_user_hof_test.almd` (runs the WASM leg): fallible instantiation ok+err, total instantiation coexists, `cb(cb(x))` double call short-circuits on first err, named `-> T!` fn routes like the lambda.
- A/B: this exact probe shape E005'd on the pre-change tree (and the released binary); now byte-identical native ⇄ wasm.
- New negative fixture `plain-slot-forwarded-callback`: a HOF that FORWARDS its callback stays E005 (broken), with the explicit-D3 spelling as its fixed side.
- Ladder: corpus totality 6029 fns, walled-real ratchet 0 (the generated twins lower — zero new walls), `almide test` 368/368 (350 via WASM), 87 cargo suites, diag lens 82/82.

Next cells (tracked on #1108): imported-module HOFs, multiple fn slots, callback forwarding (the transitive twin), effect HOFs, custom-E callbacks.


O6lvl4:
Auto-merge was disarmed for a round while the arc's differential oracle (the generated twin vs the hand-written `list.__fallible_map` over the same callback) did its job: a LOOP-bodied user HOF's twin puts `cb(..)!` in an in-loop expression-nested position whose wasm lowering **silently drops the err path** — minimized to a hand-written D3 program that diverges identically on released 0.56.0, i.e. a pre-existing wall-escape this PR did not introduce but would have paved a second road to. Filed as #1183 with the fix ladder (wall the surviving unwrap first, ANF-hoist to the proven bind-position class second).

The second commit adds the guard: a candidate whose callback is called inside a `for`/`while` DECLINES (today's honest E005 stays, pinned by the new `plain-slot-loop-body-callback` fixture — its fixed side is the bind-position D3 spelling, which runs correctly native and walls honestly on wasm). Guard lifts when #1183's capability step lands.

Re-verified ladder on the guarded tree: diag harness green, `almide test` 368/368 (350 via WASM), corpus walled-real ratchet 0. Re-arming auto-merge.

Issue #1218, opened 2026-08-11
Interp: sandboxed fs floor for the last 4 BRIDGEABLE abstains (virtual overlay, not real fs)


O6lvl4:
Stage 2 burn-down recon (2026-08-11). After fan.any / TCO / env / Option-order, the abstain ledger's BRIDGEABLE class is down to 6 = args 2 (#1217) + these 4:

| fixture | first missing prim |
|---|---|
| fan_settle_tuple | prim.read_text_file |
| fs_preopen_resolve | prim.write_text_file |
| fs_relative_path | prim.make_dir |
| host_floor_string_alloc | prim.make_dir |

The full prim vocabulary the fs fixtures reach: read_text_file, write_text_file, make_dir, path_exists, path_filestat, read_dir, remove_all, read_bytes_file (+ heap prims already classed HEAP_BOUNDARY). Native contracts are simple (`runtime/rs/src/fs.rs`: std::fs + `io_err` = Display; Result[String,String] / Result[Unit,String] value shapes).

## The design decision (why this is not a mechanical floor)

The two backend legs run as CHILD PROCESSES (harness tempdir; wasmtime `--dir=/`), so their fs effects are isolated per run. The interp runs IN-PROCESS — and `cargo test` runs `interp_abstain_ledger` and `interp_cross_target_spec` (and shards) in parallel threads. A real-`std::fs` floor would have two threads racing the same fixture's writes in one shared cwd; `chdir` is process-global and thread-unsafe.

Recommended: an **in-memory virtual fs overlay per Interpreter** — writes/mkdir/remove land in the overlay; reads/stat/exists consult the overlay first and fall back to the REAL fs read-only (fixtures that stat host paths keep working); deterministic, race-free, and the "sandboxed fs" the abstain-classes header already names. Fixtures observe only their own writes plus host reads, which the overlay serves exactly.

## Acceptance

- the 4 fixtures leave interp-abstain-ledger.txt (shrink-only gate)
- interp_cross_target_spec green — all 4 vote WITH the native==wasm consensus
- no real-fs writes from the interp leg (greppable: no std::fs::write/create_dir in the floor)

O6lvl4:
Implemented and merged in #1221 (+ the #1223 rebase re-measurement). First BRIDGEABLE-zero row:

```
corpus 369, voting 253 (68.6%), abstaining 116 (HEAP_BOUNDARY=116, BRIDGEABLE=0)
```

[...]

Issue #1217, opened 2026-08-11
Interp: bundled-module tier for args.* (last 2 non-fs BRIDGEABLE abstains)


O6lvl4:
Stage 2 burn-down recon (2026-08-11). `args_surface` and `bundled_pure_modules` abstain at `capability: args.positional` / `args.flag` — the only remaining non-fs BRIDGEABLE entries after the fan.any / TCO / env / Option-order rounds (voting 248/367, BRIDGEABLE 6 = these 2 + fs 4).

## Why it is not a one-liner

`args` is a bundled pure-Almide module over an already-floored prim (`env.args` → `prim.args_get_list`), so a pool tier keyed `("args","flag") → body` looks sufficient — scaffolding for exactly that exists and is sound (module-keyed publics so bare names can neither shadow nor be shadowed, #1087 class; `__` helpers on the flat table, #868). But `SRC_ARGS` **declines to lower under the pool recipe** (`canonicalize(.., iter::empty())`):

1. `undefined variable env` — the source does not `import env`; the splice context injects it.
2. With `import env` prepended: `cannot call effect function env.args from a pure function` — `fn raw() = env.args()` relies on the splice context's effect exemption, and the checker sets `can_call_effect = is_effect` per decl (module_inference.rs:333), so there is no env-level loosening to borrow.

The production recipe that DOES lower these bodies lives in the codegen passes (`pass_stdlib_lowering.rs`, `pass_resolve_calls.rs` — bundled sources parsed via `parse_cached`, spliced with the module context that resolves cross-stdlib refs and effect capability). The right fix is to reuse/extract THAT recipe for the interp pool rather than teach the pool checker ad-hoc exemptions — one instrument, or the two lowerings drift.

## Acceptance

- both fixtures leave `interp-abstain-ledger.txt` (gate enforces shrink-only)
- `interp_cross_target_spec` green — the two vote WITH the native==wasm consensus
- no new flat-name resolution (module-keyed tier only)

PR #1221, opened 2026-08-11
Interp sandboxed fs floor (#1218) — three more fixtures vote, BRIDGEABLE 6 → 2

O6lvl4:
## What

The fs half of the BRIDGEABLE burn-down, on the design filed in #1218: a per-interpreter **in-memory overlay** — writes (`prim.write_text_file` / `make_dir` / `remove_all`) land in the overlay and never on disk; reads (`read_text_file` / `path_exists`) consult the overlay first and fall back to the REAL filesystem read-only. The backend legs run as child processes with per-run isolation; the interp runs in-process where `cargo test` parallelism shares one cwd, so a real-`std::fs` floor would race — the overlay is deterministic, race-free, and grep-provably write-free.

- Error strings mirror the native `io_err` (= Display of `std::io::Error`); overlay-synthesized errors use the exact `(os error N)` spellings, identical on linux and macos.
- `remove_all` on a path that exists only on the HOST abstains rather than pretending — refusing to delete real files is the sandbox's point.
- **Overlay keys normalize `.` segments** — C-042's fixture writes `/tmp/x` and reads `/tmp/./x`, the identity both backends pin. `..` deliberately stays unresolved (no contract pins it).

## The oracle caught the first draft — again

Without normalization the `./`-spelled read missed the overlay, fell back to the real fs, and answered with a LEFTOVER file from the native leg's earlier run (`PREOPEN-ROUND-2` where consensus says `PREOPEN-RESOLVE-OK`). The voting gate flagged the dissent before anything shipped. That is two drafts today the 3-way gate has rejected on its own glue.

## Measured

- `interp_cross_target_spec` **green** — fan_settle_tuple, fs_preopen_resolve, fs_relative_path vote WITH the native==wasm consensus: 248 → **251 voting (68.4%)**
- host_floor_string_alloc honestly reclassifies to `prim.handle` (HEAP_BOUNDARY 113 → 114) — the interp-heap arc's ledger, not this one
- BRIDGEABLE **6 → 2** (the args pair, #1217, is all that remains)
- `almide test` 370/370 (352 WASM / 18 fallback)


O6lvl4:
Rebased over #1223 (the args pair) — the generated ledger conflicted and was **re-measured, not hand-merged**. Combined result, first measurement with an empty bridgeable class:

```
corpus 369, voting 253 (68.6%), abstaining 116 (HEAP_BOUNDARY=116, BRIDGEABLE=0)
```

Every remaining abstain is the interp-heap arc. Voting gate re-run green on the combined tree (zero dissent).

Issue #1229, opened 2026-08-11
wasm string-accumulator TCO is quadratic: the exact-size free-list never reuses a growing string (fuzz run 31486558420)


O6lvl4:
The 0.57.0 release-gate fuzz run (manually dispatched, develop tip, 4/4 shards) recorded its one finding here: a mutation of `spec/wasm_cross/tco_multi_concat.almd` with `build(65535, 0, "")` was classified **Hang__wasm_run_hung** (seed 503784934721, index 938).

## Triage: NOT a hang, NOT a correctness divergence — a quadratic perf cliff

Measured locally (develop tip):

| n | native (built binary) | wasm (wasmtime) | output |
|---|---|---|---|
| 4096 | instant | 0.05 s | identical |
| 16384 | instant | 1.03 s | identical |
| 65535 | instant | **21.7 s — completes, byte-identical** | identical ✓ |

The fuzzer's per-run timeout read 21.7 s as a hang. Bytes and termination are correct on both legs; C-104 is not violated.

## Root cause

The TCO'd accumulator loop lowers `acc + c0 + c1` as drop-old/alloc-new per iteration (try_lower_concat_str's loop-carried slot). On native this rides Rust's `String` with amortized capacity doubling → **O(n)**. On wasm every iteration frees a size-k block and allocates size-k+2 — and `$alloc`'s free-list reuses only EXACT sizes (proofs/FreeList.v's contract), so a monotonically growing string never reuses anything: full copy every iteration → **O(n²) bytes copied** (~8.6 GB at n=65535).

## Fix directions

1. **Amortized capacity for the loop-carried ConcatStr slot**: when the drop-old/alloc-new pattern targets the SAME slot each iteration, allocate with doubling capacity (cap field already exists in the canonical String layout) and append in place while `len + add <= cap`. Turns the loop linear; the free-list contract untouched.
2. Narrower: a size-class free-list (reuse >= requested) — broader blast radius on FreeList.v; not preferred.

Also worth folding in: the fuzzer could re-run a suspected Hang with a longer budget once before classifying (a "SLOW" class distinct from "Hang") — 21.7 s vs infinite is exactly the distinction a release gate wants.

## Release impact

Correctness-clean, so this does NOT block 0.57.0 by the wrong-bytes / nontermination criteria — but wasm-first says a 400x same-program gap is real debt; it stays open under the perf ledger (`research/benchmark/perf/`) discipline.

Issue #1210, opened 2026-08-11
Codegen emits invalid Rust for `Bytes? ?? fallback` inline in a call argument (E0308, Bytes only)


O6lvl4:
Found while authoring the bytes semantics manifest (the measurement program hit it on first run). Present on released 0.56.0 and develop.

## Repro
```almide
import bytes
fn pick(flag: Bool) -> Bytes? = if flag then some(bytes.from_list([7])) else none
fn main() -> Unit = println("${bytes.len(pick(false) ?? bytes.new(0))}")
```
```
error[E0308]: mismatched types
  ...bytes_len(&match t { Some(v) => v, None => RcCow::from(almide_rt_bytes_new(0i64)), })
                                   ^ expected `Vec<u8>`, found `RcCow<Vec<u8>>`
```

## Boundary (measured)
| shape | result |
|---|---|
| `bytes.len(o ?? bytes.new(0))` — inline in a call arg | **invalid Rust** (E0308) |
| `let b = o ?? bytes.new(0)` then use `b` | ok |
| explicit `match` inline | ok |
| same shape with `List[Int]?` / `String?` / `Map?` fallback | ok — **Bytes only** |

The `??` lowering wraps the None arm in `RcCow::from(...)` while the Some arm unwraps bare, and the surrounding bytes intrinsic call takes `&Vec<u8>` — the two arms disagree about the RcCow layer exactly when the expression is consumed directly by a call. Honest failure class (compile error, not accept-and-wrong), but it rejects a completely idiomatic spelling — `option.unwrap_or` over Bytes is everywhere in real code and the manifest probes had to be rewritten around it.

Issue #1204, opened 2026-08-11
COMPILER PANIC: `guard let` inside a for/while body — "internal error: entered unreachable code"


O6lvl4:
Found while building a byte-identity fixture for `if let` / `guard let` (they had spec tests but no `spec/wasm_cross` coverage). The very first heap+loop shape I wrote crashed the compiler.

## Repro

```almide
effect fn main() -> Unit = {
  for row in [["p"], []] {
    guard let h = list.first(row) else { continue }
    println(h)
  }
}
```

```
thread 'almide-main' panicked at crates/almide-frontend/src/lower/statements.rs:67:13:
internal error: entered unreachable code: guard let is desugared by the enclosing block, not lower_stmt
```

Identical on **released 0.56.0** and on develop — not a regression, a long-standing hole.

## Measured boundary

| shape | verdict |
|---|---|
| `guard let` in a **`for` body** | **PANIC** |
| `guard let` in a **`while` body** | **PANIC** |
| `guard let` in an `if`-then block | ok |
| `guard let` in a `match`-arm block | ok |
| `guard let` at fn-body top level | ok |
| `if let` in a `for` body | ok (unaffected) |

So the desugar that turns `guard let` into its bind + early-exit form runs for BLOCK positions but never for a loop BODY's statement list, and `lower_stmt` then hits the `unreachable!()` that assumes the enclosing block already handled it.

## Why it went unnoticed

`spec/lang/guard_let_test.almd` has 6 tests — Some/None, Ok/Err, nested guards — all at fn-body top level. Nothing puts a `guard let` in a loop, which is arguably its most natural use (`guard let` exists for early exit, and `continue` is the loop's early exit). The `spec/wasm_cross` byte corpus has **zero** `guard let` fixtures.

## Severity

A panic is the loudest possible failure, so this is not a silent-wrongness class — but for a language positioning itself for mission-critical use, an `internal error: entered unreachable code` on a plain composition of two documented constructs is a correctness-of-the-tool defect: the user did nothing wrong, and the diagnostic tells them nothing actionable.

## Fix direction

[...]

Issue #1197, opened 2026-08-11
wasm ⇄ native byte divergence LIVE: matrix softmax/pow/sigmoid use the PLATFORM libm on native while wasm uses the vendored musl-libm


O6lvl4:
Found by the mission-critical arc's libm audit (the sweep that followed #1196's interp vendored-libm bridge). **This is an accept-and-wrong cross-target divergence that is live on develop and on released 0.56.0** — both legs compile, both run, the outputs differ.

## Repro (identical on develop and on released 0.56.0)

```almide
effect fn main() -> Unit = {
  let m = matrix.from_lists([[0.7, 1.3, 2.1], [0.4, 1.9, 0.11], [3.7, 0.03, 1.55]])
  println("${matrix.to_lists(matrix.softmax_rows(m))}")
  println("${matrix.to_lists(matrix.pow(m, 1.7))}")
}
```

| op | native | wasm |
|---|---|---|
| `softmax_rows`[0][0] | `0.1454055037778682` | `0.1454055037792031` |
| `softmax_rows`[2][1] | `0.022309411914874525` | `0.0223094118380793` |
| `pow`[0][1] | `1.5620815356808382` | `1.5620815356808384` |

The softmax divergence is ~1e-12 RELATIVE — far beyond a last-ULP wobble, because the per-element `exp` differences amplify through the normalization sum.

## Cause

The whole point of vendoring musl-libm (`runtime/rs/src/libm.rs`, the StrictMath/fdlibm decision) is that `math.sin/exp/log/pow` must be bit-identical across targets AND deterministic across platforms. The MATRIX runtime never got the memo — it calls the PLATFORM libm directly, while the wasm leg's self-host uses the vendored algorithm:

| site | call | reachable as |
|---|---|---|
| `runtime/rs/src/matrix.rs:551` | `(*x - max).exp()` | `matrix.softmax_rows` |
| `runtime/rs/src/matrix.rs:278` | `x.powf(exp)` | `matrix.pow` |
| `runtime/rs/src/matrix.rs:440` | `(-g).exp()` | `matrix.swiglu_gate` |
| `runtime/rs/src/matrix_p2.rs:381` | `(-x).exp()` | the sigmoid in the fused path |
| `runtime/rs/src/matrix_p2.rs:200` | `theta_base.powf(exp)` | rope inv_freqs |
| `runtime/rs/src/matrix_p2.rs:692` | `theta_base.powf(exp)` | rope inv_freqs (second copy) |

[...]

PR #1236, opened 2026-08-11
Bump version to 0.57.0

O6lvl4:
Release prep per the documented procedure. Highlights since 0.56.0 (all merged on develop):

**Fixed (live in 0.56.0):**
- #1204 — COMPILER PANIC: `guard let` in a for/while body
- #1210 — `Bytes?/Matrix? ??` inline in a call argument emitted invalid Rust
- #1197 — matrix transcendentals diverged native ⇄ wasm (up to four answers by arch × target); canonical unfused fast-exp on both legs
- #1183 — wasm read the err payload slot for an expression-nested `!` (accept-and-wrong)
- $itoa_append printed negative i64 as u64 (certificate-layer prelude)

**Assurance (the mission-critical arc):**
- Stage 1c: hand-written WAT prelude fully classified + re-measured in CI; dead fns deleted; bodies executed with pinned stdout
- Stage 2: interp oracle BRIDGEABLE 32 → 0 — voting 253/369 (68.6%); trampoline TCO, env/fs/args floors, fan.any
- bytes semantics manifest: 121 measured rows; C-222/C-223/C-224 contracts + fixtures
- fuzz nightly instrument repaired twice (#1201 verdict-skip, #1225 workflow parse)

**Known limitation (not correctness):** #1229 — wasm string-accumulator TCO is quadratic at scale (byte-identical output; 21.7 s at n=65535 where native is instant).


O6lvl4:
CI caught a real drift: the bump alone fails `docs_gen_check_passes_on_clean_checkout` — `llms.txt` is a version source-of-truth (`check_version`: the Cargo version must appear in it). Added the v0.57.0 release entry in the file's own style (fixes a–e + the assurance arc + the #1229 known limitation) and demoted 0.56.0 to Previous. `docs-gen --check` and both docs_gen tests green locally. (The E015 warnings in the failing log were pre-existing stderr noise from the iface extraction, not the failure.)

Issue #1264, opened 2026-08-12
Unknown string escapes and out-of-range unicode escapes pass through silently as literal text


O6lvl4:
Measured on 0.57.0 (2026-08-12), Stage-1 (accept-and-wrong) class:

```almide
println("bad:\q")          // prints  bad:\q      — no diagnostic
println("bad:\u{110000}")  // prints  bad:\u{110000}  — beyond U+10FFFF, no diagnostic
```

An unknown escape (`\q`) and an invalid `\u{…}` codepoint (> U+10FFFF) are both ACCEPTED and emitted as literal text. Valid escapes (`\t` `\n` `\"` `\\` `\u{3042}`) interpret normally, so the language has escape semantics — it just does not reject the malformed cases.

Why this is the accept-and-wrong class: an LLM writing `"C:\temp\new"` expecting Windows-path literalness gets `\t`→tab and `\n`→newline interpreted but `\q` passed through — a mixed, silent, hard-to-notice corruption. The E024 precedent (out-of-range int literal = check-time error, "never silent wrap", ALS-E1) argues the same treatment here: unknown escape = check-time error with a hint (`did you mean \\q?`), invalid codepoint = check-time error.

Design fork (defer to ○×):
- (a) check-time error for unknown escapes AND invalid codepoints (symmetric with E024; breaking for any code relying on pass-through)
- (b) error only for invalid `\u{…}` (malformed unicode is never intentional), warn on unknown escapes
- (c) keep pass-through, document it normatively in ALS-E5 (status quo becomes the contract)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

O6lvl4:
Fixed in #1349 (landed on develop as `9d1b3a23c`), as **E047** — E046 went to the call-argument placeholder (#1346) in a parallel branch, so this one was renumbered.

```
$ almide check v.almd
error[E047]: unknown escape sequence `\q` in a string literal
  --> v.almd:1:25
  in string literal
```

Both shapes are now check-time errors with distinct messages and a caret on the escape; `"bad:\u{110000}"` (out of Unicode scalar range) is the second. Decoding is unchanged, raw strings are skipped, and literals nested inside `${…}` holes are validated too.

**The corpus sweep found a live bug**: `tools/stdlib-crawler/main.almd` joined records on `"\0"` — two characters, because `\0` was never an escape — while the receiving Python split on a real NUL. Fixed to `"\u{0}"` in the same PR.

Issue #1263, opened 2026-08-12
fmt collapses heredocs into single-line quoted strings, erasing the documented idiom


O6lvl4:
Measured on 0.57.0 (2026-08-12), same family as #1261 (fmt reprints literals from their VALUE, destroying the source form):

```almide
let h = """
  line1 "quoted" no-escape \n literal
  line2
  """
```

after `almide fmt` becomes:

```almide
let h = '   line1 "quoted" no-escape \nliteral\n   line2'
```

Three problems:

1. **The construct is erased.** CLAUDE.md and docs/CHEATSHEET.md actively recommend heredoc for static text blocks ("no escapes, reads like actual code") — but any fmt-gated tree (spec/, CI `fmt --check`) forcibly rewrites every heredoc into the one-line escaped form. The documented idiom and the formatter contradict each other.
2. **ALS-E5 fixture blocker (same mechanism as ALS-E3/#1261).** A `spec/wasm_cross` fixture cannot pin the heredoc form, the `\u{...}` escape spelling (fmt substitutes the raw character), or delimiter choice (fmt switches `"` ⇄ `'` to minimize escapes) — they do not survive the fmt gate. The string-literal coverage row can only section what fmt preserves (`\t` `\n` `\\` spellings survive; values survive).
3. **A separate measured surprise inside heredocs**: `\n` IS interpreted inside `"""…"""` (the probe printed a real newline mid-line). The CHEATSHEET wording "no escapes" reads as promising the opposite. Whichever way the ruling goes, the normative text needs to say which one is true.

Design fork (defer to ○×, one question per concrete artifact per the house rule):
- (a) fmt preserves literal source text verbatim (fixes this + #1261 in one move; formatter only touches layout around literals)
- (b) keep value-normalization as canonical, then: update CLAUDE.md/CHEATSHEET to stop recommending heredoc, and accept that ALS form-pinning for strings/floats is permanently fixture-blocked
- (c) hybrid: preserve multi-line/heredoc + escape spellings, still normalize delimiter choice

🤖 Generated with [Claude Code](https://claude.com/claude-code)

O6lvl4:
Fixed in #1349 (landed on develop as `0fd5bf0df`). Verified against the develop tip binary — a heredoc survives fmt as a heredoc, body, indentation and delimiters intact:

```
  let s = """
    line one
    line two
    """
```

[...]

Issue #1261, opened 2026-08-11
fmt rewrites float literals through their f64 value — 1e999 becomes the unparseable inf.0


O6lvl4:
Found by the ALS-E3 authoring probe (measure-before-writing). Two related behaviors, one broken:

**The bug**: `almide fmt` prints float literals from their parsed f64 VALUE, and an overflowing literal saturates to infinity at parse (per the ALS-T2 value model), so:

```
$ cat a.almd
effect fn main() -> Unit = println(float.to_string(1e999))
$ almide check a.almd   # No errors found
$ almide fmt a.almd
$ cat a.almd
effect fn main() -> Unit = println(float.to_string(inf.0))
$ almide check a.almd   # error[E003]: undefined variable 'inf'
```

fmt(valid) → invalid — the fmt-round-trip class the fuzzer's FmtInstability rung hunts, but the generator never emits overflow float literals so it never saw this.

**The adjacent design question** (secondary): fmt also rewrites finite spellings through the value — `1.5e-3` → `0.0015` — semantics-preserving under exact round-trip display, but it discards the author's chosen notation.

**The fork needs a design ruling** (deliberately NOT decided here — flagged OPEN in ALS-E3):
- (a) reject overflowing float literals at check time (symmetric with the integer literal's E024) — but then no literal spelling of infinity exists at all;
- (b) keep saturation and make fmt handle non-finite values (requires either an accepted `inf` literal form or preserving the source text in the AST float node);
- (c) preserve literal source text in the AST and have fmt print it verbatim (also resolves the spelling-rewrite question).

ALS-E3 (docs/specs/als/expressions.md) normatively pins the accepted forms and defers the overflow ruling to this issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

O6lvl4:
Dependency recorded: **ALS-E3 (float literal) cannot be sectioned until this resolves.** The spec/ tree is fmt-gated, and fmt value-normalizes float spellings (exponents, underscores) — so a fixture pinning the accepted literal FORMS cannot survive canonicalization (integer literals are unaffected: fmt preserves 0xFF / 1_000_000 / radix prefixes, which is why ALS-E1's fixture works). The element-coverage row for ExprKind::Float stays honestly UNWRITTEN with this issue as the blocker. Option (c) — preserving literal source text — unblocks E3, fixes the inf.0 corruption, and stops the spelling rewrites in one move.

O6lvl4:
[...]

PR #1348, opened 2026-08-13
fmt preserves literal source spellings (#1261, #1263); invalid escapes become E046 (#1264)

O6lvl4:
Two coupled changes in the literal/lexer area, one commit each. Every number below came from a command run on this branch.

## Part A (#1261 + #1263) — `fmt` prints a literal's SOURCE spelling

**Measured before** (`almide fmt --dry-run` on a file holding every form):

```
1e10       -> 10000000000.0      1_000.25   -> 1000.25
1.5e-3     -> 0.0015             "\u{3042}" -> "あ"
"\x41"     -> "A"                'single'   -> "single"
r"raw\nnot"        -> "raw\\nnot"
"he said \"hi\""   -> 'he said "hi"'   (delimiter switched)
"""line one\nline two"""  -> "line one\nline two"   (heredoc collapsed)
1e999      -> fmt REFUSES the file: `AST changed by formatting at
             $.decls[0].body.stmts[3].value.kind: "float" -> "tuple_index"`
             (it rendered the unparseable `inf.0`; the safety verifier caught it)
```

**Measured after**: the same file formats to itself, byte for byte — all 14 lines above, `1e999` included.

Shape: `Token` carries the string family's verbatim source text; `ExprKind::Float` / `String` / `InterpolatedString` gained `raw: Option<String>` (following `Int`'s existing `raw` precedent), and `fmt` prints it when present. `raw` is `#[serde(skip)]`, so `--emit-ast` output is unchanged and fmt's post-format verifier still compares VALUES, not spellings. Values are untouched — this is a printing change only.

**Staleness guard (measured, not reasoned).** An interpolated string's `raw` spans its `${…}` holes, so a tool that rewrites inside a hole and re-renders would silently drop the rewrite. Added `ast::strip_literal_raw` and called it from the three mutate-then-render sites (`almide fix`'s comparison rewrite, the fuzzer's mutator and minimizer). A/B on `println("cmp=${int.gt(a, b)}")`:

- guard disabled: `fix` printed `Rewrote 1 comparison function call(s)` and wrote back `${int.gt(a, b)}` — the fix vanished
- guard enabled: `${a > b}`

Pinned by `tests/fix_test.rs::fix_rewrite_inside_an_interpolation_hole_reaches_the_file`.

**Reformat**: none needed. `almide fmt --check spec/ examples/` reports `900 file(s) already formatted` before and after — an already-formatted file is at fmt's fixed point in both directions. Two files were hand-restored where fmt had previously inlined an escape into a raw byte:

- `spec/lang/string_escape_test.almd` — was committed with a literal ESC byte and a bare `あ` where its own test names say `\x1b` / `\u{3042}`
- `examples/raytracer.almd` — 3 ANSI sequences held raw ESC bytes; restored to `\x1b[`. `almide run` output is byte-identical (diffed).

`crates/almide-tools/src/fmt_tests.rs` gains the round-trip TABLE (21 spellings) plus idempotence, heredoc-shape, verifier and strip-fallback tests. A new literal form means a new row.

## Part B (#1264) — E046, invalid escapes are check-time errors

**Measured before**: `"bad:\q"` printed `bad:\q` and `"bad:\u{110000}"` printed `bad:\u{110000}` — no diagnostic, exit 0. Silent reinterpretation, the shape E024 forbids for integers.

**Measured after** — two distinct messages, carets on the escape:

```
error[E046]: unknown escape sequence `\q` in a string literal
  hint: valid escapes are \n \t \r \\ \" \$ \xNN \u{...} (and \' inside '...');
        write `\\` for a literal backslash

error[E046]: `\u{110000}` is not a Unicode scalar value
  hint: a \u{...} escape names a codepoint in U+0000..U+D7FF or U+E000..U+10FFFF —
        surrogates and anything above U+10FFFF have no character to denote
```

**E046** was free at the moment of writing (`grep -rho 'with_code("E[0-9]*")' crates/` topped out at E045 + E420). Decoding is unchanged; the validator is a pure observer sharing `lex_numeric_escape` with the decoder. Raw strings (`r"…"`) have no escape layer and never report. Interpolation holes are skipped as code, but nested literals inside them ARE checked (the sub-parser that re-lexes them discards its own diagnostics).

**Corpus sweep** — `almide fmt --check` over `spec/ examples/ stdlib/ tools/ exercises/ tests/ docs/ research/ spike/ grammar-lab/` (it surfaces parse errors), **1 finding, and it is a live bug**:

`tools/stdlib-crawler/main.almd:141` — `string.join(introspect_cmd, "\0")`. The receiving Python at line 97 splits on `"\x00"` (a real NUL), but `\0` is not an Almide escape, so the separator was the TWO characters `\` `0` and the split never matched. Fixed to `"\u{0}"`. Regex patterns across `spec/` were *not* findings: they already double their backslashes (`"\\d+"`).

Paperwork: `docs/diagnostics/E046.md`, an `llms.txt` row, and two fixture dirs (`e046-unknown-escape`, `e046-unicode-out-of-range`) with `meta.toml` + `broken.almd` + `fixed.almd`.

## Gates (all run on the branch tip)

| gate | result |
|---|---|
| `cargo test --release` (whole workspace) | 0 failed |
| `almide test` | `374 via WASM, 17 via native fallback, 0 failed (of 391 files)` |
| `make verify-trust` | exit 0 |
| `cargo test --release --test traversal_totality_lint` | 1 passed |
| `--test wasm_runtime_test` | 75 passed (`interp_cross_target_spec` + `interp_abstain_ledger` both ok) |
| `almide fmt --check spec/ examples/` | `900 file(s) already formatted` |
| `almide docs-gen --check` | ok |
| `--test diagnostic_coverage_test` / `--test diagnostic_harness_test` | 3 passed / 4 passed |
| `scripts/check-contracts.sh` | OK — 265 contracts, flagged=0 |
| `tools/xtarget-fuzz` (out-of-workspace) | builds clean |

No `proofs/*.toml` or abstain-ledger file was touched, so `gen-claims.sh` is not implicated.


O6lvl4:
Superseded by the E047-renumbered branch: E046 was taken by #1346 (call-argument placeholder), which merged first. Reopening as a fresh PR off the rebased branch — same work, escape diagnostic renumbered to E047 throughout (code, docs/diagnostics/E047.md, llms.txt row, README row, both fixture dirs).

Issue #586, opened 2026-06-11
Flight-grade program: tracking issue


O6lvl4:
> **状態** — 追跡・着手可
> **一行で** — A（工学）の 17 行は #865 の CLOSED（2026-09-18、PR #2304）で全て済み、B/C の 9 件も issue としては 2026-08-13 に本 issue へ畳まれて CLOSED になったが、本文のチェックボックスはどちらも未了のまま表示している。
>
> **やること**
> 1. 本文のチェックボックスを実態に合わせる（#865 は CLOSED、#577–#585 も CLOSED — その内容は下の 2026-08-13 コメントと `docs/roadmap/active/flight-organization.md` にある）。
> 2. `flight-grade` label に open で残る工学項目 #1479 と #1480 を A ブロックの行として載せる（今の本文には行が無い）。
> 3. critical path の行が指す #582 / #577 は closed なので、参照先を roadmap 側の C1 / B1 に差し替える。
>
> **完了条件** — 本文のチェックボックス集合が `gh issue list --label flight-grade --state open` の結果と一致し、A ブロックに未チェックの行が残らない（未了として残るのは B/C と #1479 / #1480 だけになる）。
> **根拠** — 2026-09-21 に確認: #865 CLOSED（2026-09-18、PR #2304 が cross-reference）、#577–#585 はすべて CLOSED、open な `flight-grade` は #1479 / #1480 / 本 issue の 3 件。`docs/roadmap/active/certification-grade.md` と `flight-organization.md` はツリーに存在する。
> **触る場所** — この issue 本文のチェックボックス、`docs/roadmap/active/certification-grade.md`、`docs/roadmap/active/flight-organization.md`、`docs/roadmap/active/flight-evidence-gaps.md`

---

Umbrella for the `flight-grade` label: the ladder from today's verification machinery to being selectable by certification applicants. Items marked done at creation time (Known Problems ledger, deterministic builds/byte-identity) have no issue.

## A — Engineering
- [x] #563 ALS: implementation-independent normative spec (CG-1)
- [x] #564 Promote almide-interp to normative semantics (CG-1)
- [x] #565 Spec-contract-fixture three-layer traceability (CG-1)
- [x] #566 Compiler structural coverage + MC/DC for safety passes (CG-2)
- [x] #567 Almide/Critical profile (CG-3)
- [x] #568 Static memory mode + partitioned-runtime compatibility
- [x] #569 WCET analyzability for Critical profile
[...]

Issue #585, opened 2026-06-11
Publish a position on certifying machine-written code (AI roadmap engagement)


O6lvl4:
**Category**: C — track record | **Status**: ❌ not started

Assurance frameworks assume human authors; regulators (EASA AI roadmap, FAA counterparts) are actively working out what AI-assisted development means. A toolchain born machine-writer-first with mechanical evidence is the answer that side of the table does not yet have — and being part of that conversation early is how a newcomer gets chosen rather than tolerated.

**What**: A position paper: how spec-keyed contracts, per-build certificates, and the evidence ladder discharge development-assurance expectations when the author is a model; engage the relevant working groups with it.

**Exit criteria**: paper published; at least one regulator/industry working-group interaction on record.

Market item; tracked here for program completeness.

O6lvl4:
Folded into the umbrella #586 (see https://github.com/almide/almide/issues/586#issuecomment-5278903085) and into `docs/roadmap/active/flight-organization.md`.

This issue said it itself: *"tracked here for program completeness"* — it is a legal/commercial/policy item, not engineering work someone can pick up in this repo. Nine of them in the queue made the queue lie about how much of it is actionable. **Nothing is claimed to be done**: the What, the Exit criteria and the status marker all travelled verbatim, and #586's checklist remains the tracker. Reopen if it becomes actionable engineering work.

Issue #577, opened 2026-06-11
Establish a legal entity able to sign dossiers and carry liability


O6lvl4:
**Category**: B — organization | **Status**: ❌ not started

Certification authorities and primes contract with entities, not repositories: audit responses, liability, and dossier signatures require a legal person. The agent-assisted workflow can produce the evidence; only an entity can sign it. Ferrocene and AdaCore exist as companies for exactly this reason.

**What**: Decide the corporate form and jurisdiction; define what the entity warrants (toolchain behavior per dossier) vs. what stays open source.

**Exit criteria**: an entity exists that can sign a release dossier and enter a support contract.

Not an engineering item; tracked here for program completeness.

O6lvl4:
Folded into the umbrella #586 (see https://github.com/almide/almide/issues/586#issuecomment-5278903085) and into `docs/roadmap/active/flight-organization.md`.

This issue said it itself: *"tracked here for program completeness"* — it is a legal/commercial/policy item, not engineering work someone can pick up in this repo. Nine of them in the queue made the queue lie about how much of it is actionable. **Nothing is claimed to be done**: the What, the Exit criteria and the status marker all travelled verbatim, and #586's checklist remains the tracker. Reopen if it becomes actionable engineering work.

PR #1351, opened 2026-08-13
Fold the nine non-engineering flight-grade items into a roadmap document

O6lvl4:
Issues #577–#585 are the flight-grade program's **B — Organization** and **C — Track record** sections: a legal entity, a DER engagement plan, an LTS policy, a funding sequence, a first design partner, service-history record-keeping, a position paper. Every one of them says so in its own body — *"Not an engineering item; tracked here for program completeness"* / *"Market item; tracked here for program completeness"*.

Nine such issues in a queue of engineering work make the queue lie about how much of it is actionable. This moves their content — What, Exit criteria, status marker, refs — into `docs/roadmap/active/flight-organization.md`, verbatim in substance, and links it from `certification-grade.md`. The umbrella issue #586 keeps its checklist, so nothing loses its tracking.

Nothing is claimed to be done. The status markers travel with the text (eight ❌ not started, one ◑ strategy exists), and the critical path is reproduced unchanged.


O6lvl4:
Superseded by #1367, which carries the same commit.

This PR was docs-only, and `ci.yml` excludes `docs/**` while all 13 required status checks belong to that workflow — so it reported **0 of 13** and could never merge. It sat BLOCKED with zero checks for the whole session, which is the evidence that gap needs its own fix.

Issue #1368, opened 2026-08-13
wasm reads a DIRECTORY as ok("") — $read_text_file drops fd_read's errno (native errs)


O6lvl4:
`$read_text_file` (crates/almide-mir/src/render_wasm_fs_wat.rs) **drops** `fd_read`'s errno:

```wat
(drop (call $fd_read (local.get $fd) (local.get $iov) (i32.const 1) (local.get $nread)))
```

`path_open` on a DIRECTORY succeeds, and the failure only shows up on `fd_read`
(WASI `ISDIR` = 31). With the errno dropped the function falls straight through
to the Ok arm and builds `ok("")` from a zero-length read. Native
(`std::fs::read_to_string`) returns `Err("Is a directory (os error 21)")`.

The Err text is *already* in the data segment (`FS_ERR_ISDIR_ADDR`) and the
`path_open` arm already maps errno 31 to it — only the `fd_read` errno never
reaches that mapping.

This is a SILENT wrong value, not a wall: the wasm leg compiles the program,
runs it, and returns the wrong `Result` branch.

## Reproducer

```almide
import fs

effect fn main() -> Unit = {
  let d = fs.create_temp_dir("dirprobe")!
  match fs.read_text(d) {
    ok(_) => println("OK-read_text"),
    err(e) => println("ERR-read_text: ${e}"),
  }
}
```

```
$ almide run dirprobe.almd
ERR-read_text: Is a directory (os error 21)

$ almide run dirprobe.almd --target wasm
OK-read_text
```

## Blast radius

`$read_text_file` is the single WASI read floor behind `prim.read_text_file`
AND `prim.read_bytes_file`, so every one of these diverges the same way on a
directory-at-path:

`fs.read_text`, `fs.read_lines`, `fs.read_bytes`, `fs.read_bytes_raw`,
`fs.read_text_if_exists`, `fs.read_lines_if_exists`, `fs.read_bytes_if_exists`,
`fs.read_bytes_raw_if_exists`, and everything composed on top
(`fs.fold_lines*`, `fs.copy`, …).

## Why it survived

`spec/stdlib/fs_if_exists_test.almd` — the file whose "a directory at the path
is a real error" case would have caught it — carries a hand-written
`// wasm:skip`, so the wasm leg never ran it (it is one of the rows in the
`proofs/wasm-fallback-baseline.txt` ratchet, filed under #1233 as "C-215
errno-carrying read prim not landed"). Removing the skip and running the file on
[...]

Issue #1233, opened 2026-08-11
wasm fallback 18 burn-down: the per-brick ledger (defunc closure / C-215 errno / typed twins / test-mode bridges)


O6lvl4:
Umbrella for `proofs/wasm-fallback-baseline.txt` (the shrink-only ratchet, currently 18 files). Per the wasm-first directive these are DEBT, not steady state. The bricks, from the baseline's own reason strings:

| brick | files | notes |
|---|---|---|
| test-mode heap-result match | 2 (fold_lines pair) | **#1191** — direction 2 designed (decline-first pre-scan; direction 1 measured dead against `desugar_let_unwrap`) |
| bind-position record | (codec file) | **#1064** |
| defunc closure frontier (closure-capture map fn outside the lowering subset) | 1+ | no design notes yet — the bundled-generic HOF closures brick the goal memo calls 本丸 |
| C-215 errno-carrying read prim | 1+ | the prim floor exists on wasm (`$path_filestat_q` etc.); the read prim that CARRIES errno through the self-host layer never landed |
| typed self-host twins missing: `result.zip_x`, `list.sort_by_str_key_x` | 2 | register the typed twins in `self_host_registry.rs` — the mechanical recipe every other twin followed |
| non-msi accumulator twins + `for_each_line` | 1+ | streaming-lines arc leftovers (#1143/#1144 family) |
| test mode: MUTABLE module top-lets need the per-test region bridge | 1+ | test-runner lowering, not renderer |

Rules of engagement (unchanged): every landing edits the baseline in the SAME PR (shrink-only), byte-identity A/B against native, and a brick that cannot land keeps its honest wall — never a silent fallback.

O6lvl4:
Measured tonight (PR chain #1250 + this one): the chunked file's wall is THREE bricks, now precisely named — (1) assert twins **closed** (layout-routed, 4-cell executable matrix), (2) the `_i` acc twin **landed and probe-verified**, (3) remaining: the **ladder-subject call position bypasses the acc-class router** (plain-name emission — a lowering gap, not a missing twin) and the **bare named-fn callback** declines the HOF lift (the #1108 fn-value class). Fix (3a) first: it also blocks every future twin from being exercisable under `!` in test mode.

O6lvl4:
[...]

PR #1382, opened 2026-08-13
Burn the wasm fallback ledger 17 -> 12 and fix the directory-read errno divergence

O6lvl4:
Burns down the #1233 wasm-fallback ledger from **17 files to 12**, one brick per
commit, with the shrink-only ratchet lowered in the same commit as the brick
that earned it. Every unlocked file runs the wasm leg AND asserts the same
values native does — no row was traded for a silent divergence.

## The bricks

| brick | files unlocked | what landed |
|---|---|---|
| test-mode per-test region bridge | `cross_module_cell_update_test`, `cross_module_global_state_loop_test` | `disambiguate_module_global_regions` now runs BEFORE the test-runner synthesis, so every module top-let id is program-unique and the runner's per-test re-init can name one from a main-region `Assign`. Pins **C-270**. |
| typed twin: `list.sort_by_str_key_x` | `stdlib-test` (89 tests) | `list_sort_by_str_key_rc` — the scalar twin's cached-key stable sort with `__sbr_init`'s acquire, so source and result co-own each heap element. Registered in the self-host + `coown_names` whitelists. Pins **C-271** with a 20000-round leak/double-free loop. |
| typed twin: `fs.fold_lines_chunked_x` | `fs_fold_lines_chunked_test` | `fs_fold_lines_chunked_ls` — the msi twin's byte partition with the range walker's `List[String]` accumulator. C-220 restated cell-by-cell (which accumulators wasm runs, which still wall). |
| C-215 errno-carrying read prim | `fs_if_exists_test` | see below — this one was a live miscompile, not just a gap. |

## MISCOMPILE found and fixed — #1368

`$read_text_file` **dropped** `fd_read`'s errno. `path_open` SUCCEEDS on a
directory (the failure only surfaces at `fd_read`), so the function fell through
to the Ok arm and built `ok("")` from a zero-length read:

```
$ almide run dirprobe.almd            # native
ERR-read_text: Is a directory (os error 21)
$ almide run dirprobe.almd --target wasm
OK-read_text
```

That is one shared floor behind `prim.read_text_file` AND `prim.read_bytes_file`,
so `read_text` / `read_lines` / `read_bytes` / `read_bytes_raw`, all four
`_if_exists` twins, and everything composed on them (`fs.fold_lines*`, `fs.copy`)
diverged the same way. It survived because the one test that would have caught it
carried a hand-written `wasm:skip`.

The read half now feeds its errno into the same errno → `std::io` Display mapping
`path_open` already uses, and a directory is classified from the fd's `filetype`
byte (the one `fs.is_dir` reads) because the errno a host reports for reading a
directory fd is host-specific — wasmtime says BADF, not ISDIR. New fixture
`spec/wasm_cross/fs_read_directory_errno.almd` pins the whole family plus the
regular-file / empty-file / missing-file cells so the fix cannot be "err on
everything".

## What is left, and why

The last commit replaces the four remaining `COMPILER_FRONTIER` reasons with
**measured** blockers — which function walls, the wall text, and what the brick
actually needs:

- `closure_capture_map_test` + `fs_streaming_test` share ONE brick: a computed
  closure callee whose capture set includes a mutated cell (the defunc/closure-env
  frontier). fs_streaming's three fold_lines accumulator twins are mechanical, but
  landing them still leaves its two `for_each_line`-over-a-captured-`var` tests
  walled, so that row does not move until the closure brick does.
- `codec_field_matrix_test` needs the record-literal builder for heap-container
  fields (`binds_p2_b`'s strict-mode `Init::Opaque` default-deny), not a twin.
- `result_zip_test` is not the mechanical twin recipe: the two sides carry
  different tag layouts (scalar Ok len-as-tag, heap Ok cap-as-tag @16) and the Ok
  payload is a mixed tuple whose String element needs a recursive drop route that
  does not exist for `Result[(Float, String), String]`.

## Gates

`cargo build --release` · `almide test` (**387 via WASM, 12 fallback, 0 failed of
399**; the branch point measured 379 / **17** of 396) · `almide fmt --check spec/
examples/` (919 files) · `scripts/check-contracts.sh` (271 contracts, 407
fixtures, links symmetric) · `cargo test --release --test wasm_runtime_test`
(**75 passed**, incl. `wasm_cross_target_spec` / `interp_cross_target_spec` /
`interp_abstain_ledger`) · `proofs/check-wasm-fallback.sh` ·
`proofs/check-wasm-reachability.sh` · `scripts/check-wat-prelude-audit.sh`
(reachability re-measured) · `cargo test -p almide-mir` (620 passed).

The new fs fixture reads the filesystem, so the reference interpreter abstains on
it like every other `fs_*` fixture — recorded in `interp-abstain-ledger.txt` in
this PR (both cross-target legs evaluate it; only the 3-way vote abstains).

Closes nothing in #1233 — the umbrella stays open on the four named bricks.


O6lvl4:
Independently verified this fix (built the branch head, ran both legs). It holds, and the `filetype`-not-errno call is the right one — I arrived at the same conclusion by measurement before seeing this PR: a standalone WAT probe on wasmtime 47.0.2 reports `path_open`→0, `fd_filestat_get`→0, **`fd_read`→8 (BADF)**, `nread`→0 for a directory fd. Mapping errno 8 would have been wrong; the filetype byte is right.

Verification matrix, same program both legs, and A/B'd against the released 0.57.0 binary to confirm the test actually catches the bug:

| case | native | wasm (this PR) | wasm (0.57.0) |
|---|---|---|---|
| directory at path | `ERR[Is a directory (os error 21)]` | same | **`OK[]`** |
| `_if_exists` on a directory | `ERR[Is a directory (os error 21)]` | same | **`OK[<none>]`** |
| normal file (16 B, 3 lines) | `OK[alpha\nbeta\ngamma]` | same | same |
| **empty** file | `OK[] len=0` | same | same |
| missing path | `ERR[No such file or directory (os error 2)]` | same | same |
| missing parent dir | `ERR[No such file or directory (os error 2)]` | same | same |
| permission denied (mode 000) | `ERR[Permission denied (os error 13)]` | same | same |
| not-a-directory component | `ERR[Not a directory (os error 20)]` | same | same |
| `_if_exists` absent | `OK[<none>]` | same | same |

Exactly the two directory rows changed; nothing else moved. In particular `ok("")` is still reachable for a genuinely **empty file**, so the classification is filetype-gated rather than a blanket err.

One hardening note on the same lines, not a blocker: `fd_filestat_get`'s errno is still dropped, but this PR makes that call **load-bearing** — before, the stat only supplied a size hint that `$nread` corrected afterwards; now its `filetype` byte decides the ISDIR classification. If the stat ever fails, the 64-byte `$alloc8` buffer is never written and byte@16 is read from uninitialized/recycled scratch: garbage `3` turns a readable file into `err("Is a directory")`, garbage non-`3` falls into the read path with a garbage `$fsize` from @32. Not reachable on wasmtime (measured 0 for both file and directory fds), but note `path_open` requests `rights_base = fd_read|fd_seek (6)` and does **not** request the `fd_filestat_get` right (1<<21) — a stricter WASI host would be entitled to answer `ENOTCAPABLE(76)`. Gating the filetype test on `sterr == 0` is two lines and keeps the read path as the fallback.

I also finished the dropped-errno sweep over the rest of `render_wasm_fs_wat.rs` and `render_wasm_p3.rs` (23 sites). Two findings filed separately: #1384 (`fs.list_dir` silently truncates past one 4 KiB `fd_readdir` pass — 200 entries become 130; same silent-wrong-value class, predates this PR) and #1385 (write-side floors return `"write failed"` where native gives std::io Display). Every other site is clean.

O6lvl4:
Superseded by #1399 — the same seven commits rebased onto current develop, with the sort twin's contract renumbered C-275 → C-276 after #1364 took C-275. This branch could not be force-updated from the session doing the rebase.

Issue #1366, opened 2026-08-13
almide-interp casts a WRONG third vote on `match <effect call> { ok(v) => .., err(e) => .. }` — the success value is never wrapped in the Result carrier


O6lvl4:
Found while building the fixture for #1341 (PR #1359). The 3-way oracle caught it loudly,
which is the system working — but it means a whole spelling class cannot receive a third
vote today.

## The shape

```almide
effect fn pair_sum(xs: List[Int]) -> Int = {
  let r: Result[Int, String] = match list.get(xs, 0) {
    some(a) => ok(a + 1),
    none => err("no first"),
  }
  r!
}

effect fn main() -> Unit = {
  match pair_sum([10]) {          // <- the ADR-0008 sanctioned consumption spelling
    ok(v) => println("sum ${v}"),
    err(e) => println("fail ${e}"),
  }
}
```

```
interp:  exit=1 stdout="" stderr="Error: internal: non-exhaustive match (no arm matched)"
native:  exit=0 stdout="sum 11"
wasm:    exit=0 stdout="sum 11"
```

`native == wasm` consensus, interpreter dissents — `interp_cross_target_spec`'s
BOTH-BACKENDS-WRONG banner. Here the interpreter is the one that is wrong.

## Cause

Instrumented `eval_match`'s abort to print the subject and the arm patterns:

```
subj=Int(11)  pats=["Ok { inner: Bind { var: VarId(3), ty: Int } }",
                    "Err { inner: Bind { var: VarId(4), ty: String } }"]
```

The subject is a **bare `Int(11)`**, not `Result(Ok(11))`. An `effect fn -> Int` has ABI
return type `Result[Int, String]` (`auto_wrap_abi_body`), and both backends materialize
that carrier — but the interpreter's effect convention returns the raw success value and
models the failure channel as `Flow::Return(Value::Result(Err(..)))` instead. So an `ok`
pattern can never match an effect call's result, and `try_match_variant`'s
`carrier_payload` returns `None` for every arm.

Bisected: nesting is irrelevant, and so is the `let`-bind. The trigger is exactly
`match <effect fn call> { ok(..) => .., err(..) => .. }`. A **pure** fn declared
`-> Result[T, E]` matched the same way works, which is the control.

## Why it matters

Since ADR-0008 removed implicit propagation, matching on `ok`/`err` is one of the
sanctioned ways to consume a Result (`E042`'s own hint says so: *"Matching on ok/err also
[...]

Issue #1226, opened 2026-08-11
Stage 2 endgame: model the block heap in the interp — the HEAP_BOUNDARY 116 arc (voting 68.6% → corpus-total)


O6lvl4:
The abstain ledger reached **BRIDGEABLE = 0** on 2026-08-11 (PRs #1214/#1215/#1216/#1221/#1223): every remaining interp abstention — **116 of 369 fixtures, all of HEAP_BOUNDARY** — names a heap prim (`prim.handle`, `prim.alloc_*`, `prim.load*/store*`, `prim.alloc_bytes`, …). This is the last structural gap between the executable spec and the corpus: closing it takes voting from 253 (68.6%) to the corpus total.

## Why these abstain

The interp models containers as `Value::List`/`Map`/…, not linear memory. A self-hosted stdlib body that manipulates BLOCKS (`prim.handle(x) + 12`, `prim.load64(p + 32)`, `prim.alloc_bytes(64)` — e.g. `fs_stat`, the bytes writer family, the mutable-global arena fixtures) has nothing to run against, and the bridge honestly abstains at the first heap prim.

## Shape of the fix (the "interp-heap arc")

An interpreter-local BLOCK HEAP: a growable byte arena + a handle table, with the prim floor implementing the same layout contracts the two backends share (canonical String `[rc][len@4][cap@8][bytes@12]`, DynListStr slots, cap-as-tag Result wrappers — the layouts `scalar-read-audit.toml` and the wasm renderer already document). Values crossing the boundary materialize into blocks and back (the existing `compound_repr_*_interp` fixtures pin several of these round-trips already).

Design constraints, from this arc's siblings:
- **per-Interpreter state** (like the vfs overlay, #1218) — no globals, no cross-test races
- **prim floor tier** beside argv/env/fs — the stateless bridge cannot hold a heap
- expand by **measurement**: land the arena + 2–3 prims, re-run the abstain ledger, let the reason strings name the next prim (exactly how the fan/TCO/env/fs/args rounds went)
- the voting gate is the arbiter at every step (zero-dissent rule)

Also fold in here: expanding `INTERP_BUNDLED_MODULES` (currently `["args"]`, p4_corpus.rs) when a fixture needs another bundled module — same measurement discipline.

## Ledger

`crates/almide-interp/interp-abstain-ledger.txt` (all 116 rows), classes in `interp-abstain-classes.toml`, meter rows in `research/benchmark/three-way/README.md`.

O6lvl4:
[...]

Issue #1021, opened 2026-07-31
Interp oracle: model the bytes byte-level writer family so its four fixtures stop abstaining


O6lvl4:
## What

`almide-interp` now writes back the whole-container in-place mutators
(`list.push/pop/clear`, `map.insert/delete/clear`, `string.push/clear`,
`bytes.push/clear`) by assigning to the receiver's binding — see
`crates/almide-interp/src/inplace.rs`. That recovered 10 fixtures for the
3-way oracle.

The **byte-level writer family** is deliberately still abstaining:

- `bytes.set_*` / `bytes.append_*` / `bytes.write_*` (~40 fns, matched by prefix)
- `bytes.fill` / `bytes.copy_from` / `bytes.copy_within` / `bytes.heap_restore`

These write a scalar INTO a buffer at an offset, so each carries its own bounds
rule and little-endian byte order to reproduce. A wrong third vote is worse than
an honest skip, so they abstain under their own name rather than guessing.

## Fixtures this unblocks

- `mutable_global_bytes_arena` (`bytes.set_f32_le`)
- `mutable_global_bytes_copy_from` (`bytes.set_at`)
- `mutable_global_repeat_writes` (`bytes.set_at`)
- `scalar_if_arm_effects` (`bytes.set_i32_le`)

## Approach

The interp models `Bytes` as a `List` of `Int` (`bridge.rs::bytes_fn`), so every
writer is an index-range splice on that list. The oracle for each is its
`almide_rt_bytes_*` fn in `runtime/rs/src/bytes.rs` — mirror the bounds guard
exactly (e.g. `set_at` is a silent no-op when `i >= len`, NOT an abort) and the
`to_le_bytes` encoding for the sized/float writers.

## Completeness rule

The family extends by matrix, not point-wise: the writer set is defined by the
`set_` / `append_` / `write_` prefix over `stdlib/bytes.almd`, so the gate must
assert that every prefix-matching declaration has a write-back arm — a new
family member is then covered on arrival or fails the gate.

O6lvl4:
[...]

PR #1403, opened 2026-08-13
Give the interpreter a block heap so the flat String/Bytes prim family votes

O6lvl4:
#1226 slice 1. The third oracle abstained on every self-hosted body that touched linear memory — 116 of 369 fixtures, the last structural gap between the executable spec and the corpus. This closes the flat-payload half.

## Why a handle is not like an fs path

The fs floor (#1218) works as an add-on because a path is an opaque key with no aliasing into `Value`. A handle is a **live alias**. `stdlib/base64_encode.almd` is the majority class:

```almide
let out = prim.alloc_str(chunks * 4)              // an EMPTY block
let _x = __b64_fill(prim.handle(out) + 12, ...)   // WRITES through it
out                                                // returned AFTER the writes
```

An arena that snapshotted on `prim.handle` and dropped it would return an empty String and cast a **wrong third vote** — strictly worse than the abstain it replaces, and the same failure mode as #1366 (`Hole`/`Todo` voting `Flow::Abort` against native's exit 101).

So: read materializes once and **binds**; writes go to the arena; and one **return sync** reads a block back out. The sync is driven by the fn's **declared return type**, never by guessing whether an `Int` looks like an address — a fn returning a plain Int (`string.len`) must not have its result reinterpreted as a handle.

## The bug the voting gate caught

My first version keyed the handle table on `Rc::as_ptr` alone. That is **unsound**: an `Rc` address is recycled after the last handle drops, so a later unrelated value lands on a freed pointer and inherits the earlier block. The gate found it immediately — 6 fixtures dissenting, `bytes_f16` printing `inf` for all twelve cells, `int_from_hex` answering `ok:255` for every input. Fixed by holding every bound value alive for the arena's lifetime, so no address it has answered for can be reused. **This is exactly the wrong-third-vote hazard I flagged when scoping the issue, and the gate is what caught it — not review.**

## Measured

- **heap-prim abstentions 129 → 115.** Newly voting: `base64_encode`, `encoding_base64`, `bytes_f16`, `bytes_f16_offset_overflow`, `bytes_push_inplace`, `bytes_rawptr`, `bytes_set_value_semantics`, `chunk_str`, `codec_decode_errors`, `codec_float_int`, `compound_eq`, `compound_repr_interp`, `compound_repr_records_interp`, `alias_combinator_rc`, `declaration_forms`, `datetime_format`, and the `fs_*` families.
- **`interp_cross_target_spec`: 0 dissents.**
- The remaining rows now name the next prim precisely — 48 say *"outside the flat String/Bytes family (alloc_value/alloc_map/alloc_list*/alloc_set carry tagged sub-structure — #1226 slice 2)"*, which is the honest next step rather than a generic `prim.handle`.
- `almide test` 401/401 · `cargo test --workspace --release` clean
- `almide-interp` codopsy **A (92/100), 7 warnings — exactly the baseline**. The first draft cost a point (`heap_prim` at cyclomatic 24); split one fn per family, the way `bridge.rs` splits its scalar floor.

## Deliberately still abstaining

`alloc_value` / `alloc_map` / `alloc_list*` / `alloc_set` carry tagged sub-structure rather than a flat payload. Folding them in would have made this a family the voting gate could not arbitrate cleanly, so they keep their honest abstain and their own slice.

### DIFF
--- a/crates/almide-interp/interp-abstain-classes.toml
+++ b/crates/almide-interp/interp-abstain-classes.toml
@@ -31,6 +31,12 @@ patterns = [
   "capability: prim.alloc_str",
   "capability: prim.alloc_set",
   "capability: prim.alloc_bytes",
+  # Raw pointer conversion — pointer work by definition, same class as handle.
+  "capability: prim.int_to_ptr",
+  # An access the #1226 slice-1 arena does not own. The flat String/Bytes
+  # family is modeled; a load/store landing outside it is still the heap
+  # boundary, one level in from `prim.handle`.
+  "outside this heap's arena",
 ]
 
 [[class]]
@@ -49,4 +55,16 @@ patterns = [
   "transcendental",
   "float `**`",
   "interp fuel/recursion budget exhausted",
+  # SURFACED BY #1226 slice 1, not introduced by it: these fixtures used to
+  # abstain at `prim.handle` (HEAP_BOUNDARY, first-hit) and now get past it to
+  # reveal the blocker underneath. Both are bounded dispatch work with no
+  # architectural change, which is what BRIDGEABLE means:
+  #   prim.die     — native panics with exit 101; `Flow::Abort` is this
+  #                  interpreter's exit-1 model, so voting it would be wrong
+  #                  (the #1366 lesson). Needs an abort that carries the exit
+  #                  code, not a new evaluation model.
+  #   prim.fd_write— the raw fd write. The interp already captures stdout
+  #                  (`self.stdout`), so this is glue at the argv/env/fs tier.
+  "capability: prim.die",
+  "capability: prim.fd_write",
 ]
--- a/crates/almide-interp/interp-abstain-ledger.txt
+++ b/crates/almide-interp/interp-abstain-ledger.txt
@@ -10,69 +10,66 @@
 # Preferred alternative to adding an entry: widen the interp glue
 # (bridge.rs / hofs.rs / dispatch.rs — see crates/almide-interp/CLAUDE.md).
 
-alias_combinator_rc  out-of-interp-scope capability: prim.handle
+alias_combinator_rc  out-of-interp-scope capability: prim.handle of a value outside the flat String/Bytes family (alloc_value/alloc_map/alloc_list*/alloc_set blocks carry tagged sub-structure — #1226 slice 2)
 alias_cow  out-of-interp-scope capability: prim.alloc_map
-base64_encode  out-of-interp-scope capability: prim.handle
-bytes_f16  out-of-interp-scope capability: prim.handle
-bytes_f16_offset_overflow  out-of-interp-scope capability: prim.handle
-bytes_push_inplace  out-of-interp-scope capability: prim.handle
-bytes_rawptr  out-of-interp-scope capability: prim.handle
-bytes_set_value_semantics  out-of-interp-scope capability: prim.handle
-chunk_str  out-of-interp-scope capability: prim.handle
-codec_decode_errors  out-of-interp-scope capability: prim.handle
-codec_float_int  out-of-interp-scope capability: prim.handle
+base64_encode  out-of-interp-scope capability: prim.alloc_list
+bytes_f16_offset_overflow  out-of-interp-scope capability: prim.alloc_list_f64
+bytes_rawptr  out-of-interp-scope capability: prim.int_to_ptr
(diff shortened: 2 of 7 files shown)


O6lvl4:
**Follow-up commit: classified the abstains the heap surfaced.** CI caught a gate I had not run locally (`check-abstain-classes.sh`) — four reason strings reached the classifier with no pattern, because fixtures that used to stop at `prim.handle` now get past it and report the blocker underneath.

- `prim.int_to_ptr` and *"outside this heap's arena"* → **HEAP_BOUNDARY** (pointer work by definition; an access one level in from `prim.handle`)
- `prim.die` and `prim.fd_write` → **BRIDGEABLE**. `die` needs an abort that carries native's exit **101** rather than `Flow::Abort`'s exit 1 (the #1366 lesson — voting the wrong exit code is a wrong vote); `fd_write` is glue at the existing argv/env/fs tier, since the interp already captures `self.stdout`. Both are bounded dispatch work with no architectural change, which is exactly what BRIDGEABLE means.

**Read `BRIDGEABLE 0 → 6` as fixtures moving UP a layer, not a backslide.** They were HEAP_BOUNDARY before — blocked behind the heap — and the heap work converted them into the near-horizon class. The gate has no ratchet on that count; it feeds the coverage meter.

## Same-tree coverage, before vs after

The meter's last row (2026-08-12: corpus 375, voting 255) is not a valid comparison — the corpus has since grown to 417 independently of this change. Measured on **this** tree by diffing the ledger:

| | abstaining | voting | |
|---|---:|---:|---|
| before | 130 | 287 | 68.8% |
| after | **116** | **301** | **72.2%** |

14 fixtures gained. `check-abstain-classes.sh` now reports `corpus 417, voting 301 (72.2%), abstaining 116 (HEAP_BOUNDARY=110, BRIDGEABLE=6); near-horizon target 307 (73.6%)`.

O6lvl4:
**The block heap surfaced a pre-existing wrong vote in the interpreter — fixed here.**

CI's shard 0 dissented on `nan_canonical_bytes_write`: interp `RAW-LEAK`, both backends `canonical`. It does **not** reproduce on my aarch64 machine, and that is the whole story.

## Root cause

`crates/almide-interp/src/inplace.rs::encode` wrote the **raw** float bytes:

```rust
4 => (*f as f32).to_le_bytes().to_vec(),
8 => f.to_le_bytes().to_vec(),
```

C-210 makes a byte-buffer append an **observation boundary**: raw NaN sign/payload bits are the one place the byte-identical law leaks host state, so both backends collapse NaN to a single pattern before the reinterpret — `stdlib/bytes_append_multi.almd` guards with `float.to_bits` / `__f32_obs_bits`, and `runtime/rs/src/float.rs` does the same natively. The interpreter did not.

## Why it hid

Two layers of luck, and this PR removed the first:

1. The fixture abstained at `prim.handle` before ever reaching this arm — its own header says *"the in-place byte-buffer append family is out of interp scope (#1021, ledgered)"*. The block heap let it through.
2. `0.0 / 0.0` is `0x7FF8…` on aarch64 and `0xFFF8…` on x86. On this developer machine the **raw** pattern coincidentally equals the **canonical** one, so the leak prints `canonical` and passes. Only x86 can observe it.

That is a latent wrong third vote on every NaN that reaches a byte buffer — the exact class this oracle exists to catch, caught by the oracle.

## The fix and its test

NaN is canonicalized before the width cast (`0x7FF8000000000000` / `0x7FC00000`), leaving non-NaN floats untouched and happening *before* the big-endian reversal rather than instead of it.

`a_nan_append_is_canonical_on_every_host` builds the NaN **from bits with the sign set** rather than from arithmetic, so it is host-independent. A/B against the unfixed code on aarch64 — where the original fixture cannot fail:

```
without the fix:  left: [1, 0, 0, 0, 0, 0, 248, 255]   ← raw payload leaked
                 right: [0, 0, 0, 0, 0, 0, 248, 127]   ← canonical
with the fix:    ok
```

## Re-verified

- `interp_cross_target_spec` — **0 dissents**
- `cargo test --workspace --release` clean · `almide test` 401/401
- `check-abstain-classes.sh` — `corpus 417, voting 301 (72.2%), abstaining 116 (HEAP_BOUNDARY=110, BRIDGEABLE=6)`
- `almide-interp` codopsy **A (92/100), 7 warnings — baseline**

Issue #1411, opened 2026-08-14
Failure-side corpus: spec/wasm_fail/ + host polarity + a shrink-only floor — the 9.8% problem


O6lvl4:
Tracking issue for the arc planned 2026-08-14. Full survey: `../almide-references/RESEARCH-failure-corpus.md` (8 compilers, file:line-cited).

## The problem, measured

Five defects found in one session, all one shape — **an unexecuted polarity**:

| issue | surface | what was never run |
|---|---|---|
| #1368 | a hand-written `wasm:skip` hid a miscompile | the **path** |
| #1410 | wasm swallows a declared-Option effect fn's error, yields `8260` | the **err branch** |
| #1408 | a negative offset panics natively, wasm survives | the **negative side of the input domain** |
| #1403 | a raw-NaN byte write | the **host** — invisible on aarch64 |
| #1400 | a `let`-bound range OOMs on wasm | the **scale** |

**#1410 proves this is not a gate-design problem.** Measured: `native exit=1 / wasm exit=0`, so the existing 2-way cross-target gate catches it the moment the input exists. What was missing was the *input*. C-216's fixture has a cell labelled "can-err scalar payload", a header describing this exact failure mode — and calls it only with `"42"` and `"-1"`, so `int.parse` never fails anywhere in the file.

**The number:** running every `spec/wasm_cross/*.almd` natively and counting non-zero exits or `Error:` output — **41 of 418 = 9.8%** reach an error path at all. Nothing anywhere says that should be higher. The fuzzer is currently the only generator of failing inputs, which is why *it*, and not the 418-fixture corpus, found #1410, #1408 and #1400.

## What the mature compilers do

| | compile-time failure | **runtime failure** |
|---|---:|---|
| rustc | `tests/ui` **21,086** (79.5% of all 26,515) | `run-fail` **189** + `//@ error-pattern:` |
| Zig | `compile_errors` **997** | **`test/cases/safety` 104** |
| Lean 4 | `elab_fail` **631** | — |
| Swift | 13,051, `-verify` + `expected-error{{…}}` | — |
| **Almide** | `tests/diagnostics` **92** | **no category** |

Zig's `safety/` is the model — each program deliberately triggers a runtime failure and asserts the panic **message**, passing only if the expected failure occurs; falling through to the end is itself `error.TestFailed`. rustc declares a per-test `RunFail` mode the harness enforces (`compiletest/src/common.rs:97`).

[...]

Issue #1410, opened 2026-08-14
wasm swallows the error from a declared-Option effect fn's  and yields a garbage value; native aborts (C-216's can-err cell never actually errs)


O6lvl4:
Found by the 2026-08-14 fuzz-nightly (run 31770547955, seed 508328767281 index 16, an `effect_option_explicit_bang.almd` mutation). Reproduces on current `develop`.

## Minimal repro

```almide
effect fn boom(k: String) -> Int? = {
  let n = int.parse(k)!          // fails for a non-numeric k
  if n > 0 then some(n) else none
}

effect fn main() -> Unit = {
  let used = boom("XYZ")!
  println("continued got=${int.to_string(used ?? 99)}")
}
```

| leg | result |
|---|---|
| native | `Error: invalid digit found in string` — the `!` propagates, main aborts |
| wasm | **`continued got=8260`** |

wasm neither propagates nor aborts: it continues and hands back **8260**, a value the program never computed. This is a silent wrong value, not a crash — the worst class, and worse than the panic in #1408 because nothing announces it.

Same with the binding unused (`let unused = …`) and with `let _ = …`: wasm prints `continued` either way. So it is not dead-code elimination of the effect; the call runs and its error is simply lost.

## Why the gate missed it: the can-err cell never errs

`spec/wasm_cross/effect_option_explicit_bang.almd` (C-216) has a cell it labels "can-err scalar payload", and its header describes exactly this failure mode:

> the wasm leg left the Unwrap unstripped, fed `desugar_effect_unwrap`'s err/ok match a RAW Option block (the never-err v1 ABI), and read it as a Result — a wrong value (hit=999) then an rc_dec trap

But the cell only ever calls it with inputs that **succeed**:

```almide
let ok_case = parse_pos("42")!    // int.parse succeeds
let neg = parse_pos("-1")!        // int.parse succeeds, returns none
```

`-1` exercises the `none` branch, not the error branch. `int.parse` never fails anywhere in the file, so the err path of a declared-Option effect fn has **zero coverage** — the fixture pins the shape that was fixed while leaving the shape that is broken untested. The fuzzer's mutation (`"42"` → `"ΑΒΓ"`) is the first thing to run it.

## What this needs

1. The wasm leg must propagate the err from a declared-Option effect fn's `!`, identically to native.
2. C-216's fixture gains a genuinely-erring can-err cell, so the gap cannot reopen. The existing cells stay — they cover the never-err ABI, which is a different path.

[...]

Issue #1408, opened 2026-08-14
A negative offset to any bytes accessor panics natively with a raw slice panic; wasm survives — the guard is defeated by its own overflow (38 sites)


O6lvl4:
Found by the 2026-08-14 fuzz-nightly (run 31770547955, seed 508328767282 index 602, a `matrix_select_rows_oob.almd` mutation). Reproduces on current `develop`.

## Minimal repro

```almide
effect fn main() -> Unit = {
  var b = bytes.new(8)
  bytes.set_f32_le(b, -1, 5.0)
  println("survived len=${int.to_string(bytes.len(b))}")
}
```

| leg | result |
|---|---|
| native | **panic**: `range start index 18446744073709551615 out of range for slice of length 8` |
| wasm | `survived len=8` |

Not a clean abort — a raw Rust slice panic with a backtrace note.

## Cause: the bounds guard is defeated by its own overflow

`runtime/rs/src/bytes.rs`:

```rust
pub fn almide_rt_bytes_set_f32_le(b: &mut Vec<u8>, pos: i64, val: f64) {
    let p = pos as usize;                              // -1  ->  usize::MAX
    let bytes = (val as f32).to_le_bytes();
    if p + 4 <= b.len() { b[p..p+4].copy_from_slice(&bytes); }   // MAX+4 WRAPS to 3
}
```

`pos = -1` casts to `usize::MAX`. Then `p + 4` **wraps to 3**, so `3 <= 8` is true and the guard *passes* — straight into `b[usize::MAX..]`. The check that exists to prevent this is the thing that fails.

## This is a family, not a site

**38 functions** in `runtime/rs/src/bytes.rs` carry the identical `let p = pos as usize;` shape:

`read_i16_le` `read_i16_be` `read_u16_be` `read_i32_be` `read_f32_be` `set_i16_le` `set_u16_be` `set_i16_be` `set_u32_be` `set_i32_be` `set_i64_be` `set_f32_be` `set_f64_be` `remove_at` `read_i64_be` `read_f64_be` `read_u32_be` `read_string_be` `set_f32_le` `set_u16_le` `set_u8` `set_u32_le` `set_i32_le` `set_i64_le` `set_f64_le` (+2 macro-generated) `read_bool_at` `read_string_be_at` `take_at` `read_i32_le` `read_u32_le` `read_u16_le` `read_i64_le` `read_f32_le` `read_f64_le` `read_f16_le` `read_string_at`

[...]

Issue #1400, opened 2026-08-13
A let-bound range OOMs on wasm where native iterates lazily: for-in over a large range diverges


O6lvl4:
Found by the 2026-08-13 fuzz-nightly (run 31667874852, seed 506685997633 index 522, a `spec/wasm_cross/range_first_class.almd` mutation). Reproduced against the current `develop` build; the other three findings from that night are already fixed on develop, this one is live.

## Repro

```almide
effect fn main() -> Unit = {
  let r = 0..<4294967295
  var s = 0
  for i in r { s = s + i }
  println(int.to_string(s))
}
```

| leg | result |
|---|---|
| native | `9223372030412324865`, exits 0, completes in seconds |
| wasm | `Error: out of memory` |

The fuzzer's own mutation drops the use of `s`, which makes the divergence show up as `exit code differs: native=None wasm=Some(1)`. The version above keeps the sum observable so the difference cannot be blamed on dead-code elimination — native genuinely computes the right answer without materializing anything.

## Cause

`crates/almide-mir/src/lower/binds_p2_heap.rs` (`let r = start..<end`) materializes the REAL `list.range` list on the wasm leg. That materialization is deliberate: it is #1272's fix, where a deferred-empty `Opaque` bind made wasm iterate zero times and silently print `0`. `spec/wasm_cross/range_first_class.almd` pins it, and C-238 states it.

Native does not materialize — a `let`-bound range driven through `for-in` stays a counting loop. So for a range whose length exceeds available memory, the same program succeeds on one target and fails on the other, which is what C-238 promises it will not do.

## What needs deciding

The `let`-bound range should lower to a counting loop on wasm too, the way native does, WITHOUT reopening #1272's zero-iteration bug. The range value is a `(start, end, inclusive)` triple; `for-in` over a variable holding one is the only consumer that needs the element sequence, and it can consume it lazily. Materialization would remain for the shapes that genuinely need a list (indexing, `list.*` calls on the bound name).

Until then the boundary is unstated: there is no documented range length above which the wasm leg refuses. A mission-critical target should either match native or refuse at a stated bound, not OOM.

Contract: C-238 (`ALS-E10`). Any fix updates it in the same PR.

PR #1412, opened 2026-08-14
Stand up the failure-side corpus: spec/wasm_fail asserts both legs break alike

O6lvl4:
#1411 Stage 1. Survey: `../almide-references/RESEARCH-failure-corpus.md`.

## Why

Measured 2026-08-14: **41 of 418** `wasm_cross` fixtures reach an error path (9.8%), and the three runtime bugs the fuzzer found this week (#1400, #1408, #1410) all lived on the side no fixture ran. The clarifying case is #1410: the 2-way gate is **sound** (`native exit=1 / wasm exit=0` — it catches the divergence the moment the input exists), but C-216's "can-err" cell never actually errs, so the input never existed. The gap is corpus, not gate.

Every mature compiler surveyed grew this category: Zig's `test/cases/safety/` (104 programs that pass **only if the expected panic fires** — falling through to the end is itself `error.TestFailed`), rustc's `run-fail` mode + `error-pattern` (189 tests), Lean4's `elab_fail` (631).

## What

`spec/wasm_fail/` — where `wasm_cross` asks *"do the two legs return the same value"*, this asks *"do the two legs **break the same way**"*. Separate directory per Zig's split: one judgement per directory.

Each fixture declares its failure:

```
// @expect-fail: invalid digit found in string     ← mandatory: substring of stderr
// @xf-allow: #1410 — …                            ← optional: tracked cross-leg divergence
```

The gate (`wasm_fail_corpus`, in the same shared-corpus harness as `wasm_cross_target_spec`) asserts **two halves**:

1. the native reference **actually fails as declared** — nonzero exit, stderr contains the pattern. A fixture that runs to success is a *failing test*, which makes the C-216 vacuity class (a can-err cell that never errs) unrepresentable in this directory.
2. both legs terminate **identically** (exit + stdout + stderr), with `@xf-allow` giving the same tracked/stale lifecycle as `wasm_cross`'s `@xt-allow` — a fixed divergence flags the stale directive for removal.

## Seeds

- `int_parse_err_propagates.almd` — the #1410 minimal repro, entering as the **tracked divergence** (wasm swallows the err and continues with garbage). The gate report shows it: `1 fail-identically, 1 tracked-divergence(s)`.
- `effect_err_aborts_both_legs.almd` — the control: same erring `int.parse` through a plain effect chain, which agrees on both legs today and pins the termination shape the #1410 fix must reach.

## Verified, including the negative

- **A/B on the gate itself**: planted a vacuous fixture (`@expect-fail` + a program that succeeds) → the gate fails with `the native leg did not fail as declared`; removed → green. The gate demonstrably rejects the class it was built for.
- An empty `spec/wasm_fail/` also fails loudly (the 9.8% problem must not silently return).
- `cargo test --workspace --release` clean · `rustc-warnings: 0` · fmt clean · fixtures carry no `test` blocks so `almide test` does not double-run them.

### DIFF
--- a/spec/wasm_fail/effect_err_aborts_both_legs.almd
+++ b/spec/wasm_fail/effect_err_aborts_both_legs.almd
@@ -1,0 +1,11 @@
+// @expect-fail: invalid digit found in string
+// #1411 stage 1. The CONTROL for int_parse_err_propagates: the same erring
+// int.parse, but through a plain effect chain with no declared-Option carrier
+// in the way. Both legs abort with the same message today; this file pins the
+// termination shape #1410's broken cell must be brought back to.
+effect fn parse_it(k: String) -> Int = int.parse(k)!
+
+effect fn main() -> Unit = {
+  let n = parse_it("XYZ")!
+  println("UNREACHABLE ${int.to_string(n)}")
+}
--- a/spec/wasm_fail/int_parse_err_propagates.almd
+++ b/spec/wasm_fail/int_parse_err_propagates.almd
@@ -1,0 +1,18 @@
+// @expect-fail: invalid digit found in string
+// #1411 stage 1, seeded from #1410's minimal repro. A declared-Option effect
+// fn whose body's `!` fires must propagate the err out of main — on BOTH legs,
+// with the same message and a nonzero exit. This is the cell C-216's fixture
+// labels "can-err" but never actually errs (int.parse succeeds on every input
+// it passes). The wasm leg CURRENTLY swallows this error and continues with a
+// garbage value — #1410 — so this file enters the corpus as the KNOWN
+// divergence that gate tracks until the fix lands.
+// @xf-allow: #1410 — wasm swallows the err and continues (got=8260-class)
+effect fn boom(k: String) -> Int? = {
+  let n = int.parse(k)!
+  if n > 0 then some(n) else none
+}
+
+effect fn main() -> Unit = {
+  let used = boom("XYZ")!
+  println("UNREACHABLE got=${int.to_string(used ?? 99)}")
+}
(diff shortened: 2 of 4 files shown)


O6lvl4:
**Stages 3 and 4 pushed onto this PR** (same `p6_fail_corpus.rs`, one coherent unit).

## Stage 3 — the audit found the opposite of its premise

The plan assumed the 41 error-reaching `wasm_cross` fixtures were incidental. They are not: every one is a deliberate failure test by name and assertion (`assert_abort_*`, `int_div_by_zero`, `index_bounds*`, `option_none_unwrap_term`, the `fan_*_err` family…), and the equality gate already holds them to "both legs break alike".

So they were **not moved** — 41 files of contract-evidence path churn to regain a property they already have. What they lacked was **vacuity detection**: `wasm_cross` passes when both legs agree, *including agreeing to succeed*. A change that made `int_div_by_zero` stop erroring on both legs would go green silently — the C-216 class again.

`WASM_CROSS_DECLARED_FAILING` (41 rows) + `declared_failing_fixtures_still_fail`: every listed fixture must keep failing on the native reference. Removal is a reviewed shrink, same discipline as the `wasm:skip` ledger's `GENUINE_SKIPS`.

**A/B, per this session's rule that a gate must be seen to catch its class:** neutered `int_div_by_zero`'s shielded divisor (`x - x` → `x`) →

```
int_div_by_zero: declared failing but the native leg now SUCCEEDS (exit=0 stderr="")
— the fixture went vacuous (the C-216 class).
```

→ byte-identical restore, gate green again.

## Stage 4 — the floor

`failure_side_floor`: the measured **41/418** becomes a ratio floor over (`wasm_cross` error-reaching + all of `wasm_fail`), compared by cross-multiplication so no float threshold drifts. Current reading: `43/420 failure-side (wasm_cross 41/418 + wasm_fail 2)` — the two seeds already sit above the floor.

A **ratio**, not a count: a count floor is satisfied forever by the existing 41 while the corpus grows around them, decaying the share right back toward the 9.8% that hid #1400/#1408/#1410. And **no target above the floor**: a target invites fixtures written to move a percentage.

## Independent confirmation

The floor gate's count (41/418, computed from the shared corpus's captured native legs) exactly matches the standalone shell measurement from the survey — two implementations, same number.

PR #1413, opened 2026-08-14
Run the fuzz nightly on both arches so arch-conditional bugs have an observer

O6lvl4:
#1411 Stage 2 — host polarity. Survey: `../almide-references/RESEARCH-failure-corpus.md`.

## Why this jumps ahead of the corpus inventory

#1403 (the raw-NaN byte write) was **unobservable on aarch64** — the raw pattern there happens to equal the canonical one, so every local run in that session was structurally blind to it, and only CI's x86 could see it. The inverse holds too: anything keyed on x86's sign-set NaNs is invisible to an x86-only nightly. One arch is not a subset of the other — **each sees bugs the other cannot** — and until this lands, a whole class is unobservable *today* on every machine in use.

## What

The nightly's `build` and `fuzz` jobs gain an arch axis:

| shards | runner | arch |
|---|---|---|
| 1–4 | `ubuntu-latest` | x86_64 |
| 5–8 | `ubuntu-24.04-arm` | aarch64 |

- **Shard numbers are globally unique**, so the derived seeds (`run_id * 16 + shard`) and per-shard artifact names never collide across arches, and the verdict job's `shards=k/N` keeps meaning "of the whole night". `FUZZ_SHARDS` → 8.
- Per-arch binary artifacts (`fuzz-binaries-<run>-<arch>`) and per-arch cargo cache keys.
- The wasmtime tarball is parametrized by arch — and `wasmtime-v47.0.3-aarch64-linux.tar.xz` was **verified to exist in the release assets** before wiring, since a wrong asset name would kill every ARM shard at install.
- `ubuntu-24.04-arm` is free for public repos (this repo is public), so the cost is runner-minutes, not money.

Night coverage doubles at constant wall clock (~8,000 programs/night at the measured ~210 prog/min), and the ARM half explores with different seeds rather than repeating the x86 campaign.

## Verification

- YAML parses; every remaining `x86_64` literal audited (matrix values only)
- Seed/artifact uniqueness checked across all 8 shards
- The scheduled run is the real test — after merge I will fire a short `workflow_dispatch` (`minutes=5`) to validate the ARM path end-to-end rather than waiting for tonight

### DIFF
--- a/.github/workflows/fuzz-nightly.yml
+++ b/.github/workflows/fuzz-nightly.yml
@@ -54,7 +54,7 @@ env:
   # ~210 prog/min, up from ~1,000). Deep single-seed runs stay available
   # through the `minutes` dispatch input.
   FUZZ_MINUTES: "5"
-  FUZZ_SHARDS: "4"
+  FUZZ_SHARDS: "8"
   # Cold runners gain nothing from the workspace's release-incremental
   # profile (#1006) — it only bloats the caches. Opt out.
   CARGO_INCREMENTAL: "0"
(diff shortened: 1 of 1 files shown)


O6lvl4:
**Stage 5 pushed onto this PR** (both touch `fuzz-nightly.yml`; stages 1/3/4 live in #1412).

`scripts/vacuity-mutation.sh` + a `vacuity` nightly job: mutate a spec file's input literals (spread across the file — clustered picks land on inert literals, measured: `[1,2,3]` feeding only `list.len` soundly survives element mutation) and report files whose assertions survive **every** mutant. Telemetry, never a hard red: at the budgeted K a survival can be sound, so suspects invite review rather than block the night. Rotates a fixed window by day number; smoke-run on a 4-file window worked end to end and the report reads:

```
vacuity-mutation: 401 candidate file(s); window 4 from index 4 (day 1), 3 mutant(s) each
  ok    spec/integration/codegen_functional_test.almd (3/3 mutant(s) killed)
  SUSPECT-VACUOUS spec/integration/codegen_effect_fn_test.almd — every mutant passed
```

x86 only, deliberately: the oracle is assertion-vacuity, which is arch-independent — the arch axis belongs to the differential shards this PR adds.

PR #1455, opened 2026-08-16
Close the last 33 integer-domain cells: counts, calendar signedness, and the ISO year

O6lvl4:
The three non-bytes **count** families still open in the integer-domain matrix — 19 of
the last 33 cells. Each is a count that sized an allocation without being checked.

## list.range — the subtraction IS the count

`end - start` overflows for a span like `(i64::MIN, 3)`: the count came out negative
and the allocation was sized from the wrapped value.

```
native: thread 'main' panicked … capacity overflow   (exit 101 — the ALS-T6 form)
wasm:   Error: out of memory                          (exit 1)
```

Rejecting the endpoints outright would be wrong — `range(1e9, 1e9+1)` is one element
at a large offset and must still work — so the cases split by sign: when both
endpoints share a sign the difference is bounded by `i64::MAX` and cannot overflow,
and only a span **straddling zero** can, where either half exceeding the ceiling
already settles it.

The ceiling it needed was already in its own file. `list_repeat` carries it **twelve
lines below `list_range`**, same limit, same message. Adjacency is not a mechanism.

## string.pad_start / pad_end

Sized by `n * pclen`, so the ceiling is tested by **division** — a guard computed
from the product tests the wrapped value. Without it `pad_end("a", u32::MAX, "x")`
built a 4 GiB String natively and aborted on wasm.

## io.read_n_bytes — two defects meeting in the middle

- native reserved `vec![0u8; n as usize]` **up front**, so a negative `n` asked for
  ~1.8e19 bytes;
- the wasm host floor takes an **i32**, so a large `n` arrived truncated —
  `i64::MAX` became `-1` and the read returned **nothing at all** where native read
  every available byte.

The answer here is a **clamp, not a ceiling**, because `n` is a limit whose result
may already be shorter (the latitude POSIX `read()` takes). Both legs now read up to
`min(n, 2^26)` bytes.

**64 MiB is measured, not chosen.** The wasm buffer allocates at 2^26 and fails at
2^27:

| n | native | wasm (before) |
|---|---|---|
| 2^24 | 5 | 5 |
| 2^26 | 5 | 5 |
| 2^27 | 5 | **0** |

## Measured

- 8 probe cells for `read_n_bytes` with live stdin, 7 for `range`, 7 for `pad`:
  **identical on both legs**, including every legitimate in-range case.
- `almide test` **403/403**.
- Fixture is discriminating: on v0.57.0 native dies with a raw `capacity overflow`
  panic while wasm runs on.
- Contract **C-161** extended; fixture linked bidirectionally.

Remaining from the 218: **14 arithmetic/format cells** — `datetime.to_iso` (wasm
truncates the year field to garbage), `datetime.from_parts`, `math.choose`,
`string.from_codepoint`. Those are value defects rather than allocation ones and
follow separately.


O6lvl4:
**Second commit — the 14 arithmetic/format cells, so this PR now closes all 33.**

These produced *no abort at all*. Two different answers, exit 0 on both legs.

### Three of them are one mistake

An **unsigned step in the middle of a signed computation** on the native leg, where
the self-host had always been i64:

| | native (before) | wasm |
|---|---|---|
| `from_parts(2020, -1, …)` | `m as u64` → 1.8e19, `153*(m-3)` wraps → **1540840320** | 1572566400 |
| `choose(i64::MAX, 2)` | u64 accumulator → **4611686018427387905** | -4611686018427387903 |
| `from_codepoint(2^32)` | `cp as u32` truncates to 0 → **one-byte NUL string** | `""` |

`from_codepoint` is the sharpest: `unwrap_or_default()` was already the right intent,
and the cast in front of it was silently converting **invalid input into valid input**.
Neither `choose` value is a binomial coefficient — the requirement is only that they be
the *same* non-answer, so the accumulation is now i64 and explicitly wrapping, like
Almide's `Int`.

### The fourth is the reverse shape

`datetime.to_iso`'s self-host wrote a fixed 20-byte buffer with a 4-digit year, and its
own header declared `SCOPE: {:04} matches v0 for years 0..9999`. Outside that scope it
did not fail — it wrote garbage:

```
to_iso(i64::MAX)   native 292277026596-12-04T15:30:07Z   wasm 6596-12-04T15:30:07Z
to_iso(i64::MIN)   native -292277022657-01-28T08:29:52Z  wasm  .*+)-01-28T08:29:52Z
```

The second row is the serious one: `48 + digit` with a **negative** digit is not an ASCII
character, so the wasm leg produced a **String containing non-UTF-8 bytes**, silently,
exit 0. The year field is now variable width exactly as Rust's `{:04}` is — four columns
minimum, zero-padded, the sign occupying one of them.

> A documented scope that produces a malformed value outside itself is not a scope.

### Measured

- 22-cell probe **identical on both legs**, including every in-range case: valid months,
  `choose(5,2)=10`, `0001-01-01` zero-padding, the surrogate hole `0xD800` → `""`, and
  `0x10FFFF` → one char with `0x110000` → `""`.
- `almide test` **403/403**; `cargo test --workspace` **green** (4 codegen snapshots
  updated for the new guards — the emitted Rust is what they pin).
- Two interp abstentions recorded in `interp-abstain-ledger.txt` in this PR, as the gate
  requires: both fixtures reach `prim.alloc_list` / tagged-block `prim.handle`, which is
  #1226 slice 2 territory. Shrinking the 3-way oracle's coverage is a reviewed decision,
  not silent drift — noting it here so the review is real.
- Contract **C-054** extended; 434/434 fixtures carry their header, links symmetric.

Issue #1598, opened 2026-08-26
Structural wasm leg: matrix intrinsics and the io stdin-read surface are unlinked — routed to the incumbent leg by module scan


O6lvl4:
Found by examples_parity_test (llama_block.almd: every matrix op walls) and io_read_all_test (io.read_all walls with call:io.read_all) at the commissioning switchover.

Both shapes sit outside the measured corpus by construction: stdin-reading programs cannot live in the spec suites (the harness pipes no per-file stdin), and the matrix surface is exercised only through its abort path (matrix_index_oob_aborts). The commissioning PR therefore routes any program making an explicit `matrix.*` or `io.*` MODULE call to the incumbent leg (bare println/print are Named calls and stay structural); the gates keep judging the structural leg directly through emit_program, so verification coverage is unchanged.

The route flips per-module as the structural leg links each surface: matrix via wasm_intrinsic lowerings or primitive decomposition, io stdin via the host's op-31 drain (the embedded host already implements it — only the registry linkage is missing).

Issue #1596, opened 2026-08-26
Structural wasm leg misaligns self-package top-let storage: import self projects wall (ty-mismatch) or panic (VarId out of bounds)


O6lvl4:
Found by tests/crossmod_matrix_test.rs at the commissioning switchover: routing `import self as m` project cells through the structural leg (almide::wasm_leg::lower_to_ir + almide_wasm::emit_program) fails 37 cells — some wall with `ty-mismatch:List(ETy(4))-vs-Scalar(Str)` (a top-let read resolving to the WRONG slot: m.SYSTEM answered with WORDS' storage), some panic with `index out of bounds` at almide-ir VarTable::get (a VarId one past the table).

Minimal repro (walls):
- almide.toml `name = "cellpkg"` + src/mod.almd with `let SYSTEM = "hi"` / `let WORDS = ["a","b"]` + src/main.almd `import self as m` reading `m.SYSTEM`.

The shape sits OUTSIDE the measured corpus: the run manifest sweeps spec/wasm_cross and spec/wasm_fail only, and the gauntlet's layered packages import their modules by name, never via `self`. The commissioning PR therefore routes `import self` projects to the incumbent leg by shape (src/cli/build.rs render_wasm_module_routed — no regression, the matrix stays green on the leg that owns it), and this issue tracks teaching the structural driver the self-package top-let discipline so the route can flip.

Issue #1595, opened 2026-08-26
float.min/max and math.fmin/fmax break IEEE 754-2019 zero ordering on ±0 ties: min(0.0, -0.0) returns 0.0


O6lvl4:
The canonical als ledger's C-306 (ALS-T23, adjudicated 2026-08-21, als #28) requires IEEE 754-2019 minimum/maximum zero ordering, commutative: min(-0.0, 0.0) = min(0.0, -0.0) = -0.0 and max(-0.0, 0.0) = max(0.0, -0.0) = +0.0. It explicitly retires the first-operand tie rule C-049 used to state.

The current intrinsics (almide_rt_float_min/max, almide_rt_math_fmin/fmax) compare with <=/>=, so a ±0 tie returns the FIRST operand: min(0.0, -0.0) = 0.0. All three legs agree with each other (no cross-target divergence) but all three are nonconforming to the canonical rule.

Evidence: als spec/stdlib/float_minmax_zero_order_test.almd is red against this tree by design (it was written red against 0.58.0 and is still red). The commissioning PR imported the other five als stdlib determinism tests but held this one back — it enters the tree together with the fix, the same discipline as the ALS-B3..B11 sections that wait for E070.

The fix is a coordinated behavior change: native runtime intrinsics + the wasm lowering + the interp bridge, C-049's statement rewritten to the IEEE order in the same PR, and the affected wasm_cross manifests re-ratified (float_signed_zero_minmax output changes).

O6lvl4:
Also enters with this fix: spec/wasm_cross/float_sign_minmax_ieee.almd — the canonical als copy already carries the IEEE zero-order assertions (@contract: C-049, C-306 + @ref-allow ALS-T23) while this tree's local copy still pins the retired first-operand rule. Swap the local copy for the als version in the same PR, and re-ratify its run/ast manifest rows.

Issue #1552, opened 2026-08-22
wasm: module-level `let` whose initializer references another module-level `let` walls; across a module boundary the diagnostic degrades to `use of unbound var VarId(0)`


atzurody:
**almide 0.57.2**, macOS (darwin 23.5, arm64), wasmtime.

A module-level `let` whose initializer references another module-level `let` (`let B = "x " + A + " y"`) builds and runs on native, but walls the verified wasm renderer. Two things worth reporting:

1. It is a wall at all — a computed String constant at top level is an ordinary shape (we hit it writing a disclaimer line that embeds an edition string).
2. **The cross-module form gives an unhelpful message** (`use of unbound var VarId(0)`) where the same-module form gives the honest one (`reference to a heap module-level global … no CONST initializer`). The cross-module message sent us looking for a scoping bug in our own code first.

### Repro (`almide init`, two files)

`src/catalog.almd`
```almide
let EDITION = "Guideline v1.2 (2026-03-31)"

// Native fine; wasm walls.
let DISCLAIMER = "Basis column cites " + EDITION + ". It does not certify conformance."

// Same text as a fn: both targets accept this (our workaround).
fn disclaimer() -> String = "Basis column cites " + EDITION + ". It does not certify conformance."
```

`src/main.almd`
```almide
import self.catalog
effect fn main() -> Unit = { println(catalog.DISCLAIMER) }
```

```
$ almide build && ./toplevel_concat
Basis column cites Guideline v1.2 (2026-03-31). It does not certify conformance.

$ almide build --target wasm -o t.wasm
error: this program shape is not yet supported by the verified wasm renderer
  main is outside the MIR-lowering subset: use of unbound var VarId(0)
```

### Controls (same project, only `main.almd` changed)

| shape | wasm |
|---|---|
| `println(catalog.EDITION)` — cross-module plain literal `let` | OK |
| `println(catalog.disclaimer())` — cross-module fn | OK |
| `let DISCLAIMER = "a" + "b" + "c"` in main, literal-only concatenation | OK |
| `let EDITION = …; let DISCLAIMER = "…" + EDITION + "…"` both in main | wall: `reference to a heap module-level global VarId(1) cannot be faithfully materialized in this brick (no CONST initializer — a computed init would inject an uncounted call)` |
| same, but `DISCLAIMER` lives in `catalog.almd` | wall: `use of unbound var VarId(0)` |

[...]

PR #1599, opened 2026-08-26
Stage 2 commissioning: the greenfield wasm engine becomes the product leg

O6lvl4:
# Stage 2 commissioning: the greenfield wasm engine becomes the product leg

## What this PR is

The confluence arc's Stage 2 (就航切替): the greenfield-built structural wasm
engine — measured **610/610 byte-identical to native** on this tree's full
wasm_cross corpus — comes aboard develop as five workspace crates, and
`--target wasm` routes to it by project shape. The incumbent WAT trust-spine
stays compiled and gated; nothing silently falls back anywhere.

## The commits (each buildable, each gated)

1. **Canonicalize split + cloneable checker** — additive fns for per-import-set
   caching; the verbatim path untouched.
2. **Flag-preservation discipline** — every empty `mutated_params` names why the
   fn is fresh (the c3 lesson mechanized); 28 sites audited (8 develop-only,
   incl. the continuation-lift linear-split invariant), gate in almide-ir.
3. **almide-layout** — heap block layout has ONE defining source.
4. **Interp oracle fixes** (greenfield-adjudicated): bound-range concat,
   Option-carrier normalization at pure fn boundaries, layout single-sourcing.
5. **C-302..C-309 ported from the canonical als ledger** with their 8 fixtures
   (numeric-determinism family + bounded head) and spec chapters; the reserved
   hole narrows to C-310..C-318 per the #1544 doctrine. All 8 fixtures verified
   native==wasm on this tree before import.
6. **The carry**: almide-corpus / almide-wasm / almide-wasm-run / almide-spine
   join the workspace. The s5 driver moves to the ROOT lib (`almide::wasm_leg`)
   so the product leg and the spine gates judge ONE implementation
   (`infer_module_capturing` now has one copy, delegated from compile_driver
   and spine). Every parity manifest re-ratified over this tree's corpus:
   run 599→616 rows, check 1204, ast 1200 — shared-row hashes audited against
   the greenfield goldens (check: 0 drift; 4 rows explained by fixture-content
   evolution, e.g. #1552's const-fold graduation).
7. **The switchover**: routing by PROJECT SHAPE, never by failure —
   structural leg (default): single/multi-module programs with main; `run`
   executes on the embedded host (fs/env/stdin as measured), `build` ships the
   `to_wasi` stock-runtime form (578-fixture stock-wasmtime gate).
   Incumbent: main-less libraries (#881), dependency/self-import packages
   (#1596), host-variant BUILD artifacts, `ALMIDE_FUEL_PROBE` instrumentation,
   and `ALMIDE_WASM_INCUMBENT=1` (the reversible switch, one release).
8. **The gate fleet in CI**: release-shape suites on stock wasmtime, codopsy A
   floor + 800-line file discipline, the 40-mutant verification-of-the-verifier
   — each classified in proofs/gate-verification.toml (file-discipline
   negative-tested today; mutation gate exercised through the greenfield burn,
   40/40 patches apply-clean here, 3 sampled mutants caught locally).

## The entire legacy wall corpus graduates

The wall-corpus gate fired on the switchover for the RIGHT reason, sixteen
times: **every shape the incumbent leg refused now lowers byte-identically on
the structural leg** — chained scalar pipes, list.pop/unique/enumerate over
heap and non-flat elements, heap-valued and variant-keyed maps, map.fold with
heap accumulators, three-level list towers (literal and loop-built), codec
list-default and triple-nested fields, bytes/tuple deep equality, nested
payload guards and tuple patterns. All 16 were byte-verified on both targets
and moved to spec/wasm_cross under their family contracts (C-013/C-015/C-070/
C-099/C-101/C-115/C-298) with graduation provenance headers, per the gate's
own protocol. The corpus refills with three live refusals (zlib/net intrinsic
calls, the native-only matrix block) so the gate keeps its purpose.

The run-parity manifest grows 599 → 632 rows and the interpreter reproduces
every one.

## Verification on this tree

- Carried gates, release shape: gauntlet 35 cells (floor 25), wasi_gate
  (stock wasmtime), backend_parity, fuzz_differential, alias_semantics,
  rc_budget, alloc/size ledgers (17 additions + 1 explained change each),
  section dumps, surface matrix, tail_calls, run/check/ast parity — **all green**.
- develop's own: spec 420 test files green, examples 11/11, crossmod matrix
  green, probe battery (charge_probe/heap_cap/alloc_count/rc_placement) green,
  rustc warning ratchet 0/216 units, contracts 311 with 618/618 fixtures
  linked, gate-verification 60 gates, fmt gate clean, release seals verified.
- codopsy: almide-wasm A(90), almide-spine A(94), almide-wasm-run A(98) (was
  B(82) — to_wasi and the fs dispatch split for the complexity budget),
  corpus/layout A(100).

## Develop's own suite hardened the leg further (the final battery)

Running the FULL workspace suite against the switchover surfaced and fixed,
in one hardening commit:

- **index-OOB aborts now speak the native error frame** (`Error: index out
  of bounds` on stderr, exit 1) — the structural read path bare-trapped,
  leaving stderr empty (caught by the xtarget gate's stderr comparison).
- **bytes writers join the shared-cell classification** — a captured Bytes
  var mutated through `bytes.push`/`set_at` was never cell-classified and
  printed silently wrong values (the C-319 failure mode on a class the
  corpus never covered); non-Assign writes against a cell refuse honestly.
- **entry pub fns export as DCE roots** (#457 contract): host-called
  callbacks survive DCE and appear as wasm exports when their whole call
  closure lowers.
- **a structural wall reroutes to the incumbent renderer** — both legs are
  verified, so this is NOT the retired v0 fallback; a shape neither leg
  lowers still fails hard with the incumbent's rich wall rendering.
  `ALMIDE_VERIFIED_DEBUG=1` names each reroute.
- matrix/io module calls pre-route (#1598), the ported mutation gate pins
  its collation locale, and the fmt corpus skips the gauntlet's deliberate
  reject cells.

After the hardening: wasm_runtime_test 79/79, export DCE 2/2, the xtarget
630-fixture sweep green, all carried gates re-ratified.

## Known holds (issues filed)

- #1595 float.min/max ±0 tie ordering (C-306 conformance) — the als fixture
  enters with the fix.
- #1596 structural leg's self-package top-let discipline — the route flips
  when fixed.
- url module + the L-docs examples harness — next PR (its own registry/doc
  gate surface).


O6lvl4:
**CI rounds 4–5 hardening** (each red root-caused, never band-aided):

- **check-frees-churn caught two REAL leaks** in the structural leg:
  1. Growing Str accumulators (`acc = acc + s`) leaked every outgrown block — F_CONCAT borrows and the Assign dec-skip (rhs mentions the var) meant nobody released the old accumulator. Fixed with `$str_append`: in place under rc == 1 with class slack, else concat + release (the incumbent's `__str_append1` window ported; growth stays geometric through the 16<<class slabs).
  2. Droppable params leaked across `return_call` — a tail call replaces the frame, so the epilogue's param release never ran (`grow_tco`'s Str accumulator OOM'd). Params now release at tail sites, **scoped soundly**: entry fns only, never lambdas, and only when the body derives no raw addresses — an unscoped version freed a block `string.is_whitespace` was still reading through `prim.handle` (caught by the burn-up's byte-compare on `string_whitespace.almd`, T flipped to F). The prim-escape hazard is now a named doctrine in the code.
- **walled-real baseline** gains the eight functions the incumbent's MIR still walls (the graduated corpus grew past its admitted profile by design; rows prune with the incumbent), reachability baseline prunes the two rows the graduations made buildable (`option.to_result`, `result.filter`).
- **determinism gates** (host + browser ABI) learn the graduated wall corpus: 16 tracked skips, ceiling locally re-measured; the browser gate gains the wall-vs-error discipline the host gate already had — a browser ABI *rendering* what native walls now FAILS instead of being skipped.
- **fmt gate**: a manifest-sourced wrapper excludes the gauntlet's deliberately-invalid reject cells (the bare sweep exited nonzero on parse-skips at zero drift); negative-tested and rowed in the gate-verification ledger.
- **toolchain pins → 1.94.0** across 7 workflows (the binary embeds the wasmtime 47 host; runner-class MSRV declared on the root package).
- Open question for this round: perf-ratio's idiom ceiling failed on the 1.94 runner but passes locally on 1.96 — if round 5 reproduces it, the perf job moves to stable (its baselines describe what users run) while build jobs hold the MSRV pin.

Issue #1696, opened 2026-08-30
Structural leg all the way: fix the two regressions, land the imperative corpus, say what the default guarantees, put the certificate on the structural leg, retire the incumbent emitter


O6lvl4:
> **状態** — 実装・判断待ち
> **一行で** — structural レッグが既定になった一方で、証明書（step 4）と incumbent 退役（step 5）が残っており、既定で使われるレッグの保証は今も差分試験だけ。
>
> **やること**
> 1. step 4 の続き: `proofs/gate.sh` に着地済みの phase A2 / B1 / step 4（statement + module calls）の行から、`spec/wasm_cross` の corpus 全体へ witness の受理を広げる。
> 2. C-SAFE / C-PROVEN receipt を structural ビルドから再生成できるようにする（コメントの調査が併せて要求している artifact hash の receipt への刻み込みも同じ箇所）。
> 3. step 5: `render_wasm_module_routed`（`src/cli/build.rs:1029`）を一本にし、`ALMIDE_WASM_INCUMBENT`（`crates/almide-base/src/env.rs`、README）を外し、#1690 / #1691 を moot として閉じ、`crates/almide-mir` の emitter コードを削除する（checker/witness 経路は残す）。
> 4. step 4 のルート ○×（structural front から witness を出す / 証明のためだけに MIR へ再降下）を確定させる。コメントでは前者が採られているが、最後のコメントは「ruling は未了」と書いている。
>
> **完了条件** — 本文 Acceptance の残り 2 行: structural レッグが `wasm_cross` corpus 全体で kernel-proven checker の受理する witness を出し receipt が structural ビルドから再生成される、かつ `render_wasm_module_routed` が一レッグ・`ALMIDE_WASM_INCUMBENT` が消滅・#1690 / #1691 が閉じている。
> **根拠** — 本文 Acceptance の 3/5 が済（step 1–3、2026-09-06 の追記）。2026-09-21 のツリー確認: `proofs/gate.sh:254/287/305` に #1696 の phase A2 / B1 / step 4 行、`crates/almide-wasm/src/witness.rs:65` に `WitnessRecorder`、`src/cli/build.rs:1029` に二レッグの router、`ALMIDE_WASM_INCUMBENT` は env.rs と README に現存。最後のコメントが挙げる PR #1999 は unmerged で閉じている（B1 の行自体は別経路で着地済み）。
[...]

Issue #1689, opened 2026-08-30
Structural wasm leg: `bytes.push` growth dies with `out of memory` at ~69,000 pushes while pre-sized `Bytes`, `bytes.from_list` and `list.push` of the same size are fine


O6lvl4:
## Symptom

On the **structural wasm leg**, a `Bytes` grown by `bytes.push` dies with `Error: out of memory` at ~69,000 pushes. Native and the incumbent v1 leg complete. Pre-sized `Bytes` and `List[Int]` growth of the same or larger size are fine on the structural leg, so this is specific to `bytes.push`'s growth path, not to memory.

## Minimal repro

```almide
fn main() -> Unit = {
  var b = bytes.new(0)
  var i = 0
  while i < 69000 { bytes.push(b, int.band(i, 255)); i = i + 1 }
  println("${bytes.len(b)}")
}
```

```
$ almide run repro.almd                                         → 69000
$ almide build repro.almd --target wasm -o s.wasm && wasmtime s.wasm   → Error: out of memory   (exit 1)
$ ALMIDE_WASM_INCUMBENT=1 almide build repro.almd --target wasm -o i.wasm && wasmtime i.wasm → 69000
```

## Threshold and controls (structural leg)

| pushes into `bytes.new(0)` | result |
|---:|---|
| 65,535 / 65,536 / 65,537 / 66,000 / 67,000 / 68,000 | ok |
| 69,000 / 70,000 / 100,000 / 200,000 | `out of memory` |

| control | result |
|---|---|
| `bytes.new(200000)` + `bytes.set_at` ×200,000 | ok |
| `bytes.from_list` of a 70,000-element list | ok |
| `list.push` ×200,000 into `[]` | ok |

The module declares `(memory 6)` with no max and contains a `memory.grow`, so linear memory can grow; the failure is in how `bytes.push` grows (or how the allocator serves it), not in the memory limit. The threshold sitting just above 64 KiB suggests the growth step or the reallocation is wrong past one page — not verified.

## Where it bit

A gzip inflate writing a 68 KB decompressed output. Every smaller input (up to 65 KB) is byte-identical to native on the structural leg; the 68 KB one exits 1 with no bytes.

## Environment

almide 0.60.0 (develop f0fe9cc3b), wasmtime 47.0.2 (90fed3c6a 2026-07-21), arm64 Darwin.

## Class

[...]

Issue #1688, opened 2026-08-30
Structural wasm leg emits wrong bytes when a `mut Bytes` param is forwarded to a callee inside a branch arm — native `A&B`, wasm `\0\0\0`, no diagnostic


O6lvl4:
## Symptom

On the **structural wasm leg** (the default since #1599), forwarding a `mut Bytes` parameter to a callee from inside a branch arm produces **wrong bytes silently**. Native is correct; the incumbent v1 leg refuses the same shape with an honest wall. No diagnostic, exit code 0.

## Minimal repro (9 lines)

```almide
import io
fn one(mut out: Bytes, v: Int) -> Unit = bytes.push(out, v)
fn esc(mut out: Bytes, b: Int) -> Unit = if b == 38 then one(out, 38) else bytes.push(out, b)
fn main() -> Unit = {
  var out = bytes.new(0)
  esc(out, 65)
  esc(out, 38)
  esc(out, 66)
  io.write(out)
}
```

```
$ almide run repro.almd | od -c
0000000    A   &   B
$ almide build repro.almd --target wasm -o repro.wasm     # Built … (structural leg, verified)
$ wasmtime repro.wasm | od -c
0000000   \0  \0  \0
$ ALMIDE_WASM_INCUMBENT=1 almide build repro.almd --target wasm -o repro_i.wasm
wall: exported `pub fn esc` is outside the MIR-lowering subset …: if over an unresolvable condition (Eq over Int; Var vs LitInt) with a call-bearing arm cannot take the both-arms linearization (it would run the untaken arm's effects) not in this brick
```

The incumbent's refusal is the correct answer; the structural leg's bytes are not.

## What is and is not affected (bisected)

| shape | structural |
|---|---|
| `if … then one(out, 38) else bytes.push(out, b)` (callee takes `mut out`) | **wrong** |
| same with `match b { 38 => one(out, 38), _ => bytes.push(out, b) }` | **wrong** |
| same with a block body `{ if … then { one(out, …) } else { … } }` | **wrong** |
| `match b { 38 => bytes.push(out, 38), _ => bytes.push(out, b) }` (stdlib push only, no user callee) | correct |
| the same `match` inlined in `main` instead of in a fn with `mut out` | correct |
| `esc` with no branch: `fn esc(mut out, b) = one(out, 38)` | correct |
| callee pushing many bytes (`for b in string.to_bytes(s) { bytes.push(out, b) }`) in a non-branching caller | correct |

[...]

PR #1698, opened 2026-08-30
Release 0.60.0 — the structural wasm leg ships as the default, with the imperative corpus behind it

O6lvl4:
Release 0.60.0. `develop` → `main`, 310 commits since v0.59.2 (the 0.59.2 tag itself sits on `bounded-profile`; `main` tops at 0.59.1).

## Headline

**The structural wasm leg ships as the default engine** (#1599 commissioning switchover), and the first release in which real imperative programs — a 322-line gzip inflate, a SQL tokenizer — lower on it byte-identical with native. Two regressions the flip introduced (#1688 silent wrong bytes on mut-param forwarding from a branch arm, #1689 `bytes.push` growth OOM) were found and fixed inside the rc window; both now have cross-target fixtures (C-324, C-325) and the inflate itself is contract C-326. The docs say what the default guarantees: the structural leg is *trusted* (differential evidence); the per-build certificate is on the incumbent leg (`ALMIDE_WASM_INCUMBENT=1`). Plan to one certified leg: #1696.

## Also in this release

- **Component Model**: `almide build --target wasm --component` emits a WASI 0.2 component directly; `ALMIDE_COMPONENT_P3` routes onto the WASI 0.3 async canonical ABI with stream stdio; `fan.map` prefetches over async-lowered subtasks and `fan.any` abandons losers (C-321); the p3 component gains the filesystem read/write surface.
- **Language**: or-patterns in match arms; bare `some` / `ok` / `err` are function values (ALS-E27); E080 conditional fix-it diagnostics; E081 turns the both-legs wall set into a check-time diagnostic (target-availability matrix, shrink-only and gated both ways).
- **Native performance (codec arc)**: `Value` joins borrow inference — derived decoders borrow their input, list/option drivers have by-reference twins, `value_field_ref` is a runtime fn; for-in binders iterate by reference when the body only borrows; `json.parse` walks byte offsets and pre-sizes small objects; object keys are interned; counted `while` loops unroll 8×; growing `Str` accumulators append in place. A Codec parsing-conformance corpus ported from an upstream schema library's test suite.
- **Trust spine**: StructuralRuntime / StructuralAlloc / StructuralDecode / StructuralRun in Rocq ground the structural leg's emitted runtime bytes against the proven trees; MC/DC pending debt reaches zero; the formal-credit claims table; the qualification dossier assembles on every release tag, Sigstore-attested; `--trace-map` line-level traceability for `--target rust`; the Ferrocene weekly lane; `check --profile critical` and the flight PID reference kernel (G-F4).
- **Stdlib on the structural leg**: `io.read_all`, the stdin cursor (`read_byte` / `read_line` / `read_n_bytes`), `error.{message,context,chain}`, `json.to_map`, `list.find_index`, the map predicate family, the int bit trio; the `url` module (6 new signatures — interface diff v0.59.2 → develop is **additive**, 0 removed).
- **Tooling**: `almide bench` (verify-then-time, median headline); the REPL answers through the embedded wasm host; `wasm-opt` gets the bulk-memory flag and a refusal is reported as such; the Built line names the leg that produced the bytes; `SECURITY.md` + weekly RustSec audit; `SUPPORT.md`.
- **Sizes (structural leg, verified, no post-hoc optimizer)**: hello 4,590 B (incumbent 1,096 B); the gzip inflate 17,222 B.

## Gates

- `scripts/count-release-blockers.sh` = 0 (#1688 / #1689 closed with fixtures)
- `scripts/check-interface-diff.sh v0.59.2 develop` = additive (added=6 removed=0)
- rc soak: v0.60.0-rc1 … rc3 on 2026-08-30; the regressions above were found against rc2 and fixed before the final.


O6lvl4:
Driving the final from this session: once develop CI and this PR are both green I merge this PR with a merge commit, tag `v0.60.0` on it, and let release.yml create the release (then notes/title edit + `release-seal.sh gen`). If another session is already doing the same, say so here before tagging — one tag only.

Issue #1423, opened 2026-08-14
Target-availability matrix: declare the single-leg surface, gate it both directions, surface it at check time


O6lvl4:
> **状態** — 計測器・ゲート・着手可
> **一行で** — Stage 1–3（probe / 宣言台帳 / check 時の E081）は稼働し Stage 5 も第一段（警告ヒストグラム）が入っているが、Stage 4 の burn-down が残っており、`pending_self_host_ceiling = 34` が 0 になるまで Stage 5 の finding 昇格が開かない。
>
> **やること**
> 1. `proofs/target-availability.toml` の `pending-self-host` 行を潰す。本文の Stage 4 の規則どおり、各行は「同じ PR で self-host body + cross fixture」か「justified-forever への再分類」で閉じる。
> 2. 閉じた分だけ `pending_self_host_ceiling`（shrink-only）を下げる。
> 3. ceiling が 0 になったら `tools/xtarget-fuzz/src/availability.rs` の警告ヒストグラムを finding クラスへ昇格する（2026-09-04 の裁定が 0 を条件にしている）。
> 4. コメントで裁定済みの 2 行を反映する — `env.set` は overlay で wasm 提供可、`testing.assert_throws` は native-only-forever。reason 語彙を `pending-self-host` / `pending-port` / `native-only-forever` のクラスに構造化する follow-up がその受け皿。
>
> **完了条件** — `proofs/target-availability.toml` の `pending_self_host_ceiling` が 0 になり、`bash scripts/check-target-availability.sh` が measured == declared の両方向で通り、`availability.rs` の該当行が warning ではなく finding として出る。
> **根拠** — 2026-09-21 のツリー: `proofs/target-availability.toml:61` が `pending_self_host_ceiling = 34`（2026-09-03 のコメント時点は 74）。E081 は `src/cli/check.rs` / `src/cli/build.rs` から出ており Stage 3 は live。Stage 5 の第一段は着地済み（`tools/xtarget-fuzz/src/availability.rs:1-7`: 2026-09-04 の裁定で警告ヒストグラム、finding 昇格は pending-self-host が 0 になってから）— 2026-09-03 のコメントの「Stage 5 は未着手」はこの点で古い。
> **触る場所** — `proofs/target-availability.toml`、`scripts/check-target-availability.sh`、`tools/xtarget-fuzz/src/availability.rs`、`stdlib/` + `spec/wasm_cross/`（burn-down 1 行ごとの body と fixture）

---

[...]

PR #1702, opened 2026-08-30
Bucket A burned: fold_lines_range and env.sleep_ms leave the both-legs wall (23 → 21)

O6lvl4:
Both #1423 bucket-A cells, landed after the v0.60.0 final.

## fs.fold_lines_range — the twin matrix completes

The fn had exactly one typed self-host twin (`_ls`); the probe's `A = Int` instantiation routed to an unregistered `_x` name and walled. `fold_lines_range` gains `_i`, `_s`, `_msi` (transcribed from the proven `_ls` walker over the same range-ownership rule) and `fold_lines_chunked` its missing `_s`, so **every accumulating streaming walker carries all four accumulator twins** — the `mod_p4_h.rs` routing drops its per-walker exceptions, and a new gate in `tests/fs_streaming_family_gate_test.rs` makes a missing twin cell a build failure. The embedded-size baseline rises to 904,266 on purpose (the four twins). Spec: a twin-matrix block in the range test exercising all four accumulator types.

## env.sleep_ms — every leg pauses

- **Embedded host**: fs_call **op 36** (the ms count rides the a_len slot with a null a_ptr — the op-35 scalar discipline, handled before the buffer reads); the host sleeps the thread.
- **Stock-WASI p1 artifact**: a **monotonic busy-wait** over the `clock_time_get` the shim already imports. WASI p1 has no sleep short of `poll_oneoff`, and importing a sixth WASI function would shift every defined function index while the element section is copied verbatim — a funcref-table corruption of exactly the #1688 silent class. The spin is confined to stock p1 artifacts and retires with the component leg's `wasi:clocks#wait-for` (the async-ABI form the vendored 0.3 WIT already carries — the reference Wasm-GC compiler in ../almide-references binds exactly that, which is the quality bar for the component arc).
- **Build path**: a **reverse verified-to-verified handover** — a shape-routed host-variant program the incumbent walls now gets one structural attempt before the build fails (forced routes and the main-less library form stay final). Verified end-to-end: `almide build` + stock `wasmtime run` sleeps 300 ms and prints.

## Matrix, by measurement

The four-direction gate: `340 lower, 49 native-only (pending-self-host 15/15), 21 wasm-unavailable` — two rows deleted by the gate's own demand, the range row re-declared honestly as `structural-leg-gap`, the pending-self-host ratchet 16 → 15. E081 no longer fires for either fn (verified on real programs). Contract C-327 + fixture `env_sleep_pause.almd`; ledgers ratified (six + surface golden + interp abstain + embedded-size).

Remaining #1423 walls: 9 wasi:http port work, 12 justified-forever candidates pending rulings.


O6lvl4:
CI event delivery stalled for the last two pushes (9576c77, nudge 943e582) — re-firing via close/reopen.

PR #1711, opened 2026-08-30
Structural witness phase A: the emitter records its ownership events, the floor holds them balanced

O6lvl4:
#1696 step 4, phase A of the adopted design (witness from the structural emitter — the mob-point comment on the issue). Stacked on #1709 (env.set); the diff shrinks to the witness work when that merges.

## What lands

- **`crates/almide-wasm/src/witness.rs`** — the `WitnessRecorder`: per-object refcount-event streams in certificate v0 vocabulary (`i` fresh/param ownership, `a` share, `d` release, `m` move-out), recorded **at emission time** by hooks at the exact instructions: the Bind route's `rc_inc_top`/fresh-alloc settlement, the ret-inc + move-out pair of a heap Var return, and the fall-through epilogue's `$dec_flat`s. The witness describes the emitted instruction stream — one contract level closer to the bytes than the incumbent's MIR-side projection.
- **The straightline gate** — recording engages only for a body the pre-scan proves covered by those hooks: literal/Var binds, a scalar/Unit tail or a bound-Var heap return; no effect wrap, no captures, no top-lets. Everything else declines. A gate/hook disagreement poisons the certificate into a loud `!poison` sentinel — never a silent under-count.
- **`tests/witness_floor.rs`** — sweeps the run-manifest corpus with the sink armed and holds: zero poison, every certificate balanced under the Rust mirror of the proven rule (prefix-nonnegative, zero final), the two emission passes byte-agree per fn, and the witnessed count ≥ `golden/witness-floor.txt` (grow-only; **floor 3**, including one organic corpus fn — `id_list` in `r5_lowmisc_param_try_err` — witnessing `iaamdd`).
- Fixture `spec/wasm_cross/witness_straightline.almd` (C-033 evidence): the alias/heap-return shapes, byte-identical on both legs.

## What phase A honestly claims

The recorder + balance mirror + floor. The kernel-proven checker does NOT yet re-verify these certificates — that is phase A2 (gate.sh hookup), then branches/loops (B), names+caps and the receipt-row flip (C), retirement (D), per the issue plan.

Ledgers: run/check/ast manifests, alloc/size rows, contract evidence, file discipline all green; `almide test` 426/426; backend-parity/alias/wasi gates green.


O6lvl4:
Landing via #1718 instead: the stack's tip PR contains this branch verbatim, and with the merge queue now active one queue validation covers the whole stack — three serial validations collapse to one. Content is unchanged; reopen only if #1718 fails.

PR #1712, opened 2026-08-30
Unused fixed-slot helpers ship as stubs — hello 4,670 → 2,225 B, the whole corpus −28.5 %, zero output changes

### LINKED ISSUE #1699: Structural leg ships every runtime helper: hello is 4,590 B of which wasm-opt proves 92% dead — prune helper emission by reachability inside the verified path
## Symptom

The structural leg's default build carries a fixed floor of ~3.5 KB that the program does not use. Hello, world at v0.60.0:

| | bytes |
|---|---:|
| structural leg, shipped (verified) | **4,590** |
| incumbent v1 leg, same program | 1,096 |
| structural + `wasm-opt -Oz` (not verified) | **364** |

Section anatomy of the shipped structural module: `code` 3,824 B across **40 functions**, `types` 19, `globals` 14, `imports` 5, `data` 261 B. `wasm-opt -Oz` removes 92 % of it — and wasm-opt only deletes what is provably unreachable, so ~35 of those 40 functions are runtime helpers the program never calls (rc inc/dec/free variants, the bytes-push growth helper added in 6af76c99b (+131 B on every module), string builders, list drops, …). The same floor shows on every program: pi_approx 10,813 vs 8,767 incumbent; the noprint pi variant 5,051 vs 1,242.

## Why it matters

The size story is a real differentiator on the wasm target and the README's size table is the first thing a reader compares. Since 0.60 the *default* build is the structural leg, so the number a user gets is 4,590, four times the incumbent's and larger than every small-Wasm language's hello we would want to sit next to. The verified module must stay verified — the answer is not to run wasm-opt, it is to not emit what is dead.

## Ask

Reachability-pruned helper emission on the structural leg: emit a runtime helper only when the program (transitively) references it, keeping the emitted bytes inside the verified path (StructuralDecode/StructuralRuntime already ground the helper trees byte-for-byte, so pruning is a selection, not a rewrite). Target: hello within ~1.2× the incumbent's 1,096 B; the `docs/benchmarks/wasm-size.txt` ratchet then records the floor. The helper set is small (40 fns) so a per-helper `used` bit from the emitter's own call graph should be enough; the `$alloc`/rc core stays unconditional.

Measured 2026-08-30 with the v0.60.0 release binary, wasm-tools 1.257, wasm-opt 132.

O6lvl4:
Closes #1699.

The 34 fixed-slot runtime helpers shipped in every structural module regardless of use; in a small program most were dead weight (hello: 3,824 B of code, ~92 % provably unreachable by `wasm-opt`). Helper indices are position constants, so this prunes by **stub, never renumber** — the same discipline unreached registry bodies already use three lines below in `assemble_module`.

## Mechanism

- The assembly builds the 34 helpers as an indexed list (a `debug_assert` pins the list to the `F_*` slot order), then computes the used set: roots are every body that ships after the helper block (the reachable lowered fns, `main`, the program-driven extras), edges inside the helper set come from **scanning each helper's own encoded body** — wasmparser over the same bytes `wasm_encoder` wrote, `call` and `return_call` both — folded to a fixpoint. Nothing is hand-listed, so a helper growing a new callee cannot drift the edge set; a body that fails to parse claims every helper (over-keeping is bytes, under-keeping is a trap).
- The proven runtime core is seeded unconditionally: `$alloc`, `$free`, `$inc`, `$dec_flat`, `$cow` — the trees StructuralRuntime.v grounds byte-for-byte. The grounding gate reads those bytes from the emitter's own `dump_runtime_bytes` (unit-level), so it is independent of any module's stubbing either way; keeping them shipped is the conservative choice.
- An unused helper's slot keeps its declared type and index and ships as the 2-byte stack-polymorphic `unreachable` stub.

## Numbers (structural leg, verified, no post-hoc optimizer)

| program | before | after |
|---|---:|---:|
| Hello, world | 4,670 | **2,225** (incumbent: 1,096) |
| pi_approx (1M Leibniz + float print) | 10,813 | **8,465** |
| the C-326 gzip inflate | 17,222 | **15,180** |
| whole `wasm_cross` corpus (642 sized fixtures) | 4,994,841 | **3,571,846** (−28.5 %) |

Every one of the 642 fixtures shrank, none grew, and **no fixture changed its structural/refused classification**. Output byte-identity is untouched: the spine run-manifest oracle is green, `almide test` is 426/426 (417 wasm), and the pruned gunzip still decodes all six reference inputs byte-identical with `gzip -dc`.

## Ledgers regenerated in this PR

- `size-baseline.txt` re-tightened (`ALMIDE_UPDATE_SIZES=1`) — the ratchet would have passed silently on the slack, so the 642-row shrink is committed as the new floor.
- Section-dump goldens (`ALMIDE_UPDATE_DUMPS=1`): the dumped test modules now show stubs where their unused helpers sat. Reviewers should note the dumps pin fewer helper-body encodings than before — the used helpers' encodings are still pinned wherever a fixture reaches them, and the proven four by the grounding gate.
- `docs/benchmarks/wasm-size.txt` + README block restamped (`gen-readme-stats.sh --measure`): structural hello 2,225 B.

Local `check-host-determinism.sh` could not build its wasm32 harness on this machine (no rustup) — identical failure on clean develop, so that gate's verdict comes from CI.


O6lvl4:
Heads-up from the session driving the #1709→#1718 stack: that ordered stack is mid-merge and touches the same size/section ledgers this PR regenerates (size-baseline 642 rows, dumps, README stats). To avoid a mutual-clobber, the stack lands first; if this PR is unattended afterwards I will rebase it and re-run its regen steps (ALMIDE_UPDATE_SIZES / ALMIDE_UPDATE_DUMPS / gen-readme-stats) before merging — the pruning logic itself is untouched.

Issue #1715, opened 2026-08-31
runtime/rs becomes a linkable crate so the embedded host stops carrying a textual copy of the http client


O6lvl4:
Debt filed deliberately (2026-08-31, #1710 increment 1): the embedded host's http client (`crates/almide-wasm-run/src/http_client.rs`) is a **verbatim textual transcription** of `runtime/rs/src/http.rs`'s client core, because runtime/rs is splice-source for generated programs (references `AlmideMap` and prelude types, has no standalone build), not a crate anything can link.

The copy is net-protected — the `spec/embedded_cross` fixtures run BOTH copies on the same probes and fail on drift (C-328) — but a copy is a copy, and every future client fix (timeout wording, framing tolerance) must land twice.

## Repayment

Split runtime/rs into (a) a real `almide-rt-core` crate holding the prelude-independent cores (the http client at minimum; candidates: hash/hex/base64), linkable by almide-wasm-run AND spliced-into by the template layer, and (b) the template glue that stays splice-source. The `#419` regeneration discipline (rust_runtime.rs embeds source text) must keep working — the buildscript embed either includes the core crate's source or the template re-exports it.

Exit: `http_client.rs` deleted; almide-wasm-run links the shared core; the embedded_cross net stays green through the swap.

O6lvl4:
Shipped in PR #1775 (the linear replacement of #1771): the http client core moved to crates/almide-rt-core — parse/read/TLS/chunked/exchange in one source, spliced into the generated runtime by the codegen buildscript and linked normally by almide-wasm-run, so the embedded host no longer carries a textual copy. C-328's shared-code wording guarantee now holds by construction.

Issue #1710, opened 2026-08-30
wasi:http@0.3 client leg: the http family rides the p3 async plumbing and fan keeps its parallelism


O6lvl4:
> **状態** — 実装・着手可
> **一行で** — http クライアント族は framed 族まで p3 コンポーネントで動き、fan×http の並列性も実測で生き残った。残っているのは `request_stream`（0.3 streams）と、`proofs/target-availability.toml` に `p3-component` レッグの行が一つも宣言されていないこと。
>
> **やること**
> 1. `request_stream`（`stdlib/http.almd:190` の `@intrinsic("almide_http_request_stream")`）を 0.3 の `stream<u8>` に乗せる。チャンクのドレインは既存の応答 body 読み出しループと同じ形。
> 2. `proofs/target-availability.toml` に `p3-component` レッグの行を起こす（同ファイル 26-27 行が今も「no rows declare it yet」と書いている）。availability は per-leg リストで shrink-only。
> 3. 本文 Acceptance 3: サービスを得たレッグ側で E081 / walled-real / skip 台帳の行を退役させる。
> 4. 本文 Acceptance 4: クライアント族のクロスターゲット contract fixture を同じ PR で足す（`// @contract:` 行付き）。
>
> **完了条件** — `http.request_stream` の unavailable 行が p3 レッグで消え、`proofs/target-availability.toml` の declared == measured 掃引が `p3-component` を含み、その台帳行が縮み、クライアント族のクロスターゲット fixture が contract 行付きで存在する。
> **根拠** — コメントの実測（2026-09-06）: onomly の probe workload（2 名 × 18 本の実ネットワーク probe を一つの `fan.settle`）が p3 コンポーネントで 8.5 s、native バイナリ 10.5 s / embedded 8.9 s → Acceptance 2 は達成済み。increment の PR #1754 / #1778 / #1917 と #1930 はいずれも MERGED。2026-09-21 のツリー確認: `proofs/target-availability.toml:283` に `http.request_stream` の unavailable 行が structural / stock-p1 / embedded で現存。
> **触る場所** — `crates/almide-wasm-run/src/wasi_p3_http.rs`、`proofs/target-availability.toml`、`stdlib/http.almd`、`tests/component_p3_test.rs`

---
The named follow-up of #1628, and bucket B of the #1423 triage: give the `http.*` client family a wasm leg by connecting it to `wasi:http@0.3` on the p3 async plumbing the fs surface established.

## Why 0.3 and not 0.2

[...]

PR #1717, opened 2026-08-31
The http string family crosses to the embedded wasm lane (#1710 increment 1)

O6lvl4:
First increment of #1710, stacked on #1709 → #1711 (the diff shrinks as they merge).

## What lands

`http.get/post/put/patch/delete` run on the **embedded wasm lane** (`almide run --target wasm`) answering **byte-identically to native** — verified on closed-port refusals, DNS failures, AND a live TLS fetch on both legs.

- **Guest**: fs_call **ops 43..=47** (the method is the op; url in a, body in b — the fs.write two-string convention), decoded by the existing `fs_result_string`. The `http` module joins the host-call dispatch with a **fallthrough that preserves the family's pre-port route verbatim** (`lower_module_call_c`): the pure fns — `url_decode`, the `response` constructors — keep their self-host registry service (the first fallthrough attempt regressed them to walls; the four-direction gate caught it).
- **Embedded host**: `crates/almide-wasm-run/src/http_client.rs` — a verbatim transcription of the native runtime template's rustls client (provenance and the drift net named in the header; the copy is **debt #1715**, repaid when runtime/rs becomes a crate). Error texts match native by shared code, not replication.
- **Stock artifacts refuse at build** through the #1709 op audit (no runtime refusals on a runtime the developer never ran). The stock/component story is the rest of #1710.

## The new net

`spec/embedded_cross/` + `tests/embedded_cross_test.rs`: every fixture runs native and embedded and must match stdout+exit byte-for-byte — the lane the stock sweeps structurally cannot cover. Offline-deterministic (closed ports, invalid names). Contract **C-328**.

## Matrix, by measurement

The four-direction gate demanded each step: `347 lower (+5), 44 native-only (−5), 23 wasm-unavailable, all four directions agree` — the five clients join the structural lower set; their `wasm-unavailable` rows stay (the stock build refusal) with `embedded-lane-served` reasons pointing at the component port.

Battery: `almide test` 426/426, embedded_cross green, contract ledger 328/646 bidirectional, clippy ledger held.


O6lvl4:
Landing via #1718 instead (same reason as #1711): the tip PR carries this branch verbatim, one queue validation covers the stack. The mutant 007/025 refresh and the walled-real row ride along unchanged.

Issue #1729, opened 2026-08-31
Embedded wasm host: five perf-corpus workloads die 'Error: out of memory' that native completes in 20-50 ms


O6lvl4:
Found measuring #1701's runtime ledger (2026-08-31). `almide bench --target wasm` (embedded host) on the perf corpus:

| workload | native median | embedded wasm |
|---|---:|---|
| fft | 38.9 ms | **OOM** |
| strchurn | 51.1 ms | **OOM** |
| listbuild_append | 21.1 ms | **OOM** |
| listbuild_combinator | 21.1 ms | **OOM** |
| listbuild_prealloc | 21.0 ms | **OOM** |

All five print the DEFINED C-197 `Error: out of memory` + exit 1 (so the failure surface is correct); the question is why workloads that fit native memory comfortably exhaust the 32-bit linear memory / allocator budget on the structural leg — geometric bump growth with no free-list reuse for the churn shapes is the suspect (`runtime_alloc.rs`: size-class free lists exist, so measure where the growth actually goes before concluding).

Sibling context: nbody/spectralnorm/binarytrees DO measure (2.7/3.4/2.8 ms — binarytrees is FASTER than native's 5.35 ms), so this is workload-shaped, not a blanket ceiling.

Ledger rows carry `oom-embedded` status and the #1701 gate re-measures them every run — a fix here flips the row to `measured` and tightens the ratchet in the same change.

O6lvl4:
Minimal repro + a leg comparison that narrows this a lot. Found while benchmarking against another wasm-targeting language on 0.61.0 (arm64 Darwin, wasmtime 47.0.3).

**The trigger is index-writes in a loop over a large `List`, not the allocation.** Allocating the list is fine at any size I tried; the OOM appears once enough `xs[i] = ...` writes run against it.

```almide
fn main() -> Unit = {
  var xs = list.repeat(true, 100000)
  var i = 0
  while i < 10000 {
    xs[i] = false
    i = i + 1
  }
  println(int.to_string(list.len(xs)))
}
```

| build | result |
|---|---|
| native | `100000` |
| `--target wasm` (structural leg, default) | `Error: out of memory` |
| `ALMIDE_WASM_INCUMBENT=1 --target wasm` | `100000` |

**The incumbent leg handles it; only the structural leg dies.** That should localise the search a lot — it is not the shared allocator alone.

Write-count threshold at `len = 100000`, holding the list size constant:

| writes | structural leg |
|---:|---|
| 100 | OK |
[...]

PR #1730, opened 2026-08-31
Wasm runtime gets its number: a per-benchmark ledger, honest holes, and a ratio ratchet

### LINKED ISSUE #1701: Wasm runtime numbers are still "deliberately absent": publish an `almide bench --target wasm` ledger with a ratchet now that the structural leg is the default
## Symptom

README, "Performance": *"Wasm runtime numbers are deliberately absent rather than estimated."* The native leg has a CI-gated ratio ratchet against handwritten Rust; the wasm legs have a size ledger and a build-speed table but **no published runtime number** — not for the embedded host, not for stock wasmtime, not per commit.

That was the right call while the wasm leg was moving; since 0.60 the structural leg is the default target and `almide bench --target wasm` exists (verify-then-time, median headline, embedded host). The instrument is there; the number is not.

## Why it matters

Perceus-style reference counting in linear memory versus a host GC is exactly the runtime trade-off a reader wants quantified, and we cannot say which way it goes today. A competitor-shaped comparison table has an empty cell where our strongest engineering argument should be. It also means a wasm runtime regression has no gate.

## Ask

1. A `docs/benchmarks/wasm-runtime.txt` ledger produced by `almide bench --target wasm` over the perf corpus that `research/benchmark/perf/` already holds (n-body, spectral-norm, fasta, FFT, the list-materialization case), with the same N-run median discipline the build-speed table uses, on both the embedded host and stock wasmtime.
2. A README block rendered from it by `gen-readme-stats.sh`, with the native ratio beside it so the wasm/native factor is explicit.
3. A ratchet in the commissioned wasm gates (release shape) at a stated tolerance, like the native 1.5× build-speed ratchet — measured on CI's own hardware and demoted to informational on local runs, as the clippy count already is.
4. Per-commit history is optional; the gate is not.

O6lvl4:
Closes #1701.

**The ledger** (`docs/benchmarks/wasm-runtime.txt`, `--measure`-stamped, never hand-edited): every perf-corpus benchmark, both legs, verify-then-time medians. Twelve rows, four honest statuses:
- **measured (3)** — nbody **2.34×**, spectralnorm **~2.1×**, and binarytrees **0.85× — the embedded wasm host BEATS native** (its thread pool runs the fan arms), which is exactly the Perceus-vs-host-runtime datapoint the README could not state before.
- **routed-incumbent (3)** — fasta/fannkuchredux/mandelbrot route to the incumbent artifact; the embedded bench honestly refuses.
- **walled (1)** — onebrc (`string.split_once` wall).
- **oom-embedded (5)** — fft/strchurn/listbuild×3 exhaust the embedded heap; filed as #1729 with the native-vs-OOM table.

**The ratchet** (`scripts/check-wasm-runtime-ratio.sh`, wired into the Commissioned wasm gates job): re-measures the RATIO wasm/native per measured row — the ratio cancels the machine, the same discipline as the native perf-ratio gate — with a wide two-direction budget (fan benches get 200%: binarytrees legitimately swung 0.52→0.89 between back-to-back local stamps, and a flaking gate teaches people to ignore it). Every NON-measured row is re-measured too: a hole that starts benching FAILS the run with the good news until its row is promoted, and a measured row that stops benching fails as a regression — bench-ability itself is ratcheted.

**The README**: "deliberately absent rather than estimated" is replaced by a rendered block (gen-readme-stats.sh, same committed-source rule as the size block) with the measured ratios and the hole counts.

**Corpus repairs in passing** (verified against the released binary): binarytrees and mandelbrot had rotted to E041 (pre-ADR-0008 `fan.map` binds — one `!` each); fft self-timed via `env.millis()`, which `almide bench`'s output-verify rightly refuses — the self-timing is gone (bench measures), bench.py's line1 note updated.


O6lvl4:
CI paid for a design lesson on the first run: the wasm/native ratio does NOT cancel hardware across engines — nbody measured 2.09 on the stamping M-series and **19.95 on the 2-core ubuntu runner** (the embedded engine's fixed costs and the fan thread-pool advantage both scale with the machine; binarytrees flipped from 0.85 to 3.88 the same way). The premise held for the native/Rust gate because both legs share one engine; it breaks cross-engine.

Reworked in the follow-up commit: **CI now gates the status taxonomy** (a hole that starts benching, a measured row that stops, a changed refusal class all fail — that part was green on CI and is the real regression net), and the **ratio verdict runs on the stamping machine class only** (local runs, or `WASM_RUNTIME_RATIO_VERDICT=1`). Ledger header and the README block say exactly that; both modes verified locally (full verdict green, `GITHUB_ACTIONS=1` simulation green with informational ratios).

PR #1761, opened 2026-09-01
flat_map chunks and index writes stop retaining O(n^2) generations on the structural leg

O6lvl4:
## Summary
Increment 2 of #1729: the `listbuild_combinator` row (`list.range |> list.flat_map`, the CHEATSHEET-recommended idiom) OOM'd on the structural embedded leg at 2^16 elements where its true working set is ~1.5 MiB.

**Root cause** — `lower_list_flat_map` merged every callback chunk with a per-element `$concat`: the whole accumulator was re-copied each iteration and **neither the outgrown accumulator nor the chunk was ever freed**, so n chunks retained O(n²) bytes. The self-hosted two-pass `list_flatmap.almd` never runs for this shape — the inline HOF path is the only path.

**Fix** — 8-byte scalar chunks (Int/Float, the same gate as increment 1's assign-site append window) now push element-by-element into the accumulator via the amortized `$list_push_8` and the chunk is freed per iteration (`$dec_flat` is a full free for flat blocks). A borrowed chunk — a callback returning its captured list — takes inc-before-dec, so fresh and aliased chunks balance identically. Handle-element chunks keep the concat merge (freeing after a raw byte copy needs the Dup discipline, the C-186 trap — same boundary as the append gate).

## Evidence
- `listbuild_combinator` flips **oom-embedded → measured**: 92.33 ms embedded vs 21.33 ms native (ratio 4.33; the append row is 4.08). Checksum byte-identical to native at 2^20.
- Aliased-chunk / empty-chunk / string-chunk probes agree native == structural leg.
- New `flat_map_chunks_recycle_under_fixed_budget` in `tests/static_memory_test.rs`: 100k fresh-chunk pushes + 50k aliased-chunk pushes complete under a fixed 8 MiB `--heap-cap` and the shared list survives — no fixed budget survives the old quadratic retention.
- `almide test` 427/427; `cargo test --workspace` green; alloc ledger ratified (one row, smaller).

Remaining #1729 shapes: `listbuild_prealloc`/`fft` (index writes), `strchurn` (over-512KiB block churn).

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,12 +170,13 @@ build. Regenerate with `almide run tools/almide-gates/src/main.almd -- bench`; t
 <!-- wasm-runtime:generated:start — rendered from docs/benchmarks/wasm-runtime.txt by scripts/gen-readme-stats.sh; DO NOT EDIT between the markers -->
 | Benchmark (`almide bench`, verify-then-time, median of 5) | wasm/native ratio |
 |---|---:|
-| nbody | **2.16×** |
-| spectralnorm | **2.25×** |
-| binarytrees | **0.81×** |
-| listbuild_append | **4.12×** |
+| nbody | **2.12×** |
+| spectralnorm | **1.94×** |
+| binarytrees | **1.15×** |
+| listbuild_append | **4.08×** |
+| listbuild_combinator | **4.33×** |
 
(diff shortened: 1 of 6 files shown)


O6lvl4:
Second commit widens the increment: `lower_index_assign` still did the RC-3-era **unconditional $block_copy per write** (a fresh generation materialized on every `data[i] = v`, never freed). It now takes the RC-5 COW judge like every other in-place mutation entry — a uniquely-held list writes in place, a shared one copies once and releases a source ref. Value-semantics probes (shared snapshot keeps its zeros) agree native == structural leg, and a new `index_writes_stay_in_place_under_fixed_budget` test pins both halves under an 8 MiB cap.

Ledger: **fft oom→measured (205.05 ms, ratio 5.17)** and **listbuild_prealloc oom→measured (82.62 ms, ratio 3.92)**. Of #1729's four remaining shapes only `strchurn` (over-512KiB block churn) is left. `almide test` 427/427, workspace green, alloc ledger ratified.

PR #1769, opened 2026-09-01
zlib self-hosts: one decoder on every leg, a stored-block encoder, and exposed checksums

### LINKED ISSUE #1700: `zlib` is native-intrinsic only and walls on both wasm legs — self-host it from the C-326 inflate so a program that imports zlib builds for wasm
## Symptom

`import zlib` compiles natively but every `zlib.*` call walls on both wasm legs:

```
wall: main is outside the MIR-lowering subset: effectful/impure stdlib Module call zlib.gunzip needs a declared capability not in this brick
```

`stdlib/zlib.almd` is eight `@intrinsic("almide_rt_zlib_*")` declarations with no Almide body, so there is nothing for the wasm legs to link. The README's "APIs not yet implemented on wasm are compile- or run-time refusals" covers this honestly — but zlib is the one module a real-program benchmark reaches for first (a gzip-from-stdin decoder is a standard size/perf row), and today that row can only be written by hand.

## What exists now

`spec/wasm_cross/gzip_inflate_members.almd` (C-326, #1697) is a complete pure-Almide gzip decoder: bit reader through tuple returns, stored + fixed + dynamic Huffman blocks, LZ77 back-copy into a `mut Bytes`, CRC-32, ISIZE, multi-member, FNAME header — 322 lines, byte-identical with `gzip -dc` on both native and the structural leg, 17,222 B as a wasm module. That is the decoder half of a self-hosted `zlib`, already verified.

## Ask

Make `zlib` self-hosted so it lowers on wasm like the rest of the stdlib:

1. `zlib.gunzip` / `zlib.inflate` / `zlib.decompress` from the C-326 decoder (gzip container, raw deflate, zlib header+adler32), registered in `self_host_registry.rs`; the native intrinsic stays as the oracle in the 3-way check.
2. `zlib.gzip` / `zlib.deflate` / `zlib.compress` — a stored-block + fixed-Huffman encoder is ~150 lines and enough for round-trip correctness; level-aware LZ77 can come later behind `compress_level`.
3. A `crc32` / `adler32` pair exposed (the decoder needs them anyway) — today a program has to carry its own CRC table.
4. Cross-target fixtures for every entry point (round-trip + known-answer vectors), the purity registry row, and the stdlib doc index regenerated.

Not asking for: matching a C zlib's compression ratio. The point is that a program which imports `zlib` should build for wasm.

O6lvl4:
## Summary
Closes #1700. `import zlib` now builds and runs on the wasm legs. (Stacks on #1765 — the first three commits are that PR's; this PR's own change is the last commit.)

**The decoder** (`stdlib/zlib_inflate.almd`): the C-326 gzip decoder promoted to stdlib — bit reader, stored/fixed/dynamic Huffman, LZ77 back-copy, crc32/ISIZE, multi-member — serving `zlib.gunzip`, `zlib.inflate` (raw DEFLATE), and `zlib.decompress` (zlib header + adler32 verify). DEFLATE decoding is exact, so success outputs are byte-identical across legs by construction: the C-330 fixture pins three NATIVE-flate2-encoded vectors (dynamic Huffman included) decoding identically on native, structural-wasm, and the reference interpreter.

**The encoder** (`stdlib/zlib_deflate.almd`): stored-block DEFLATE + the zlib/gzip containers for `zlib.deflate[_level]` / `zlib.compress[_level]` / `zlib.gzip`. Native keeps flate2 (real ratios) — compressed bytes are LEG-DEPENDENT BY DESIGN, the bounded divergence C-330 declares; the promise is the round trip, pinned on every leg. The `_level` twins accept and ignore the level on wasm.

**Checksums**: `zlib.crc32` / `zlib.adler32` — pure Almide on every target (no intrinsic), known-answer tested (stdlib 971 → 973).

**Three compiler/oracle bugs fixed underneath** (the first registry module with `mut` params + index writes flushed them out):
- `mut_param.rs`: rewrite_signatures / rewrite_call_sites / hoist_branch_writebacks allocated MODULE fns' writeback locals in the PROGRAM var table — VarIds past the module table's length, an ICE in any spliced mut-param module.
- `almide-interp`: inside the pool tier a binding can hold a slot-block ADDRESS (the registry's prim-based `list.repeat`); index READ and index ASSIGN on that model are now served through the arena (len-header bounds + 8-byte slots) instead of aborting "malformed index-assign".
- `purity.rs`: the zlib surface is enumerated as deterministic data transforms (the http-codec admission class), so the v1 leg lowers it.

**Ledgers**: whitelist sum-tier audit note for the 8 impls; `wasm_skip_ledger`'s zlib row RETIRED (the skip's own comment said "lifting this means implementing DEFLATE in Almide" — done); 8 wasm-reachability rows pruned (join parity at next regen); contract C-330 + `spec/wasm_cross/zlib_selfhost.almd` + `spec/stdlib/zlib_test.almd` (merged with the old native-only tests; the ratio assertion left with the skip marker — stored blocks do not shrink and never promised to); manifests, alloc/size/surface goldens, doc index, README stats (973/43, 330 contracts), SPEC/WASM-OUTPUT counts.

## Evidence
- Cross fixture: 8/8 lines identical on native == structural wasm == interp (3-way oracle 650 fixtures green; spine run-parity green).
- `almide test` 427/427; `cargo test --workspace` green; `make verify-trust` green locally (walled-real 12/12, no new rows); availability + contracts + docs-gen gates green.

Known deliberate remainder: `almide test`'s v1 wasm leg still falls back to native for `zlib_test.almd` (the incumbent walls the decoder's tuple write-back loops — the same C-326 division of labor; prunes with #1696 steps 4-5). The structural leg runs it all.


O6lvl4:
Superseded by #1776: same work, linear history + the C-331 renumber (C-330 went to the p3 http lane, which merged first). The original branch's stack-lineage merge commits made the rebase queue drop it with merge_conflict before CI — the #1771→#1775 class.

Issue #1592, opened 2026-08-26
http.request always fails against servers that close without TLS close_notify (e.g. api.osv.dev) — complete response is discarded


atzurody:
## Summary

`http.request` fails with `read failed: peer closed connection without sending TLS close_notify` against any server whose front-end closes the TLS connection without sending `close_notify` — even though the complete HTTP response body has already been received. The whole response is discarded and the call returns `err`.

`api.osv.dev` (Google front-end) closes this way on every request, so OSV — and presumably many other Google-fronted APIs — are unusable from Almide today. Cloudflare-fronted APIs send `close_notify`, which is why e.g. the Workers AI endpoint works fine.

## Repro (almide 0.59.0, macOS arm64 native)

```toml
[permissions]
allow = ["Net", "IO"]
```

```almide
import http

effect fn main() -> Unit = {
  let q = "{\"queries\":[{\"package\":{\"name\":\"lodash\",\"ecosystem\":\"npm\"},\"version\":\"4.17.15\"}]}"
  var hs = map.new()
  hs = map.set(hs, "Content-Type", "application/json")
  match http.request("POST", "https://api.osv.dev/v1/querybatch", q, hs) {
    ok(resp) => println("OK " + int.to_string(string.len(resp))),
    err(e) => println("ERR " + e),
  }
}
```

Output (100% reproducible, GET fails identically):

```
ERR read failed: peer closed connection without sending TLS close_notify: https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof
```

`curl` against the same URL/body succeeds and returns the full JSON.

## Analysis

The client always sends `HTTP/1.1 ... Connection: close` (visible in the binary's strings) and appears to read the body until EOF. When the peer closes without `close_notify`, rustls reports `UnexpectedEof` and the error is propagated even if the response carried `Content-Length` and the body is already complete.

Suggested fixes (either is enough):
1. When the response has `Content-Length` (or chunked encoding is complete), stop reading at the end of the body instead of reading to EOF — the missing `close_notify` then never surfaces.
2. Otherwise, treat `UnexpectedEof` after a syntactically complete response the way curl and browsers do: return the data.

## Context

Found while moving teastia's OSV fetch (previously a curl script) into the binary. Happy to test a fix.

atzurody:
### Still reproduces on 0.59.2 — and **not only against Google-fronted hosts**

[...]

Issue #1561, opened 2026-08-24
native `http.request` fails after ~30 s with `read failed: Resource temporarily unavailable (os error 35)` on slow responses — timeout is not configurable (blocks local-LLM endpoints)


atzurody:
## What happened

`http.request("POST", …)` against an HTTP server that takes longer than ~30 seconds to start answering fails with:

```
err: read failed: Resource temporarily unavailable (os error 35)
```

os error 35 is `EAGAIN` on macOS — it looks like a read timeout on a non-blocking socket surfaced as a hard error. There is no parameter (or environment variable I could find) to raise the limit.

## Why it matters

Calling an OpenAI-compatible endpoint served by a **local LLM** (Ollama on the same machine) routinely takes 30–120 s of prompt evaluation before the first response byte on consumer hardware. Today any such call dies at the ~30 s mark, so an Almide program cannot talk to a local model unless the prompt is small enough to be processed in under ~30 s (measured on an M-series mac: a ~12 KB prompt with qwen2.5-coder:7b fits at ~24 s; 20 KB+ does not).

## Repro (almide 0.59.0, macOS arm64)

1. `python3 slow.py 45 8899 &` — a minimal server that sleeps 45 s, then answers 200 JSON (below).
2. `almide init`; add `[permissions] allow = ["Net", "IO"]`; `main.almd` (below) does one `http.request` POST.
3. `almide build && ./app http://127.0.0.1:8899/v1/chat/completions`

Expected: `ok: {"choices"…` after 45 s (or a *configurable* timeout).
Actual: `err: read failed: Resource temporarily unavailable (os error 35)` after ~30 s.

With `slow.py 5`, the same program prints `ok: …` — only the long wait fails.

## Ask

Either a configurable timeout (per-request option or env var) or blocking reads without the ~30 s cutoff. Happy to test a develop build.

Measured on the repro: fails at **31 s** with `slow.py 45`; succeeds with `slow.py 5`.

### main.almd

```almide
import http
import process

effect fn main() -> Unit = {
  let argv = process.args()
  let url = list.get(argv, 1) ?? "http://127.0.0.1:8899/v1/chat/completions"
  var hs = map.new()
  hs = map.set(hs, "Content-Type", "application/json")
  match http.request("POST", url, "{\"model\":\"m\",\"messages\":[]}", hs) {
    ok(body) => println("ok: " + string.take(body, 60)),
    err(e) => {
      eprintln("err: " + e)
      process.exit(1)
    },
  }
}
```

[...]

Issue #419, opened 2026-06-08
stdlib: string.index_of returns a byte offset but take/drop/char_at are char-indexed


O6lvl4:
## Symptom
Feeding `string.index_of` result into `string.take`/`string.drop` splits at the wrong place for multibyte (e.g. Japanese) strings.

## Confirmed in code (runtime/rs/src/string.rs)
- `almide_rt_string_index_of(s, sub) = s.find(sub)` → **byte offset**
- `almide_rt_string_take(s, n) = s.chars().take(n)` → **char count**
- `almide_rt_string_drop(s, n) = s.chars().skip(n)` → **char count**

So `take(s, index_of(s, sub))` mixes a byte offset with a char count → wrong for non-ASCII. The slicing API (take/drop/char_at/len/slice) is char-indexed; `index_of`/`last_index_of` are the odd ones out (byte-indexed). Fix: make index_of/last_index_of return a char index on BOTH targets (native + wasm), with a cross-target fixture.

## Workaround
Avoid index arithmetic; use `string.split`.

O6lvl4:
Deferred — deeper than a one-liner. Probing (`"a"+from_codepoint(233)+"=b"`) showed `string.len` returns a BYTE count (contradicting the cheatsheet's "char count"), and the byte-vs-codepoint unit is muddled across `len`/`index_of`/`last_index_of`/`slice`/`take`/`drop` (`char_at` isn't even exposed on the `string` surface). A correct fix unifies the unit to codepoints across ALL of them, on BOTH targets, byte-identical, with a multibyte `spec/wasm_cross/` fixture (none exists today, which is why the divergence went unnoticed). A native-only char-ified draft was attempted but is buggy — it made `index_of` return the wrong value and diverge native↔wasm — so it must not be applied as-is.

PR #1771, opened 2026-09-01
The http client core becomes almide-rt-core: spliced and linked from one source

### LINKED ISSUE #1715: runtime/rs becomes a linkable crate so the embedded host stops carrying a textual copy of the http client

O6lvl4:
## Summary
Repays #1715 exactly as filed. `crates/almide-rt-core` is a real workspace crate holding the prelude-independent http client core — parse/timeout/framing (the #1592 close-tolerance, the #1561 timeout wording), TLS, and all three exchange/result shapes (String, status-preserving, bytes) with headers as a plain `&[(String, String)]` slice.

**One source, two consumers**:
- `runtime/rs/src/http.rs` `include!`s the core file; the embed resolver (crates/almide-codegen/buildscript) inlines it at generation time, so the splice template's client IS this text. The AlmideMap-facing entry points became 3-line adapters over a `header_pairs` conversion; the SSE stream client and the server stay in http.rs and call the inlined core primitives.
- `almide-wasm-run` links the crate and serves fs_call ops 43..=47 through the same `request` — **`http_client.rs` (the pinned textual copy) is deleted**, per its own header's retirement condition.

The C-328 equality (error texts, timeout wording, close-without-close_notify tolerance, chunked framing) now holds by SHARED CODE; the embedded_cross behavioral net still pins the observable and runs green through the swap.

**Splice discipline** encoded in the core file itself: no `crate::` paths, TLS types fully qualified (the flat-splice assembler dedups only exact `use` lines — a group-vs-group overlap was E0252, caught by the cargo-path probe), io/net imports split as complements between the core and http.rs's remainder. The registry gains `cargo:rerun-if-changed=crates/almide-rt-core/src` so a core edit re-embeds (the #419 desync class); `rust_runtime.rs` regenerated and committed in the same PR.

**Also shared now, beyond the issue's minimum**: the status and bytes exchanges (previously duplicated framing inside http.rs) — one framing walk serves all three result shapes, and #1710 PR B can link them for the embedded host without another transcription.

## Evidence
- `cargo test --release --test embedded_cross_test` green (the C-328 net, both lanes on the same probes).
- Cargo-path smoke: an `http.get`/`get_status` program compiles and answers `connection failed: Connection refused (os error 61)` — the same texts as before the split.
- `almide test` 427/427; `cargo test --workspace` green; clippy clean on the new crate and almide-wasm-run.
- The generated-project dependency detection (`almide_rt_http_` / `use rustls` markers in cargo_build.rs) still fires: the inlined text carries both.

Closes #1715.


O6lvl4:
Superseded by #1775: identical tree, linear history — the rebase queue removed this branch with merge_conflict on every entry because of the stack-lineage merge commits and patch-duplicate commits it carried.

PR #1779, opened 2026-09-01
The epilogue releases each local once: mut-param writebacks no longer double-free the returned buffer

### LINKED ISSUE #1770: mut-param forwarding: passing a mut param through to another user mut fn drops the outer writeback in spliced modules
Found during #1700 (the first registry module with `mut` params): a helper shape

```almide
fn push_u16(mut out: Bytes, v: Int) -> Unit = { bytes.push(out, ...) ... }
fn stream(data: Bytes, mut out: Bytes) -> Unit = {
  bytes.push(out, final)
  push_u16(out, take)      // ← the nested mut-fn call
  ...
}
```

compiled on the structural leg but the bytes written by the OUTER fn's own `bytes.push` calls before/after the nested call were LOST (read back as 0) — the write-back chain for a mut param passed as a mut ARGUMENT to another rewritten mut fn mis-wires in the spliced-module context. Direct `bytes.push` (the intrinsic special case) is fine; only user-mut→user-mut forwarding breaks. zlib_deflate.almd works around it by inlining the helpers (see the PR); the C-132 move-mode rewrite (mut_param.rs rewrite_call_sites) is the suspect — note the same PR fixed the sibling per-module var-table bugs there, but the forwarding shape still mis-behaved after those fixes and needs its own probe.

Repro sketch: registry-splice a module with the shape above, call it from a wasm-leg program, read back the header bytes.

O6lvl4:
Fixes #1770 — and the class turned out broader than the report: no splice needed, ANY user program with the shape miscompiled on the structural leg.

## The bug

```almide
fn push_u16_le(mut out: Bytes, v: Int) -> Unit = { bytes.push(out, lo); bytes.push(out, hi) }
fn stream(mut out: Bytes, n: Int) -> Unit = {
  bytes.push(out, 1)
  push_u16_le(out, n)        // ← writeback makes `out` an rc_owned member
  … n payload pushes …       // ← buffer reallocates here
}
```

native `1,16,0,239,255,…` — wasm `0,0,0,0,255,…`.

The mut-param writeback's Assign registered the PARAM's local in `rc_owned`; the epilogue then released it in BOTH passes (the rc_owned pass and the droppable-param pass) — a double dec that freed the buffer the fn was returning. The freed block's freelist next-link write zeroed its first payload word, which is exactly the 4-byte hole. Sub-realloc sizes appeared fine only because nothing re-served the bogusly freed block before the read.

Found by restoring zlib_deflate's natural helper shape (the #1700 workaround was inlining), bisected with a no-free-patched artifact (correct output ⇒ free-while-live confirmed), then read directly out of the disassembly: the epilogue dec'd local `$0` twice.

## The fix

Sink-side dedupe: the epilogue's param pass skips locals the Bind/Assign routes already put in `rc_owned`, so each local is released exactly once — for every fn kind (a first guard at the insert sites failed for module fns, whose `rc_droppable_params` tail-release vec is deliberately empty; the sink is the only place both passes meet).

## Evidence

- `spec/wasm_cross/mut_param_writeback_growth.almd` (C-226, statement extended with the addendum) — the user shape, byte-identical on every leg.
- `stdlib/zlib_deflate.almd` returns to the forwarding shape (`zld_push_u16_le` / `zld_push_byte`, two levels of user-mut forwarding) — the C-331 round-trip fixtures now pin the class through a real consumer; `inflate(deflate(x))`, `decompress(compress(x))`, `gunzip(gzip(x))` all true on wasm.

## Verification

`almide test` 427/427; `cargo test --workspace --release` green; alloc/size ledgers restamped (the new fixture + the reshaped zlib rows); contract gate 651/651 bidirectional.


O6lvl4:
Superseded by #1780, which carries these exact commits rebased onto current develop with the generated contract docs re-derived — the queue rejected this branch's merge with a stale conformance.md (the merge-ref regen class). #1770 closes when #1780 lands.

Issue #1461, opened 2026-08-16
Pattern language is last in the cohort: no or-patterns, no as-patterns, no list-rest patterns


O6lvl4:
Found by the compiler-championship survey, which counted pattern constructors across ten compilers.

## The measurement

| language | pattern constructors | or | as | list-rest |
|---|---:|:--:|:--:|:--:|
| roc | 18 | ✓ | ✓ | ✓ |
| gleam | 13 | ✗ | ✓ | ✓ |
| moonbit | 12 | ✓ | ✓ | ✓ (+ string-spread) |
| grain | 11 + 3 | ✓ | ✓ | ✓ |
| **almide** | **11** | **✗** | **✗** | **✗** |
| koka | 6 surface | ✗ | ✓ | desugared |

Almide is last among the languages that have a real pattern language. Koka is smaller but compensates with a *list* of guards per branch; Almide has single guards.

Source: `crates/almide-syntax/src/ast.rs:110-121`.

## Why these three specifically

They are the three every writer reaches for, and each has a bad workaround today:

**or-patterns** — `Red | Green => ...`. Today: duplicate the arm body, or fall back to a guard. Duplicating an arm body is exactly the edit an LLM gets wrong when the two copies drift.

**as-patterns** — `p @ Point { x, .. } => use(p, x)`. Today: match, then rebuild the value from its parts, or match on a binding and re-destructure inside. The rebuild is a correctness hazard when the type gains a field.

**list-rest patterns** — `[first, ..rest]`. Today: `list.head` + `list.drop(1)` with the emptiness check written by hand. This is the single most common list idiom in every other language in the cohort.

## The MSR argument

Pattern matching is the construct `docs/CHEATSHEET.md` tells models to prefer over if/else chains, so it is disproportionately represented in generated code. A missing constructor does not produce a clean error — it produces a *workaround*, and workarounds are where modification survival goes to die: the duplicated or-arm that drifts, the rebuilt record that loses a new field.

## Definition of done

Three constructors, each a self-contained change through parser → checker (exhaustiveness!) → lowering → both renderers, with the usual obligations:

- exhaustiveness checking must understand each new form (`check/exhaustiveness.rs`)
- `almide fmt` must round-trip them (the AST-conservation verifier will enforce this)
- `spec/lang/` coverage plus a `spec/wasm_cross/` fixture per form, since these lower differently on the two legs
- ALS spec sections + `docs/CHEATSHEET.md`

[...]

PR #1780, opened 2026-09-01
List-rest patterns: [h, ..t] binds the tail on every leg

### LINKED ISSUE #1770: mut-param forwarding: passing a mut param through to another user mut fn drops the outer writeback in spliced modules

O6lvl4:
#1461's second constructor (or-patterns already shipped in 0.60): `[h, ..t]` matches any list of length >= the prefix and binds the tail as a list of the subject's type; `[h, ..]` ignores it; the rest slot is LAST-only (`[..t, x]` refuses with the rule named). This is the single most common list idiom in the cohort — the recursive head/tail split now reads:

```almide
fn sum(xs: List[Int]) -> Int = match xs {
  [] => 0,
  [h, ..t] => h + sum(t),
}
```

**Stacked on #1779** (the mut-param double-free fix) — the queue drops the shared commits when it lands.

## Per leg

- **Rust**: ListPatternLowering emits `len >= N` + a `list.drop(subject, N)` bind — and gains a **length-coverage ladder**: guard-free irrefutable arms cover their lengths (exact) or everything above their prefix (rest); an arm whose coverage swallows the uncovered region is TERMINAL, emitted as the chain's else. This closes the else-Unit fall-off the exact-length chain hit at rustc (a fully-covered `[] / [x] / [a, b, ..]` chain needed no wildcard yet failed E0308 before).
- **Structural wasm**: the block-length test relaxes to `>=`; a named tail materializes as a fresh block (alloc + payload copy, the list.drop shape). Deliberately NOT epilogue-owned: registering it freed a returned tail under its reader (the freelist link zeroed element 0 — caught by the string-tail probe), so it follows the drop result's leak-not-dangle discipline.
- **Interp**: slices the value vector; the 3-way oracle judges the cross fixture.
- **Incumbent**: walls honestly (`list-rest pattern binds are outside this brick`) — the structural leg is the default route; the test harness's incumbent lane sends rest files to the authoritative native fallback.

The IR carries the rest slot as a boxed sub-pattern (`Bind`/`Wildcard`), so every visitor, VarId remap and bind collector recurses with no special case. A named rest counts as a binder for the or-pattern E080 rule. The old three-dot hint now teaches the two-dot spelling (try_snippet_test updated, with a mirror probe asserting `[h, ..tail]` checks clean).

## Record

C-332 (ALS-E18 amended — list patterns and the rest form join the 受理形), CHEATSHEET pattern table + a checkable head/tail fence, fixtures on three suites (`spec/wasm_cross/list_rest_pattern.almd` byte-identical across legs, `spec/lang/list_rest_pattern_test.almd`, `tests/diagnostics/list-rest-not-last`), manifests + alloc/size/surface restamped (the structural surface gains `pattern:List`).

## Verification

`almide test` 428/428; `cargo test --workspace --release` green (wall-span ratchet satisfied, surface matrix restamped); clippy at the develop baseline (1423); fmt gate green; `check-contracts.sh` 332 contracts bidirectional.

Remaining for #1461: as-patterns (`p @ pattern`) — the last constructor.

Fixes #1770 — this PR carries the superseded #1779's epilogue fix (the mut-param writeback double-free) rebased onto current develop.


O6lvl4:
Carries #1779's mut-param double-free fix (its two commits are the base of this branch) — this PR **fixes #1770** as well as landing list-rest patterns.

Issue #1844, opened 2026-09-02
The interp oracle abstains on html.*, url.parse and a Result handle: three gaps the chain fixtures exposed


O6lvl4:
Found by regenerating `crates/almide-interp/interp-abstain-ledger.txt` over the 2026-09-02 fixture chain (#1817, #1814, #1837). Each PR records its own line in the same PR, as the gate demands; this issue is the widening work that shrinks the ledger back.

| fixture | abstain reason | root cause |
|---|---|---|
| `availability_pure_batch` (#1817) | `out-of-interp-scope capability: html.escape` | `stdlib/html.almd` is not a `self_host_registry` source, so the interp's `stdlib_pool` has no body for it; `html` (and its `mod type SafeHtml = String` newtype) are served on the structural leg by IR erasure only |
| `stdlib_type_shadow` (#1837) | `out-of-interp-scope capability: url.parse` | `url` is likewise absent from the pool / `bridge.rs` |
| `fs_glob_segments` (#1814) | `out-of-interp-scope capability: prim.handle of a Result (outside the slice-2 heap family)` | the prim heap floor models the slice-2 family only; a `Result` payload handle has no shape |

**Expected.** Preferred fix per the ledger header: widen the glue (`bridge.rs` / `hofs.rs` / `dispatch.rs`, see `crates/almide-interp/CLAUDE.md`) so the three fixtures evaluate, then remove the three ledger lines (the gate fails on stale entries, so the shrink is enforced). For `html`/`url` the cheapest route is making their `.almd` sources lowerable in the pool (the newtype needs the same erasure the structural leg applies, or a pool-side single-variant unwrap); for the `Result` handle, extend the prim heap family.

Repro: `ALMIDE_UPDATE_INTERP_LEDGER=1 cargo test --release --test wasm_runtime_test interp_abstain_ledger` on develop once the three PRs have landed, then delete a line and watch the gate report it as unledgered.

O6lvl4:
[...]

PR #1862, opened 2026-09-03
An opaque alias declared under a stdlib-owned name or inside a module takes a scoped identity, so the checker refuses mixing it with the stdlib type

### LINKED ISSUE #1835: An opaque mod type named after a stdlib-owned type still shares the stdlib's bare identity
Found while landing #1828 (a user type named after a stdlib-owned type takes the `self.X` / `m.X` identity). That landing deliberately leaves one shape on the old bare-key path: an OPAQUE alias — `mod type X = String` / `local type X = …` — declared under a stdlib-owned name.

## The gap

```almide
import json
mod type Value = String
fn mk(s: String) -> Value = Value(s)
fn un(v: Value) -> String = match v { Value(s) => s }
effect fn main() -> Unit = {
  let doc = json.parse("[1]")!
  println("${un(mk("x"))} ${json.stringify(doc)}")
}
```

| leg | result |
|---|---|
| `almide check` | accepted (the user's `Value` and the json `Value` share the bare key, as every shape did before #1828) |
| native | rustc E0308 — `json.parse` hands back `AlmideValue`, the binding is the user's newtype (ALS-T6, check-passes/build-fails) |
| `--target wasm` | prints `x [1]` here, but a program that `match`es the json value against `Value(s)` type-checks and reads the payload as a String |

## Why it stayed on the bare key

The opaque newtype's constructor call and pattern are spelled by the BARE name on the native leg (`Value(s)` against a struct the flatten pass named `almide_rt_self_Value(String)`), and the wasm legs wall the form for any module scope — a module's own `mod type Token = String` + `Token(s)` fails the same way on 0.61.1 (native E0531/E0425, wasm "both legs refused"). Giving the opaque shadow the scoped identity needs the opaque-ctor call target and the `Constructor` pattern to carry the canonical name through lowering, the flatten rename, MIR and the wasm renderer — the module case first, since it is the same plumbing.

Until then: `spec/lang/opaque_type_test.almd` (`mod type SafeHtml = String` beside html's) keeps passing on the old path, and `stdlib_info::STDLIB_OWNED_TYPES` documents the boundary at `register_type_decl`.

O6lvl4:
Fixes #1835.

#1837 gave a user `type X = …` that shadows a stdlib-owned name the scoped identity `self.X` / `m.X`, but left the OPAQUE alias `mod type Value = String` on the bare key: its constructor call and pattern were spelled by the bare name, so `almide check` accepted a program mixing the user's `Value` with json's, native failed in rustc (E0308, check-passes/build-fails) and the wasm legs could read a json value's payload as a String.

**Design.** `opaque_alias_identity` gives the newtype the same scope as records (`self.X` for an entry-program shadow of a stdlib-owned name, `m.X` for any user module, bare for bundled modules, entry-owned modules and plain entry aliases); the three `opaque_alias_*` tables key on it and `TypeEnv::opaque_alias_key` resolves a spelled `Value(..)` back to it (own module → shadow → bare → unique foreign for E033). Lowering pins that identity on the ctor call target and the ctor pattern, `lower_type_decl` names the decl the same way, the flatten pass renames the call target and pattern with the struct, and the wasm/MIR newtype erasers key on the decl name so they erase unchanged. A new **E048** (with the E013 shadow note) refuses a user ctor or record pattern aimed at the stdlib's bare type — this also closes the sibling hole where `Value { n }` against a json value passed `check`.

| program | check | native | wasm structural | wasm incumbent |
|---|---|---|---|---|
| issue repro (`un(mk("x"))` + json) | accepted → accepted | rustc E0308 → `x [1]` | `x [1]` | `x [1]` |
| `un(doc)` (json Value into the user newtype) | accepted → **E005** `expects self.Value but got Value` | E0308 → n/a | garbage → n/a | n/a |
| `match doc { Value(s) => s }` | accepted → **E048** | E0308 → n/a | payload read as String → n/a | n/a |
| module `mod type Token` + `Token(s)` | accepted | rustc E0531 → `x` | refused → `x` | → `x` |

**Evidence.** E013/E048 diagnostic fixtures; `spec/wasm_cross/stdlib_opaque_shadow.almd` (C-076 `EXTENDED (#1835, 0.61.2)`, byte-identical on native / structural / incumbent); `spec/integration/modules/opaque_module_type_test.almd` + `opaquelib/`; docs (E013.md, E048.md, module-system.md §4.2). `spec/lang/opaque_type_test.almd` keeps passing on both lanes; its `check` output loses a spurious E015 (the user's `SafeHtml` is now `self.SafeHtml`) — the one non-additive check-manifest row.

**Gates.** frontend/types/tools/wasm crate tests, diagnostics harness 6/6, interface-spelling and owned-types pins, `almide test` spec/lang 208 · integration 52 · stdlib 160, cross-target gate, alloc/size ledgers, check-contracts, abstain classes (the third oracle abstains on the fixture: a named call `self.Value` — ledgered BRIDGEABLE, closed by the #1844 follow-up), fmt, codopsy A (frontend, codegen). The incumbent runs the fixture without a wall (no walled-real row, MAX_WALLED unchanged).


O6lvl4:
Closing: this branch's content landed in develop through #1854, which was stacked on it in the merge queue (commits 4c4f26377, 7096e846b, dfabdb1be, 13ff0241b are on develop). The remaining diff of this PR against develop is ledger churn only.

Issue #1127, opened 2026-08-06
Pipe-position ?? over a Result-returning call lowers an Option-shaped match (native E0308)


O6lvl4:
Found while landing #1108's spec tests. Pre-existing — reproduces on the released v0.53.6.

## Repro

```almide
fn parse1(s: String) -> Result[Int, String] = int.parse(s)

fn main() -> Unit = {
  let r = ["1", "2"] |> list.try_map((s) => parse1(s)) ?? [0]
  println("r=${r}")
}
```

Native codegen emits `match <try_map call> { Some(v) => v, None => ... }` — an **Option-shaped unwrap over a `Result<Vec<i64>, String>`** — and rustc rejects with E0308 ("expected Result, found Option"). check accepts, so this is the acceptance-parity class ("check green, build red").

The DIRECT spelling (`let r = list.try_map(...) ?? [0]`, no pipe) lowers correctly; only the pipe path picks the Option-shaped unwrap for a Result operand. The polymorphic form `xs |> list.map((s) => f(s)!) ?? fb` hits the same path post-#1108 (its spec test uses the match spelling meanwhile).

A/B: identical failure on v0.53.6 (`scratchpad` repro `pipeqq.almd`).

O6lvl4:
修正済み(2c81e5da、0.56.1 予定分): `infer_pipe_unwrap_or` を直呼び規則と鏡映に — payload を先に剥がして fallback と constrain(旧: 包み全体と unify)+ pipe-`!` 枝と同じ type_map insert で lowering が Result 形を選べるように。A/B: v0.55.0 で E0308×3 → 修正後 r=[1,2]/bad=[0]。wasm は既知の heap-result-?? frontier へ honest wall。回帰: spec/lang/fallible_hof_test.almd の pipe-?? テスト。

PR #1872, opened 2026-09-03
ADR-0005 D6/D1/D3/D5: the pipe × ?? desugar order is fixed and pinned, the operator equivalences are contract fixtures, no-op ? warns

O6lvl4:
Refs #1107

## What this PR does

The four bounded ADR-0005 items were audited against develop; three of them (D1, D3, D5) plus D2 had already landed on develop in 0.57.x via 35119908c / 9938810fc but the umbrella's checklist was never ticked. This PR lands the remaining bounded item (D6, the pipe × `??` miscompile) as a cross-target family pin, adds the one missing D2 documentation line, and ticks the issue with evidence. D4 (eager `unwrap_or` deprecation + removal) is deliberately left open: it is a dialect change with a deprecation window.

### D6 — pipe × `??` (the BUG-ledger cells)

**Which leg was wrong and why.** The parser binds `??` tighter than `|>`, so `xs |> stage ?? fb` is `Pipe(xs, UnwrapOr(stage, fb))`. `infer_pipe` re-associates that to `UnwrapOr(stage(xs), fb)`, and the lowering's `lower_pipe` mirrors it — but the lowering needs to know whether the piped operand is an Option or a Result to pick the some/none vs ok/err match. Through 0.53.6 it guessed Option when the checker had not annotated the operand. For a Result operand:

- **native**: emitted an Option-shaped `match … { Some(v) => v, None => … }` over a `Result<Vec<i64>, String>` — rustc E0308, build red (the "check green, build red" class)
- **wasm**: read the Result block as an Option — `["1", "x"] |> list.try_map(parse) ?? [0]` printed a list of heap garbage with **exit 0** (the silent-wrong cell the ADR cites)

The cause was removed by #1127 (the checker annotates the piped operand with its RESOLVED type before the lowering's pipe transform — the "desugar before the pipe transform" the ADR asks for, in effect) and the fn-tail position by #1134 (C-218's tail rewrite). Both landed in 0.56.0; the issue's checklist was never closed and no cross-target fixture pinned the pipe spelling as a family.

**The pin**: `spec/wasm_cross/pipe_unwrap_or_family.almd` on C-218 (statement `EXTENDED (#1107, 0.61.2)`): the ledger cells in bind position, as a fn tail, and piped on after the fallback, plus the scalar-Option pipe shapes (ident target, call target, filtered chain, user Result fn, mid-chain `?? 0 |> next`, heap-Option payload, `??` inside a lambda stage, fallback laziness). Byte-identical native == wasm; 3-way vote (interp evaluates it, no abstain row); incumbent renderer emits it too (harness rc 0), so no walled-real row / MAX_WALLED bump.

**A/B evidence**

| binary | `["1","x"] |> list.try_map(parse) ?? [0]` (0.53.6 spelling) | the fixture |
|---|---|---|
| v0.53.6 (release asset) | native: rustc E0308 (Option match over Result); wasm: `b=[2334106421097295465, 8027139070191167844, …]` exit 0 | E022 at check (pre-ADR-0006 spelling) |
| 0.61.1 (installed) | n/a (`try_map` removed in 0.56) | passes, byte-identical on both legs |
| this branch | — | passes, byte-identical on both legs |

Note for the reviewer: the fixture therefore does NOT fail on 0.61.1 — the defect was fixed in 0.56.0; the A/B the issue itself asks for (v0.53.6 → wrong value) is what is reproduced above. I probed ~45 further pipe × `??` shapes on 0.61.1 (bind / tail / arg / interpolation / if-arm / match-arm / loop / lambda body / Result and Option / heap and scalar payloads); every one that compiles is byte-identical on both legs. Two pipe-adjacent defects turned up that are NOT `??` related and are filed separately rather than folded in here (see the issue links in the follow-up comment).

Dropped from the fixture on purpose: `xs |> list.map(f!) |> result.map(g) ?? d` (a Result STAGE after the fallible map) walls on the INCUMBENT renderer only (`?? over an Option/Result operand in tail position outside the executable subset`) while the structural leg and native agree; keeping it would have walled the whole fixture on the incumbent and cost the D6 cells their incumbent coverage. It is an incumbent brick gap, not a pipe-desugar defect.

### D1 / D3 / D5 / D2 — already on develop, now ticked

- D1: C-291 + `spec/wasm_cross/adr0005_operator_equivalence.almd` (both spellings side by side for `??` Option/Result, `?`, `?.`), spec table in `docs/specs/result-option-effect.md`.
- D3: landed as **E056** — a `Diagnostic::warning` (the repo numbers warnings in the E series: E035/E036/E052 are warnings too, there is no W series), with `tests/diagnostics/e056-noop-to-option` broken (positive) / fixed (negative). No new code was taken.
- D5: `docs/design/REJECTED_PATTERNS.md` `??` entry reversed (ADR-0005).
- D2: E055 exists with its fixture; the spec has the `o?.x ?? 0` idiom; this PR adds the `o?.x ?? default` line and the desugar table to the CHEATSHEET (the one sub-item that was missing).

## Ledgers touched

run / check / ast manifests, alloc and size baselines (one row each), contracts.toml + regenerated conformance.md / README.md. surface matrix and interp-abstain ledger regenerated with no change. `gen-ledger-counts.sh` NOT run. No Rust source touched, so clippy baseline / codopsy are unaffected.

## Gates run locally (all green)

- `cargo test --release --test wasm_runtime_cross_target` (668 fixtures, 437s)
- `cargo test --release --test wasm_runtime_interp_oracle`
- `./target/release/almide test spec/lang/` and `spec/stdlib/` (160 files, all pass)
- `cargo test --release --test diagnostic_harness_test` (6 passed), `--test traversal_totality_lint`
- `bash scripts/check-contracts.sh` (335 contracts, 668/668 fixtures linked), `check-readme-numbers.sh`, `check-doc-fences.sh`
- `almide fmt --check` on the fixture
- host-determinism: not run in full (wasm32 harness build); the native harness emits the fixture (rc 0, no wall).

Auto-merge intentionally NOT enabled (fixture ledgers move; develop's MAX_WALLED is 25 and untouched here).


O6lvl4:
Pipe-adjacent defects found while probing for D6, filed separately (not `??` related, both are check-green/build-red, neither is a silent wrong value): #1873 (tuple-pattern lambda as a pipe stage) and #1874 (`m[k]` with a String parameter key renders `&&str` natively).

Issue #1488, opened 2026-08-16
Adopt roc's semver API differ: fail the release unless the version bump matches the interface diff


O6lvl4:
Found by the compiler-championship survey while comparing package tooling.

## The mechanism

roc ships `roc bump` — a **semver API differ**, 2,457 lines (`src/cli/cli_args.zig:1300`). It compares the package's public interface against the previous released version, computes the smallest legal version bump, and `--expect X.Y.Z` **fails CI unless the declared version bumps at least that far**.

So "I added a breaking change and released it as a patch" becomes a build failure rather than a downstream surprise.

## Why this fits Almide unusually well

We already have both halves and have never joined them:

- **`almide compile --json`** emits the module interface, and `tests/interface_stability_test.rs` (landed this week) proves that interface is a function of **signatures only** — a body change does not move it, a signature change does. That test exists precisely so this comparison is meaningful.
- **`proofs/dialect-epochs.toml`** already records language-surface breaks as a monotone epoch.
- **`scripts/release-seal.sh`** already freezes per-release evidence and re-measures `[derived]` fields against the tag forever.

The missing piece is the join: compare the interface JSON at the release tag against the previous tag, classify the diff, and assert the version bump is at least as large.

## What "at least as large" means here

Almide is pre-1.0 and cuts releases every few hours, so full semver ceremony would be noise. The useful assertions are narrower:

1. **A removed or signature-changed public name requires a dialect epoch bump** (#1484's RC trigger) — this is the one that matters, and today nothing checks it.
2. **A purely additive interface change requires at least a patch bump** — trivially true today, but worth asserting so it stays true.
3. **A release whose interface is byte-identical to the previous one** is fine and should say so, rather than being silently indistinguishable from a breaking one.

## Definition of done

1. `scripts/check-interface-diff.sh <prev-tag> <tag>`: run `almide compile --json` for every stdlib module at both revisions, diff, classify as `identical` / `additive` / `breaking`.
2. Wire the verdict into `scripts/release-seal.sh gen` as a `[derived]` field, so it is re-measured against the tag forever like every other seal field.
3. Assert rule 1 above: `breaking` without an epoch bump fails.
4. Negative controls proving each classification fires, per the gate-verification discipline.

Related: #1484 (RC channel) uses the same trigger — a dialect bump — so these two compose into one release rule: *interface diff says breaking ⇒ epoch bump ⇒ RC ⇒ soak ⇒ final.*

O6lvl4:
[...]

Issue #566, opened 2026-06-11
Measure and ratchet compiler structural coverage; MC/DC for safety passes


O6lvl4:
**Category**: A — engineering | **Status**: ◑ rung 1 (measure + ratchet) LIVE since 2026-07-04; rung 2 (MC/DC-grade evidence for the safety set) not started

The verification machinery covers emitted programs; the compiler's own branches were unmeasured. Unreached compiler code is either a dead branch (delete) or a missing fixture (add) — both outcomes improve the system.

## What is already running (this issue's first rung)

- `proofs/coverage.sh` — manual llvm-cov pipeline (`--check` / `--update`), toolchain-stamped; instruments the almide-mir + almide-codegen test suites, `render_program` over all runnable spec (v1 path) AND `almide test spec/` through the CLI (frontend→codegen production path). The header "Scope note" still says almide-mir-only — stale, to be rewritten.
- Ratchet: `proofs/coverage-baseline.txt` (currently 5819 = 58.19% TOTAL line; the TOTAL spans every workspace crate linked into the measured binaries, not only the crates shown as rows).
- CI: `.github/workflows/fuzz-nightly.yml` job "Compiler structural coverage ratchet (#566)", toolchain pinned 1.89.0, nightly `--check`. Measured duration of the last three runs: 14m06s / 14m07s / 11m18s.
- `proofs/COVERAGE.md` — 2026-07-04 table and low-coverage ledger (control_p5 43%, tail.rs 61%, binds_p2/p4 63–64%).

## Measured constraints for rung 2 (2026-08-26)

- **rustc's MC/DC instrumentation no longer exists.** `-Zcoverage-options=mcdc` was added in 2024 and REMOVED in Aug 2025 (rust-lang/rust#144999, maintenance burden + incomplete); on 1.96.1 the accepted values are `block | branch | condition` (probed). `condition` works end-to-end (probed: per-condition true/false counts including the RHS of `||`, which plain `branch` misses). Consequence: MC/DC-grade evidence must come from a per-decision argument (single-condition decisions + independence-pair test vectors), with `condition` coverage as the instrumented backstop — not from a compiler flag.
- **The originally named passes no longer exist under those names.** PerceusVerifyPass and StackBalancePass predate the v0 retirement (569a09de1 deleted the orphaned second Perceus); StackBalance is a Rocq-spine property today, not a Rust pass. The real safety set for MC/DC-grade evidence:
[...]

PR #1892, opened 2026-09-04
Two release-path gates read true again: interface-diff sees through the legacy placeholder rendering, and the per-condition coverage job finds its test binaries

O6lvl4:
Two release-path gates that had quietly stopped meaning what they say.

**interface-diff (#1488)** — `check-interface-diff.sh v0.61.1 develop` reported `breaking (added=63 removed=51)`. Every "removed" line is the same function re-rendered: indexes written before 394c64294 spelled every tuple as a bare `()` and every unnamed type as `?` (`list.zip(...) -> List[()]`, `bytes.as_ptr(b: Bytes) -> ?`), and the textual diff cannot see through that. Left alone, the 0.62.0 tag would have to declare `--allow-breaking` for a diff that breaks nothing. A prev-only line carrying a placeholder is now matched against the cur-only lines with the placeholder as a wildcard (`()` → any parenthesised tuple, `?` → any type; name and arity fixed); exactly one match reclassifies the pair as a re-rendering, reported on its own `re-rendered … — not counted` line. A genuine function type `() -> A` is not a placeholder (the `()` is followed by `->`), a placeholder line with no current twin stays a removal, and a twin whose parameter list changed still does not match. The real verdict is now `additive (added=12 removed=0)` with 51 re-rendered. Two controls join the negative script (legacy tuple → identical; legacy orphan → breaking).

**MC/DC per-condition measurement (#566)** — the fuzz-nightly job "Condition-level coverage" has failed every night since 2026-08-27 at step 2/4: `find <target-dir>/release/deps` finds nothing under the dated nightly's build-dir layout (`build/<crate>/<hash>/out/`), so the gate exits with "NO test binaries found" while the stable-toolchain run passes on the old layout. The test binaries are now taken from `cargo test --no-run --message-format=json`'s `executable` field (layout-independent); the `find` stays as the fallback and the no-binaries guard keeps its teeth.

### DIFF
--- a/proofs/coverage.sh
+++ b/proofs/coverage.sh
@@ -101,8 +101,19 @@ echo "== 1/4 instrumented build (almide-mir + almide-codegen + almide-wasm tests
 # default `--target wasm` leg (and `almide test`'s wasm phase workload) runs
 # the structural emitter, so a spine-crate-only measurement halves the TOTAL
 # while the system's actual hot path goes unmeasured.
+# The test binaries are taken from cargo's own artifact messages, not from a
+# `find` over `<target-dir>/release/deps`: the dated nightly the CONDITION
+# mode pins lays its artifacts out under `build/<crate>/<hash>/out/` and the
+# `deps` directory does not exist there — from 2026-08-27 every nightly
+# per-condition measurement died at step 2/4 with "NO test binaries found"
+# (8 red nights) while the stable-toolchain run kept passing on the old
+# layout. The JSON `executable` field is the one location that does not
+# depend on the layout; the `find` below stays as the fallback.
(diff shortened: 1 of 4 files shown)


O6lvl4:
Third commit: the first queue run failed `snapshot_update_test::a_plain_run_fails_with_the_accept_hint_and_writes_nothing` on shard 0 (one run in three — #1876's and #1893's runs passed). Root cause is in the T18 block parser, not the test: libtest prints `test <name> ... ` without a newline before running the test, the snapshot abort exits the process, and when that partial line WAS flushed the captured stdout+stderr reads `test tests::x ... Error: snapshot mismatch` — the header no longer starts its line, the block goes unrecognised, and the report falls back to raw output without the accept hint. `parse_t18` now strips libtest's progress fragment before reading the header (unit test with the glued form added).

PR #1895, opened 2026-09-04
list.range and list.len join the native v1 runtime floor

### LINKED ISSUE #1869: list.range is outside the native v1 runtime floor, so any program that materializes a range falls to the v3 codegen fallback
Found while fixing #1857 (PR #1868). A `let`-bound range that is indexed or measured (`idx[2]`, `list.len(idx)`) needs `list.range` to materialize; the native v1 trust-spine render cannot link it (`not in the native runtime floor`), so the WHOLE program falls to the v3 codegen fallback. #1868 made v3's head-only ranges count instead of materializing, so the 16 GiB `Vec<i64>` is gone, but the routing itself is still a hole: every such program leaves the certified leg for the trusted one, silently.

Expected: `list.range` joins the native v1 runtime floor (the same way the other list ctors did), so the program stays on the trust-spine; add the shape to the floor's pin test and re-measure `proofs/corpus-wall.sh` (the walled-real count should drop for the fixtures that only walled on `list.range`).

O6lvl4:
Closes #1869.

A `let`-bound range that is indexed or measured (`idx[2]`, `list.len(idx)`) lowers to `CallFn list.range`; the native v1 trust-spine render had no floor entry for it — nor for `list.len` over a scalar list — so the whole program left the certified leg for the v3 codegen, silently (`ALMIDE_VERIFIED_DEBUG` was the only witness). Both join `render_native_shims` as closed shims: `rt_list_range(s, e) -> Vec<i64>` (empty when `e <= s`, the self-host body's contract) and `rt_list_len(&[i64]) -> i64`. The closed-shim call binds a `Vec` result as `Vec<i64>` (it knew only `String` / `i64`), and the declared-mode arg check gains the borrowed-list mode (an owned `Vec` local by reference, a `&[i64]` param as is). The probe now reports `native: v1 trust-spine render` and prints the same bytes on both targets; `tests/native_floor_list_range_test.rs` pins the route and the output.

Two things that fell out once a capped program reached v1 for the first time:

- `inject_heap_cap` splits the generated source after the leading run of `#![...]` / blank lines to keep inner attributes first — but the v1 render opens with a `// Generated by …` comment BEFORE its `#![allow(..)]`, so the runtime landed ahead of the attribute ("an inner attribute is not permitted in this context"). The leading run now includes `//` lines. `heap_cap_test::native_cap_enforcement_is_live` is the witness (its churn probe uses `list.len`, and walled v1 until now).
- `heap_cap_test`'s walled-range probe (#1857) is only a witness while it walls v1; it moves onto a top-level `let` (rung 1 has none), same output, and its comment says so.

Re-measured: `proofs/corpus-wall.sh` is unchanged at 40 walled-real rows — that gate classifies the MIR lowering, and `list.range` never walled there; the drop the issue expected shows up on the native route, not in that ledger. `almide test spec/lang/` 208/208, almide-mir unit tests 631/631, the native test files green.

### DIFF
--- a/crates/almide-mir/src/render_native_b.rs
+++ b/crates/almide-mir/src/render_native_b.rs
@@ -372,7 +372,11 @@ fn render_closed_shim_call(
     match (dst, ret_ty) {
         (Some(d), Some(t)) => {
             tys.insert(*d, t);
-            let ty_name = if t == NTy::Str { "String" } else { "i64" };
+            let ty_name = match t {
+                NTy::Str => "String",
+                NTy::Vec => "Vec<i64>",
+                _ => "i64",
+            };
             line!("let mut {}: {} = {};", var(*d), ty_name, call);
         }
         (None, _) => line!("{call};"),
@@ -401,6 +405,13 @@ fn shim_arg_in_declared_mode(
             }
             Ok(code)
         }
+        // A scalar-list param (#1869): borrowed at the MIR call boundary —
+        // an owned `Vec<i64>` local by reference, a `&[i64]` param as is.
+        NTy::VecRef | NTy::Vec => match got {
+            NTy::Vec => Ok(format!("&{code}")),
+            NTy::VecRef => Ok(code),
+            _ => Err(wall(format!("native: shim `{name}` arg type mismatch"))),
+        },
         _ => {
             if !got.is_stringy() {
                 return Err(wall(format!("native: shim `{name}` arg type mismatch")));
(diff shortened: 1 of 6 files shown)


O6lvl4:
Third commit, after the first CI run: (1) `range_materialize_oom` diverged — the shim's `(s..e).collect()` died as a raw `capacity overflow` panic (exit 101) where the wasm leg and the v3 runtime answer the C-197 `Error: out of memory` abort (exit 1); the shim now carries `almide_rt_list_range`'s body verbatim (`saturating_sub` span, `try_reserve_exact`, the abort). (2) `native_v1_differential_test::out_of_subset_walls_honestly` used `list.len` over a literal as its out-of-subset witness; it now uses `list.reverse`, which is still outside the floor.

PR #1896, opened 2026-09-04
The interp serves a Result handle as the cap-as-tag block and stores its fs overlay as bytes

O6lvl4:
Part of #1844 — the `prim.handle of a Result` cell, which was ten of the ledger's rows.

The fs and fan floors hand their stdlib bodies a `Result` and the bodies read it through the arena: `load_handle(h + 12)` for the payload and `load32(h + 16)` for the tag. That block is the cap-as-tag family (`render_wasm_fs_wat` `$rtf_result`): one slot, `len@4 = cap@8 = 1`, the payload's u32 in the low half, the 0/1 tag in the high half. The interp had no shape for it and abstained at the handle. It now materializes a `Value::Result` exactly so — Unit as 0, Bool as 0/1, an Int only when it fits the half (a wider one abstains by name), a Str / List / Map / Tuple payload as its own bound block's address — and `load_handle` / `load_str` read 4 bytes, the `i32.load` both backends emit (`PrimKind::LoadHandle`); the old 8-byte read folded the tag into the address on exactly this block.

Getting past the handle exposed a second gap: `fs.write_bytes` fills an `alloc_str` block byte by byte and passes its ADDRESS to `prim.write_text_file`, and the overlay stored `String`s — the arm aborted ("expects (String, String)"), which the 3-way gate rightly reported as an interp dissent on two fixtures. The overlay now stores bytes; `read_text` answers the backends' `stream did not contain valid UTF-8` on non-UTF-8 content (the `fs_read_text_utf8` contract); the vfs prims take a `Str` or a Str-block address for the path and a `Str` or a Str/Bytes-block address for the content, and anything else is an honest abstain named by shape, never an abort.

Ledger: 10 rows → 8. `fan_effect_mapper_capability` and `fs_read_lines` evaluate and agree (650 of 677 fixtures `interp == native == wasm`, 0 dissent, 1 pre-existing backend split); the other eight now abstain one prim deeper, each named — `prim.read_dir` (×3), `prim.path_filestat` (×2), `prim.read_bytes_file` (×2), `fs.__fallible_fold_lines` — classified BRIDGEABLE (sandboxed-fs glue). The issue's html / url cells are no longer in the ledger (their fixtures evaluate on develop); `fs_glob_segments` now stops at `path_filestat`, so the issue stays open for the vfs listing/stat/bytes-read glue.

### DIFF
--- a/crates/almide-interp/interp-abstain-ledger.txt
+++ b/crates/almide-interp/interp-abstain-ledger.txt
@@ -13,7 +13,6 @@
 alias_combinator_rc  out-of-interp-scope capability: prim.handle of a Option (outside the slice-2 heap family)
 availability_pure_batch  out-of-interp-scope capability: html.escape
 bytes_rawptr  out-of-interp-scope capability: prim.int_to_ptr
-bytes_temp_receiver  out-of-interp-scope capability: in-place container mutation `bytes.append_i16_be` through a non-variable receiver (a record field / index / temporary has no single binding to write back to)
 count_domain_nonbytes  out-of-interp-scope capability: prim.read_n_bytes
 effect_tco_err_rewrap  interp fuel/recursion budget exhausted
 env_set_overlay  out-of-interp-scope capability: env.set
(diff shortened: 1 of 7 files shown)


O6lvl4:
Superseded by #1897: this PR was opened on the `fix-1844` branch name, which another in-progress branch already owned; that branch is restored to its own commit (1f2abb5d8) and this work moved to `fix-1844-result-handle`.

PR #1897, opened 2026-09-04
The interp serves a Result handle as the cap-as-tag block and stores its fs overlay as bytes

### LINKED ISSUE #1844: The interp oracle abstains on html.*, url.parse and a Result handle: three gaps the chain fixtures exposed

O6lvl4:
Closes #1844 — started as the `prim.handle of a Result` cell (ten of the ledger's rows) and, one prim deeper each pass, ends with every one of them evaluating.

The fs and fan floors hand their stdlib bodies a `Result` and the bodies read it through the arena: `load_handle(h + 12)` for the payload and `load32(h + 16)` for the tag. That block is the cap-as-tag family (`render_wasm_fs_wat` `$rtf_result`): one slot, `len@4 = cap@8 = 1`, the payload's u32 in the low half, the 0/1 tag in the high half. The interp had no shape for it and abstained at the handle. It now materializes a `Value::Result` exactly so — Unit as 0, Bool as 0/1, an Int only when it fits the half (a wider one abstains by name), a Str / List / Map / Tuple payload as its own bound block's address — and `load_handle` / `load_str` read 4 bytes, the `i32.load` both backends emit (`PrimKind::LoadHandle`); the old 8-byte read folded the tag into the address on exactly this block.

Getting past the handle exposed a second gap: `fs.write_bytes` fills an `alloc_str` block byte by byte and passes its ADDRESS to `prim.write_text_file`, and the overlay stored `String`s — the arm aborted ("expects (String, String)"), which the 3-way gate rightly reported as an interp dissent on two fixtures. The overlay now stores bytes; `read_text` answers the backends' `stream did not contain valid UTF-8` on non-UTF-8 content (the `fs_read_text_utf8` contract); the vfs prims take a `Str` or a Str-block address for the path and a `Str` or a Str/Bytes-block address for the content, and anything else is an honest abstain named by shape, never an abort.

Ledger: 10 rows → 8. `fan_effect_mapper_capability` and `fs_read_lines` evaluate and agree (650 of 677 fixtures `interp == native == wasm`, 0 dissent, 1 pre-existing backend split); the other eight now abstain one prim deeper, each named — `prim.read_dir` (×3), `prim.path_filestat` (×2), `prim.read_bytes_file` (×2), `fs.__fallible_fold_lines` — classified BRIDGEABLE (sandboxed-fs glue). The issue's html / url cells are no longer in the ledger (their fixtures evaluate on develop); `fs_glob_segments` now stops at `path_filestat`, so the issue stays open for the vfs listing/stat/bytes-read glue.

### DIFF
--- a/crates/almide-interp/interp-abstain-classes.toml
+++ b/crates/almide-interp/interp-abstain-classes.toml
@@ -61,10 +61,10 @@ patterns = [
   # next vfs prim the overlay does not serve — a directory listing, a stat,
   # a bytes read, the fallible fold callback. Sandboxed-fs glue at the
   # argv/env tier, bounded work each.
-  "capability: prim.read_dir",
-  "capability: prim.path_filestat",
-  "capability: prim.read_bytes_file",
   "capability: fs.__fallible_fold_lines",
+  # One prim deeper again once the overlay served read_dir / path_filestat /
+  # read_bytes_file: the rename floor is the next bounded piece of vfs glue.
+  "capability: prim.rename",
   # SURFACED BY the #1416 ctor glue (map.new/set.new/with_capacity), not
   # introduced by it: map_filter / map_set_typechange used to abstain at
   # `prim.alloc_map` (HEAP_BOUNDARY, first-hit) and now get past the
--- a/crates/almide-interp/interp-abstain-classes.toml
+++ b/crates/almide-interp/interp-abstain-classes.toml
@@ -55,6 +55,16 @@ patterns = [
   # tier, same class as random_get/read_n_bytes — bounded, not architectural.
   "capability: env.set",
   "capability: HOF map.fold",
+  # SURFACED BY the #1844 Result-handle block (`prim.handle` of a Result is
+  # now the cap-as-tag slot the fs floors hand back), not introduced by it:
+  # the fs fixtures used to abstain at the handle and now get past it to the
+  # next vfs prim the overlay does not serve — a directory listing, a stat,
+  # a bytes read, the fallible fold callback. Sandboxed-fs glue at the
+  # argv/env tier, bounded work each.
+  "capability: prim.read_dir",
+  "capability: prim.path_filestat",
+  "capability: prim.read_bytes_file",
+  "capability: fs.__fallible_fold_lines",
   # SURFACED BY the #1416 ctor glue (map.new/set.new/with_capacity), not
   # introduced by it: map_filter / map_set_typechange used to abstain at
   # `prim.alloc_map` (HEAP_BOUNDARY, first-hit) and now get past the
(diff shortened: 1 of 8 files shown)


O6lvl4:
Follow-up commits: the overlay now serves `prim.read_bytes_file` (the bytes as written), `prim.path_filestat` (filetype@16 / size@32 / mtim@48 into the caller's scratch, WASI ENOENT 44 for a missing path) and `prim.read_dir` (overlay children merged with the real directory, `names.sort()`ed like both legs, native ENOENT / ENOTDIR). Oracle: 655 of 677 agree, 21 skipped, 0 dissent. Ledger 8 → 3 rows: `fs_composition_family` and `fs_write_errno` stop one prim deeper at `prim.rename`, `fs_fallible_stream_callback` at the fold callback.

O6lvl4:
Final state of this PR: the overlay also serves `prim.rename` (subtree move, native ENOENT for a missing source, host-only source abstains) and `prim.path_filestat_nofollow`, and answers the native ENOTDIR for a write / mkdir through a plain file (the `fs_write_errno` cases the first pass got wrong). Full oracle: **657 of 677 agree, 19 skipped, 0 dissent**; abstain-classes voting 658 (97.2%). Ledger from the original 10 Result-handle rows: one remains (`fs_fallible_stream_callback` at the fold callback). Also adds `ALMIDE_CORPUS_FILTER=<substring>` to the 3-way harnesses for a single-fixture developer loop (seconds instead of ~6 min; never set in CI).

O6lvl4:
Last commit: `fs.__fallible_fold_lines` joins the interp's HOF surface (overlay read, `read_line` split, first-err short-circuit) — `fs_fallible_stream_callback` agrees on all three legs. Every row this PR started from now evaluates; the ledger is down to the pre-existing 18 rows, so this closes #1844.

PR #1898, opened 2026-09-04
Version 0.62.0

O6lvl4:
Cargo.toml 0.61.1 → 0.62.0 (the cycle carries language-surface additions: assert_snapshot, list-rest / as-patterns, the int bit and bytes families, the http *_response family — so a minor bump, and the release goes through an rc1 soak).

Also in the bump: the seven contracts stamped `since = "0.61.2"` during the cycle (C-329, C-331, C-334, C-335, C-337, C-338, C-339) take `0.62.0`, the contracts README re-renders, and llms.txt names the 0.62.0 train (docs-gen requires the package version to appear there).

Release path after this lands: wait for a clean fuzz-nightly on the fixed head, `check-interface-diff.sh v0.61.1 <sha>` (additive today), PR develop → main, tag `v0.62.0-rc1`, soak, final tag, seal.


O6lvl4:
Curated release-notes draft for v0.62.0, written 2026-09-07 from the 144 PRs merged since v0.61.1. When the tag exists: save this body to a file and run `gh release edit v0.62.0 --title "v0.62.0 — <headline>" --notes-file <file>`, then fill the final section from the seal.

```markdown
DRAFT — v0.62.0 is not tagged yet (develop carries Cargo.toml 0.62.0, 144 PRs since v0.61.1). Apply after the workflow creates the release:
  gh release edit v0.62.0 --title "v0.62.0 — <headline>" --notes-file <this file>
Fill the "Gates at the tag" section from the seal once it exists. Headline candidate:
  "v0.62.0 — patterns bind the rest, the docs run, and the wasm legs stop leaking"

A minor release: language-surface additions (list-rest and as-patterns, `assert_snapshot`, the int bit and bytes families, the http `*_response` family), the stdlib docs turned into executable doctests, the embedded-OOM class emptied on the structural leg, and the contract ledger gaining its three sides. Goes through an rc1 soak (#1484).

## Language

- **List-rest patterns**: `[h, ..t]` binds the tail on every leg (#1780).
- **As-patterns**: `name @ pattern` binds the whole value beside its destructuring (#1781); E060's usage walk descends through them (#1822).
- **`almide test` gains `assert_snapshot`** with an `--update-snapshots` accept step and a `--ci` mode where drift fails (#1876). Scratch artifacts are keyed per run and by absolute path, so same-named files and parallel runs cannot race (#1878). A bare `almide test` without an almide.toml refuses with the check hint (#1938).
- **ADR-0005 landed**: the pipe × `??` desugar order is fixed and pinned, the operator equivalences are contract fixtures, a no-op `?` warns (#1872). `option.unwrap_or` / `result.unwrap_or` open their deprecation window toward `??`, and `almide fix` rewrites them (#1902).
- **ADR-0013 accepted**: String becomes a shared immutable buffer with copy-on-write on the mut surface (#1692).
- A tuple-pattern lambda as a pipe stage destructures the piped value (#1879). List patterns bind irrefutable tuple sub-patterns on native (#1939). A guard's `else err(..)!` renders as the err return on native (#1936).
- `almide check --target wasm` reports the wasm build route at check time; the both-legs wall is E082 (#1983).

## Stdlib

- **http `*_response` family** beside every verb: status, headers and body (#1886). The http client core becomes `almide-rt-core`, spliced and linked from one source (#1775); the framed request family serves the embedded wasm lane (#1778) and the wasi:http@0.3 p3 client leg serves the string family through an async-lowered exchange (#1772, #1917, #1930).
- **zlib self-hosts**: one decoder on every leg, a stored-block encoder, exposed checksums (#1776).
- `int.is_even` / `int.is_odd` (#1932); `list.unique_by` admits any equatable key on both legs (#1811, #1818); `map.set` writes in place when the block is solely owned (#1819); `fs.glob` matches segment-wise on both legs (#1814); `path.extension` treats a leading dot as a hidden-file marker (#1825); `list.get_or` / `map.get_or` evaluate their default before the select (#1920).
- **The stdlib docs run**: 396 fences across string, list, math/float/json, map/set/int, datetime/random/regex/path/args, option/result/error/codec, http/compute, fs/process/env/io/testing promoted to executable or type-checked doctests (#1790–#1804); prose rows now state the accuracy and ordering the implementation has (#1823, #1824).
- Runtime types take reserved spellings in the emitted Rust, so a user `HttpRequest`, `Value` or `Endian` no longer collides (#1830); runtime helpers leave the flat namespace (#1964); stdlib signatures resolve canonically beside a same-named user type (#1837, #1959).

## Codegen and the wasm legs

- **The embedded-OOM class is emptied on the structural leg**: flat_map chunks and index writes stop retaining O(n²) generations (#1761), two-pass list_join (#1764), Map/Set lookups indexed by a side table so the 20k-entry build stops doubling per step (#1891), the region window for `consume(produce(scalars))` and nullary variant cases as pool statics (#1980).
- **Ownership on the wasm legs**: a closure call takes +1 on a borrowed heap argument, C-334 (#1788); a user fn's droppable result carries exactly one owned credit (#1987); a `return_call` releases the owned locals before replacing the frame (#1989); a temporary receiver of a Unit-returning bytes intrinsic is dropped after the call (#1851); the structural allocator's bump-head wrap takes the C-197 abort (#1929).
- **Composition family stage 2**: guard held across, match subject, interpolation, closure capture, and the nested Option[Result] payload (#1969); the wasm legs propagate a guard's err and a tail bang over a declared-Result callee (#1971); effect callbacks reach every fan mapper head (#1887, #1888); a value-returning effect fn with a mut param that can err lowers through a Result carrier, so the DDD package tree runs byte-identically (#1889).
- Artifact size: the RC core and the WASI host-op shims ship only when the emitted module reaches them (#1975); the p1 environ/args shim reclaims 1,033 B on every env-free artifact (#1885); host routing is decided from the emitted op set (#1974).
- Availability: the sweep covers the whole public surface — 973 fns probed, 167 walls declared (#1831); 46 pending-self-host rows close (#1817, #1890); env programs route to the structural leg by default (#1933).
- Regex engine backtracks through groups and alternations on both legs (#1951). Native main restores the default SIGPIPE disposition (#1954). Package builds are byte-identical across builds: the initializer walk takes declaration order (#1809).
- Four cross-target divergences found by the 60-minute fuzz dispatch are closed: unwrap_or strictness, matrix fill ceiling, from_bytes offset domain, copy_from negative length (#1913); matrix loaders take the element ceiling before rows × cols (#1945).

## Diagnostics and the negative corpus

- The negative-test corpus reaches 1,000 pairs; multi-file fixtures retire the E033/E420 exemptions and opaque aliases become constructible (#1774, #1784, #1803).
- E060 covers self-module imports, type-only uses and as-patterns — seven false warnings retire (#1785, #1822, #1858).
- The embedded host reports a named die as its one Error line (#1915); the wasm compile path prints checker warnings like the native driver (#1916).

## Evidence and gates

- **The contract ledger gains its three sides**: every promise can link its declaration heading, its refusal case and its Dojo challenge family, gated both ways (#2002, the contract-relative survival arc #1998).
- The judge's edits come back from almide/als: 16 fixtures, the reconciled ledger, ADR-0013..0017 and the advanced ALS chapters; the ALS pin gate refuses a contract id the judge has not seen (#1981, #1982).
- The interface diff sees effect fns, with a negative control (#1861); two release-path gates read true again (#1892); the size ratchet pins the shipped to_wasi form (#1894); the optimization pass roster per target is a table a gate keeps honest (#1863).
- `scripts/fuzz.sh` runs one local campaign of the generative differential fuzzer (#1900); the fuzz summary names the declared-available walls (#1901).
- codopsy A restored on almide-mir, almide-interp and almide-wasm-run (#1816, #1882, #1947).

## CI

- The build job caches dependencies only and ships a stripped test archive (−4 GB cache, −3 GB artifact) (#1870); `wasm_runtime_test` splits into per-part binaries so the shards spread its 1,552 s (#1848); shard weights refreshed (#1846, #1867, #1883); the clippy ratchet fails only on growth in a PR and enforces the shrink nightly (#1875); the stock-wasmtime installer pins v27.0.0 (#1847); fuzz-nightly re-runs reclaimed shards (#1940, #1946).
- Dependencies: wasm-tools family 0.258 (#1766), salsa 0.28 (#1768), sha2 0.11 (#1777).

## Gates at the tag

(fill from proofs/releases/v0.62.0.toml: blockers, interface diff v0.61.1 → v0.62.0, rc1 soak, fuzz run, assets, known problems)
```

PR #1899, opened 2026-09-04
The interp writes both Result block conventions and serves result.to_list at value level; condition-mode coverage gets its own floor

O6lvl4:
Fixes the 23 correctness findings of the 2026-09-04 fuzz-nightly ([run 33848741367](https://github.com/almide/almide/actions/runs/33848741367), tracked on #924) — one shape, `result.to_list(ok(v))` answering `[]` (or the raw slot) on the interp leg — and the condition-mode coverage job that went red on the first night it could measure.

**Cause (bisected to 3e8a01fed, #1897).** `prim.handle` of a `Result` stopped abstaining and materialized every Result in the cap-as-tag shape (`len@4 = 1`, tag in the slot's high half — `$rtf_result`, the fs floors' block). The backends write TWO shapes: `ok(<scalar>)` is the len-as-tag block (`Init::ResOkScalar`: `len@4 = 0` IS the Ok tag, the payload i64 fills the 8-byte slot), and `ok(<heap>)` / `err(e)` are cap-as-tag (`len = 1`, address in the low half, tag in the high). The registry body `result_to_list` (stdlib/result_core.almd) reads `len == 0` for Ok, saw `len = 1` on `ok(1)`, and answered the empty list; the fuzzer generates `result.to_list` freely, so the night lit up.

**Fix, two parts.**
- The interp writes both conventions (`heap_result_block`): scalar ok → `len 0` + the full i64 (Int in full, Bool 0/1, Unit 0, Float as bits); heap ok → `len 1` + address; err → `len 1` + address (or a low-half scalar) with the tag bit. Every fs/fan fixture that reads the cap-as-tag form still agrees.
- `result.to_list` becomes a value-level op like `option.to_list`: the registry body is the Int-payload twin, and the pool tier ran it for a String / Bool / Float payload too (12 of the 23). Two eval tests pin every payload class and the fold / is_empty / join shapes the findings used.

All 23 minimized repros replay clean through the fuzzer's ladder; the full 3-way oracle result is in the follow-up comment.

**Condition-mode coverage.** #1892 made the per-condition job find its test binaries; its first real measurement (56.55%) then failed the LINE-mode floor (58.19%) — the two modes count differently. Condition mode now ratchets against `proofs/coverage-baseline-condition.txt` (seeded at the measured 5655), same shrink-only rule.


O6lvl4:
Full 3-way oracle on this branch: 658 of 677 agree, 18 skipped (the pre-existing ledger, unchanged), 1 pre-existing backend split, **0 interp dissent**. A 4-minute local campaign (`xtarget-fuzz run --minutes 4 --jobs 4 --seed 20260904`): 904 programs, 878 clean, 3 walls, 0 findings.

Issue #1923, opened 2026-09-04
int.is_even / int.is_odd: the parity predicate pair is reflex vocabulary and E002 today — family ruling for the sized ints included


O6lvl4:
## Symptom

`int.is_even` / `int.is_odd` do not exist. Writing a parity guard, the natural first spelling is:

```almide
guard int.is_even(n) else err("odd: ${n}")
```

→ `error[E002]: undefined function 'int.is_even'` (the hint is fine and points at the CHEATSHEET). The working spelling is `n % 2 == 0`.

## Why file it

This is the exact reflex-vocabulary class the mission optimizes for: the predicate pair exists in the stdlibs a model has read (Gleam `int.is_even/is_odd`, Ruby, Python's `%` idiom aside), and the miss costs an edit-turn. The repo's own API-family rule applies: predicates come as the **pair**, on `int` — with the completeness statement for the sized ints (either the pair exists on every sized int module, or the omission is declared intentional — scalar `int` only — and the matrix gate asserts whichever is chosen).

## Ask

`int.is_even(n: Int) -> Bool` / `int.is_odd(n: Int) -> Bool` as self-hosted one-liners, the family-completeness ruling for the sized ints recorded in the matrix gate, spec tests, and the doc index row. Nothing else.

PR #1944, opened 2026-09-06
int.is_even / int.is_odd: the parity pair lands on scalar int, negatives included

O6lvl4:
Closes the core of #1923.

`int.is_even` was `E002` and the working spelling was `n % 2 == 0`. The
pair exists in the stdlibs a model has already read, so the miss costs an
edit turn for nothing.

## The change

Two self-hosted one-liners, placed with the other pure-Almide int fns:

```almide
// ── Parity ──
// `n % 2 != 0` (rather than `== 1`) keeps these correct for negative
// numbers: Rust's `%` takes the dividend's sign, so -3 % 2 == -1.
fn is_even(n: Int) -> Bool = n % 2 == 0

fn is_odd(n: Int) -> Bool = n % 2 != 0
```

`is_odd` is `!= 0` rather than `== 1` on purpose: `-3 % 2` is `-1`, so the
`== 1` spelling calls every negative odd number even. `spec/stdlib/int_parity_test.almd`
pins the answer for negatives rather than the spelling, and also pins that
the pair partitions — `is_even(n) and is_odd(n)` is never true, `or` is
never false.

Alongside: the doc index row (regenerated by `tools/gen-stdlib-doc-index.py`),
a naming-conventions line in the CHEATSHEET, and the stdlib count moving
969 → 971 in README / SPEC / WASM-OUTPUT.

## Not in this change — needs a ruling

#1923 also asks for the **family-completeness ruling for the sized ints**:
either the pair exists on every sized-int module, or the omission is
declared intentional (scalar `int` only) and the matrix gate asserts
whichever is chosen. That is a design call rather than a mechanical one,
so it is left open. The CHEATSHEET line currently reads "scalar `int` only
for now" and the matrix gate asserts nothing either way; both want
revisiting once the ruling is made.

## Gates

- `almide test spec/stdlib` — 160/160 files pass
- `almide docs-gen --check` — no drift
- `almide fmt --check stdlib/int.almd` — clean

## Provenance

The stdlib change and the negative-number reasoning were produced by an
agent (famulus5 driving `cf:glm-5.3-flash`) pointed at this worktree with
the spec test as its only success criterion; the spec test, the docs
follow-through and this description are mine. Reviewed before opening.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ The Perceus proof above proves one compiler pass, once. v1 generalizes that prin
 <!-- stats:generated:start — derived from docs/stdlib/*.md, spec/, and docs/contracts/contracts.toml by scripts/gen-readme-stats.sh; DO NOT EDIT between the markers -->
 | Derived count | Value |
 |---|---|
-| Stdlib | 969 functions across 43 modules — self-hosted `.almd`, signature indexes regenerated from the compiler by `tools/gen-stdlib-doc-index.py` |
+| Stdlib | 971 functions across 43 modules — self-hosted `.almd`, signature indexes regenerated from the compiler by `tools/gen-stdlib-doc-index.py` |
 | Tests | 425 `.almd` test files under `spec/` (`almide test spec/`) + the 326-contract cross-target ledger |
 <!-- stats:generated:end -->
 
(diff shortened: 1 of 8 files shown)


O6lvl4:
Closing: this duplicates work already on `develop`, and #1923 is closed.

I built it on a local checkout that was 476 commits behind `origin/develop`, and picked the issue from a stale listing. The landed implementation is byte-identical in its two function bodies —

```almide
fn is_even(n: Int) -> Bool = n % 2 == 0
fn is_odd(n: Int) -> Bool = n % 2 != 0
```

— and the landed comment carries the same `-3 % 2 == -1` reasoning plus the sized-int family ruling this PR deliberately left open. Nothing here is worth salvaging. Sorry for the noise.

Issue #1948, opened 2026-09-06
codegen: String bound by a match pattern, captured by a closure and used afterwards → rustc E0382 (borrow of moved value)


O6lvl4:
## Repro (almide 0.62.0, native build)

```almide
effect fn pick(names: List[String], want: Option[String]) -> Result[String, String] = match want {
  some(name) => {
    let near = list.filter(names, (s) => string.contains(string.to_lower(s), string.to_lower(name)))
    err("no `${name}` (near: ${list.join(near, ", ")})")
  },
  none => ok("all"),
}
effect fn main() -> Unit = match pick(["apply", "applied"], some("Appl")) { ok(s) => println(s), err(e) => println(e) }
```

```
$ almide build moved3.almd -o moved3
error[E0382]: borrow of moved value: `name`
```

The generated Rust moves `name` into the `move` closure (`almide_rt_string_to_lower(&*name)`) and then uses `name` again in the interpolated string. `almide check` accepts the program.

## Notes

- The same code with `name` as a **function parameter** instead of a match binding compiles fine (the parameter gets cloned into the closure). Only the match-pattern binding is moved.
- Workaround: bind a copy first — `let needle = string.to_lower(name)` — and capture `needle` in the closure.
- Hit while writing https://github.com/O6lvl4/peek (`select_windows` in src/main.almd).

Issue #1942, opened 2026-09-06
Incumbent wasm leg: a let-bound `??` over a tuple payload reads garbage (`let t = o ?? ("", 0); t.0` prints \0 / \254 bytes, exit 0)


O6lvl4:
Found while working #1904 (develop 4e9a79f5c, `ALMIDE_WASM_INCUMBENT=1` — the route fs/process programs and `almide test --target wasm` files still take).

```almide
fn main() -> Unit = {
  let o: (String, Int)? = some(("a", 1))
  let t = o ?? ("", 0)
  println(t.0)
  println(int.to_string(t.1))
  let n: (String, Int)? = none
  let v = n ?? ("dflt", 7)
  println(v.0)
}
```

| leg | stdout |
|---|---|
| native | `a` / `1` / `dflt` |
| structural (default `--target wasm`) | `a` / `1` / `dflt` |
| incumbent (`ALMIDE_WASM_INCUMBENT=1`) | `\0` / `1` / `\254` — exit 0, no wall |

The Int half of the tuple is right, the String half is a garbage handle on BOTH the some and the none arms. The eager `option.unwrap_or(o, ("", 0))` spelling in the same position prints `a`. #1904 documents the INLINE `(o ?? d).0` form as a wall on this leg; the let-bound form is worse — it is served and silently wrong, and no fixture carries it (the D4 fixtures kept the eager spelling on purpose, and the fuzzer runs the structural route).

Release-relevant: `??` is the spelling the D4 window recommends.

Issue #1926, opened 2026-09-06
guard … else err(…)! in an effect fn -> Unit renders `return (Err(..))?` (check green, rustc E0308)


O6lvl4:
## Repro (almide 0.62.0, native target)

```almide
effect fn show(path: String) -> Unit = {
  guard path != "" else err("usage: show <file>")!
  println(path)
}

effect fn main() -> Unit = show("x")
```

`almide check guard_bug.almd` → `No errors found`
`almide build -o guard_bug guard_bug.almd` → fails in rustc:

```
error[E0308]: `?` operator has incompatible types
6003 | ...tring) -> Result<(), String> {
     |              ------------------ expected `Result<(), String>` because of return type
6003 | ... { return (Err::<(), String>("usage: show <file>".to_string().to_string()))? }
     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Result<(), String>`, found `()`
     = note: `?` operator cannot convert from `()` to `Result<(), String>`
```

The guard's `else` branch is rendered as `return (Err(..))?` — the `!` on the else expression becomes a `?` applied to the value that is already the early-return, so the `return` receives `()` instead of the `Result`.

## Expected

Either of:
- `!` on a guard's else arm propagates the `err` as the early return (render `return Err(..)`), or
- E0xx at check time saying `!` is redundant / not allowed on a guard else arm, so the wall is not a rustc surprise.

Same construct without `!` (`guard path != "" else err("usage: show <file>")`) checks and builds fine, so the workaround is to drop the `!`.

## Context

Hit while writing an `effect fn cmd_outline(cfg, path: String) -> Unit` CLI handler in ctxgate; `almide test` failed on the whole file with the rustc error while `almide check` was green, which is the confusing part.

Issue #1925, opened 2026-09-05
Clone insertion misses a match-arm-bound String captured by a closure (check green, rustc E0382)


O6lvl4:
Found on released **v0.62.0** while running an LLM coding agent over
`research/benchmark/exercises` — `config-merger` failed to build on every
attempt, and the model could not act on the error because it points at a
generated file it has never seen.

## Shape

A `String` bound by a **match-arm pattern**, captured by a closure passed
to a stdlib higher-order fn, then reused in the same arm:

```almide
import list

fn check(o: String?, acc: List[List[String]]) -> Result[Int, String] = {
  match o {
    none => ok(0),
    some(key) => match list.find(acc, (p) => p[0] == key) {
      some(_) => err("dup: ${key}"),          // ← key reused after capture
      none => ok(0),
    },
  }
}

test "dup" { assert_eq(check(some("a"), [["a"]]), err("dup: a")) }
```

`almide check` — `No errors found`.
`almide test` / `almide run` — `codegen produced invalid Rust — this is an Almide bug.`

```
error[E0382]: borrow of moved value: `key`
   --> …/almide_test_main.rs:3903:61
3902 | match almide_rt_list_find(acc, (Rc::new(move |p: Vec<String>| almide_eq!(almide_index!(p, 0i64).clone(), key.clone(…
     |                                         --------------------- value moved into closure here   --- variable moved due to use in closure
3903 |   Some(_) => Err(format!("dup: {}", key).to_string()),
     |                                     ^^^ value borrowed here after move
```

The closure is emitted `move` and captures `key` by value; the per-use
clone insertion that covers this elsewhere does not fire for a
pattern-bound binding at a closure-capture site.

## Boundary

Same body, only the binding form changes:

| how `key` is bound | result |
|---|---|
| `let key = k` | builds, passes |
| fn parameter `key: String` | builds, passes |
| match arm `some(key) =>` | **E0382** |
| match arm `some((key, value)) =>` | **E0382** |

So it is the pattern binding, not the tuple and not the closure alone.

## Family

Same acceptance-parity family as #1033 (per-use clone decisions).
Nearest prior instance is #1130, which is the *synthesized loop
continuation* site; this is the closure-capture site, and #1130's repro
does not cover it. Also related: #1025, and #899's "passes check, fails
to build" umbrella.

## Why it stings more than the usual codegen gap

The diagnostic names a line in a temp file under
`/var/folders/.../almide-test/native/…/almide_test_main.rs`. For a human
[...]

PR #1966, opened 2026-09-06
xtarget-fuzz composition family: operator round trips over a known literal, judged by construction

O6lvl4:
Stage 1 of the operator-composition gate (ratified ○ on 2026-09-06).

The bug class this hunts is one cell at a time in the hand-written spec: `??` over a tuple payload (#1904, #1942), `guard … else err(…)!` in an effect fn (#1926), a tuple-pattern lambda as a pipe stage (#1873), a match-bound String captured by a closure (#1925, #1948). Each is an operator × payload class × position cell; a hand-maintained matrix fills only the cells someone already stepped on. This family draws the product, with the answer known by construction like the identity family (#1332).

- `tools/xtarget-fuzz/src/generator/composition.rs`: every probe wraps the literal K (Int, String of length K, tuple-with-heap, record) in 1–3 operator round trips — `some(x) ?? d`, `ok(x) ?? d`, `ok(x)!`, `ok(x)? ?? d`, `x |> id`, `match some(x) { … }`, `(x, 7).0` — in tail, `let`-bound, or call-argument position of a plain or effect fn, and prints K. Fallbacks carry K + 1, so a wrong branch, field, or layout changes the number. `// @expect` lines carry the oracle in-source, so `ladder repro.almd` and the minimizer work unchanged; shrinking is plan-level (drop a probe, peel an operator, move to tail).
- `generator/mod.rs`: `Family::Composition`, `FamilyPlan` (identity | composition) as the plan carrier, a separate RNG sub-stream for the family roll so archived identity and synthesis seeds keep replaying; `--family composition` and part of `all` (3 : 10 of the non-identity share). `minimize.rs` dispatches on the plan; `findings.rs` meta names the family.
- Rendered output is `almide fmt`-stable (`T?`, one blank line between items, fmt's match layout), so the committed sample passes the fmt gate untouched: `spec/wasm_cross/composition_fuzz_batch.almd` (seed 7, index 39: four probes, all four payload classes, five operator kinds), contract C-343.
- Unit tests: 300 generated programs parse and round-trip their `@expect` header; every shrink candidate is still a composition program; 120 generated programs pass `almide check` (the family promises well-typed-by-construction — a reject is a generator defect, never a finding).

Local campaign, `--family composition --count 200`: 200 generated, 0 generator rejects, 0 findings, 190 clean, 10 incumbent-subset walls (heap-result `??` / match — the #1904 family, reported as subset debt, not divergence). 150 further programs: 0 check rejects; 60: 0 fmt drift.

Stage 2 (next PR): `guard … else err(…)!`, the match subject, string interpolation, closure bodies, nested `Option[Result[…]]` payloads.

https://claude.ai/code/session_01NvweJKc7fNXQn4BzB7GQQM


O6lvl4:
Its two commits landed on develop as 4ab09a523 / 140da974b when the stacked #1969 merged by rebase (the branch is now behind develop with nothing new in its tree), so this PR is empty; closing.

Issue #1990, opened 2026-09-07
Module-space fns skip the tail-site param release, so every bytes.append_* call leaks the old buffer: a 2,000-iteration append loop holds 5.4 MB (3.84x for 2x N)


O6lvl4:
Surfaced by the #1696 phase-B1 witness: 130 registry fns certify as `iam|im` — the callee-owned param is shared into the tail callee and never released. `populate_tail_release_set` (func.rs) leaves the set EMPTY for any fn with `cur_module.is_some()` ("Pool/registry bodies keep the pre-existing accounting"), so a module-space fn whose body is a tail call — `fn bytes_append_u16_le(b: Bytes, v: Int) -> Bytes = __bam(b, v, 2, true)` and all its `append_{u16,i16,u32,i32,i64,f32,f64}_{le,be}` siblings — takes the `return_call` route with no param release, and the fall-through epilogue that would have released `b` is dead code after the jump. Each call therefore leaks the OLD buffer, and a growing buffer makes it quadratic.

Measured on the structural leg (harness `heap_end`), develop + #1987 + #1989:

| program | N=1000 | N=2000 | ratio |
|---|---:|---:|---:|
| `var buf = bytes.new(0); for i in 0..<N { bytes.append_u16_le(buf, i) }` | 1,427,024 | 5,486,160 | **3.84×** (quadratic; linear would be ~2×) |

The hazard the exclusion guards against is the raw-pointer one (a `prim.handle(param)` handed to the tail callee), and `body_uses_prim` already covers it. Fix: populate the tail-release set for module-space fns too (lambdas and prim-using bodies stay excluded), so a registry tail call releases its owned params before the jump exactly as an entry fn does; pin the append loop's high-water mark as linear in `tests/call_result_credit.rs`; regenerate the alloc ledger (the bytes writer rows drop).

O6lvl4:
Status after a day of measurement (all on the structural leg, harness `heap_end`):

**Landed / in flight**
- #1987 (merged): a user fn's droppable result is owned at bind/assign/return — the 64 B-per-call leak class.
- #1989 (CI rerunning): the tail-site release covers the owned locals when the frame is truly replaced. Two corrections found on the way: a SELF `return_call` is loop-converted (tco.rs), so its frame lives on and only the params may be released there (releasing the locals double-freed lisp's `parse_list` `t`); and the release stays off for module-space / prim-using bodies (raw views).
[...]

PR #1992, opened 2026-09-07
A registry-table module call's droppable result is owned, and the linked bytes append arm releases the receiver's old block before the write-back

O6lvl4:
Part of #1990 (does not close it — the quadratic `bytes.append_*` loop stays open on the prim-tier convention, see the issue's status comment). Stacked on #1989.

- **Registry-table results are owned** (`rc_owned_result` for `CallTarget::Module`): every `lower_module_call` entry takes a sequence number kept on a stack while its dispatch and arguments lower; the registry table path (`lower_linked_call`, the only module route with the callee-owned epilogue convention) stamps the enclosing entry's number. A bind / assign / return route that captured the counter before lowering its rhs asks whether the rhs call itself took the table path (`stamp == seq0 + 1`). Special forms keep the conservative +1 (their own conventions), nested table calls never match — the predicate mirrors the actual dispatch instead of guessing from names.
- **The linked `bytes.append_*` / `write_*` arm releases the receiver's OLD block** before storing the twin's fresh result (the push intrinsic frees the outgrown block inside `$bytes_push`; the linked family never did).
- Two further mechanisms were tried here and reverted with their reasons recorded in the code and on #1990: module-space fns in the tail-release set (`regex_fuzz_batch` printed freelist bytes and trapped — transitive raw views) and `prim.alloc_*` owned at bind (`regex_engine`'s captures printed garbage — the prim bodies use the bind's +1 as the keep-alive for raw escapes).

Verified: `almide test spec/` 436/436, wasm crate suite, cross-target + interp-ledger + examples-parity tests, and `regex_engine` / `regex_fuzz_batch` / `examples/lisp.almd` byte-identical native vs wasm.


O6lvl4:
Superseded by #2012 (the same change rebased to three commits; the queue refused this branch's already-landed commits).

Issue #1988, opened 2026-09-06
A return_call frame replacement releases the owned params but not the owned locals: { let x = mk(n); take(x) } leaks x on every call


O6lvl4:
Surfaced by the #1696 phase-B1 witness (the call-boundary recorder showed the local's stream ending `iam` — a credit received, shared, moved, never released). func.rs's epilogue comment already names it: "Early returns and tail calls skip this — a leak, never a dangle." Measured on the structural leg (harness `heap_end`, N = 1000 vs 8000 iterations), develop + #1987:

| shape | N=1000 | N=8000 | per call |
|---|---:|---:|---:|
| `fn f(n) = { let x = mk(n); take(x) }` — the tail call takes the `return_call` route | 97,760 | 321,760 | **32 B** |
| `fn f(n) = { let x = mk(n); let r = take(x); r }` — same program, no frame replacement | 65,792 | 65,792 | flat |

`emit_tail_param_release` (calls.rs) runs before a `return_call` and decs `rc_droppable_params` only; the `rc_owned` locals the epilogue would have released are skipped because the epilogue never runs after the jump. Fix: release the owned locals there too, in the epilogue's order and with the epilogue's #1770 rule (a param the Assign routes made an `rc_owned` member is released once). Safe by the same argument as the epilogue: the tail call's arguments were already lowered and `rc_arg_guard`-inc'd, `rc_owned` holds only flat blocks (no heap interiors), and a local is never the tail call's result. Early returns (`Op::Return` inside arms) are the remaining skip and are not in this fix.

PR #1993, opened 2026-09-07
A bytes.append_* loop grows linearly: prim.alloc_* results are owned and a module-space wrapper releases its params before a return_call to a returns-fresh callee

O6lvl4:
Closes #1990.

Stacked on #1992 (base `fix-module-tail-release`); retarget to `develop` after it lands.

**Root cause, two halves.** (1) A `prim.alloc_*` result was not owned at its bind, so every linked buffer came back at rc 2 and the caller's single release left it at rc 1. (2) The module-space wrapper `bytes_append_u16_le(b, v) = __bam(b, v, 2, true)` never released `b` at its `return_call` site (module-space bodies keep every release on the epilogue — the raw-address rule of #1988 — and a tail call has no epilogue), so the OLD buffer stayed alive on every append: quadratic over a loop (N=2000 peaked at 3.84× N=1000).

**Fix.**
- `rc_owned_result`: `prim.alloc_*` is owned. The earlier revert blamed this rule for the regex capture garbage; the bisect with an allow-list shows the culprit was the unconditional module-space tail release, not the alloc rule — with the tail release gated as below, `regex_engine`, `regex_fuzz_batch` and `examples/lisp.almd` are byte-identical native vs wasm.
- `fresh.rs` (new): RETURNS-FRESH fns, a least fixpoint over the table — a fn joins when its tail is a literal / constructor / `prim.alloc_*` / another returns-fresh call, through `Var`s bound to such and never re-assigned, and through every `if` / `match` arm. `__bam` and every `bytes_append_*` / `bytes_write_*` wrapper are in the set (`ALMIDE_FRESH_DEBUG=1` prints it).
- `Emitter::tail_release_fresh_only`: a module-space, prim-free body releases its droppable PARAMS at a `return_call` site only when the callee is returns-fresh — its result cannot be a view of a param, so the release is exactly the epilogue's, moved before the jump (the callee still holds rc_arg_guard's +1 and decs at its epilogue). Owned locals stay on the raw-address rule; indirect and self tail sites are unchanged.

**Evidence.**
- `tests/call_result_credit.rs` gains `a_registry_tail_call_releases_its_owned_param`: the append loop's high-water at N=2000 must be < 2.5× N=1000 (was 3.84×).
- Alloc ledger: 62 watermarks regenerated, every one DOWN (e.g. `bytes_temp_receiver` 10,311,184 → 7,750,848); none up. Witness floor 3 → 9 RC-bearing certs.
- Full sweep green: `almide test spec/` 436/436, the whole `almide-wasm` crate, cross-target + interp ledger + examples parity.


O6lvl4:
Superseded by #2018 (rebased onto develop after #2012 landed).

Issue #1986, opened 2026-09-06
The structural leg leaks every droppable heap value a user fn returns: the bind and the return tail add +1 to the credit the call already transferred


O6lvl4:
Found while extending the #1696 phase-A witness to call boundaries: the per-frame accounting says a user-fn call result arrives with exactly one owned credit (the callee's ret-inc, or the fresh alloc of a certainly-fresh tail), yet the caller's Bind route treats a `Call` rhs as *borrowed* and adds `rc_inc_top` — so the block ends its life at rc 1 and is never freed. Measured with the harness `heap_end` (max of `__heap`/`__heap_high`) at N = 1000 vs 8000 loop iterations, structural leg, develop ddd019866 + #1980:

| shape (inside `for i in 0..<N`) | heap N=1000 | heap N=8000 | per iteration |
|---|---:|---:|---:|
| `let x = [i, i, i]` (control) | 65,888 | 65,888 | flat |
| `let x = mk(i)` — `mk` binds a fresh list and returns the var | 129,760 | 577,760 | **64 B** |
| `let x = mk2(i)` — `mk2 = [n, n, n]` (certainly-fresh tail) | 129,760 | 577,760 | **64 B** |
| `let x = via(i)` — `via(n) = mk(n)` (call in tail position) | 129,760 | 577,760 | **64 B** |
| `total = total + list.len(mk(i))` (call result as a temporary, never bound) | 129,760 | 577,760 | **64 B** |
| `let y = string.trim(" abc ")` (module helper result) | 209,808 | 1,217,808 | **144 B** |
| `let y = tag("abc")` (Str from a pool static) | 65,776 | 65,776 | flat (statics are below the heap floor) |

Sites that decide fresh-vs-borrowed with `rc_certainly_fresh` alone and over-count a call result: `stmts.rs` Bind (+1 at bind) and Assign, `func.rs` return tail (a tail `Call` takes a ret-inc it does not need — the `via` row), `list.rs`'s push-window chunk. `rc_arg_guard` is already right (no +1 on a call arg — the credit moves into the callee) and `rc_share_guard` stores a call result without +1.

[...]

PR #1999, opened 2026-09-07
The structural witness covers the call boundary, and the 63 module-space tail-site leaks it exposed are fixed (#1696 step 4, phase B1)

O6lvl4:
Refs #1696 (step 4, phase B1 — the call boundary in the structural witness, still certificate v0).

Stacked on the module-space tail-release branch (`fix-module-tail-release-all`, itself on #1993 → #1992); retarget to `develop` as the stack lands.

**What the witness now records** (`crates/almide-wasm/src/witness.rs`): a droppable Var argument shares (`a`, the site's real `rc_inc`) and its credit moves into the callee (`m`); a fresh temporary is born and moves (`im`); a user fn's droppable result is a received credit (`i`, #1986); an owned tail moves out (`im`); a `return_call` releases the owned params and locals before the jump (`d`), and the frame-replaced flag is set only for a TRUE `return_call` — a self tail call is loop-converted (tco.rs) and leaves the straight-line subset (the recorder is not loop-aware; it produced an `iamdd` over-release before the gate learned the shape). `witness_bind` takes the module-call sequence stamp so its ownership verdict mirrors the Bind route exactly.

**What it found.** With the honest frame flag, 63 module-space wrappers certified `iam` — no release — because module space sat outside the tail-release set (`fan_map`, `fan_any`, `http_set_header`, `http_get_header`, `__gby_add`, …). The base branch of this PR fixes that: params and owned locals are released at every true `return_call`; only prim-using bodies and lifted lambdas keep the raw-address rule. Every certificate now balances: 216 RC-bearing certs, floor raised 9 → 220. (No flat heap probe can pin those 63 — `fan.map` is lowered natively by fan.rs and Map/Result blocks are never freed on the structural leg yet — so the floor test IS the executable evidence, which is why the fix and the witness land as one stack.)

**Evidence.**
- `tests/witness_call_boundary.rs`: exact certificates for `pass`/`both`/`owned_tail`/`fresh_tail`/`bind_then_pass`/`temp_arg`/`take`, each balanced; `self_tail` must be declined.
- `tests/witness_floor.rs`: every witnessed cert balances, none poisoned, floor 220 RC-bearing certs.
- `spec/wasm_cross/witness_straightline.almd` gains `take` / `pass` / `bind_then_pass`; `proofs/gate.sh` phase-B1 rows run them through the extracted kernel-proven checker (`id`, `iamd`, `iamd` ACCEPT) plus the tamper drill `iamd → iam` (unreleased tail-site param) → REJECT, verified locally with `proofs/checker`.
- Full sweep green: `almide test spec/` 436/436, the whole `almide-wasm` crate with ledgers regenerated, cross-target + interp ledger + examples parity.


O6lvl4:
Superseded by #2023 (rebased onto develop after #2021 merged).

PR #2000, opened 2026-09-07
Module-space bodies release their params and owned locals at every true return_call, retiring the returns-fresh gate

O6lvl4:
Stacked on #1993 (base `fix-1990-append-linear`); retarget to `develop` as the stack lands. Its executable evidence is the call-boundary witness in #1999, which stacks on this branch.

**Why.** The module-space exclusion from the tail-release set (#1988) was a misattribution: the two traps blamed on releasing at module-space tail sites (the regex capture garbage, lisp's `parse_list`) were the loop-form double free that #1989 fixed. With an honest frame-replaced flag, the structural witness certified 63 module-space wrappers `iam` — a `return_call` that never released its params or owned locals (`fan_map`, `fan_any`, `http_set_header`, `http_get_header`, `__gby_add`, …).

**Fix.** `populate_tail_release_set` admits module-space bodies; only prim-using bodies (raw views) and lifted lambdas keep every release on the epilogue. The returns-fresh fixpoint (`fresh.rs`, #1993) and `tail_release_fresh_only` are retired — one rule: params and owned locals are released at every TRUE `return_call` site, params only at a loop-converted self call.

**Evidence.** `regex_engine`, `regex_fuzz_batch`, `examples/lisp.almd`, `fan_deterministic`, `http_response_headers`, `map_insertion_order` byte-identical native vs wasm; alloc ledger 58 watermarks regenerated, every one DOWN, none up; full sweep green (`almide test spec/` 436/436, the whole `almide-wasm` crate, cross-target + interp ledger + examples parity). The witness floor in #1999 goes from 63 unbalanced certificates to 0 (216 RC-bearing certs, all balanced).


O6lvl4:
Superseded by #2021 (rebased onto develop after #2018 merged).

Issue #1995, opened 2026-09-07
A checked ExitPlan owns every function exit: per-edge credit conservation makes the #1988/#1990 leak class unrepresentable


O6lvl4:
## What

Extract the ownership decisions the structural emitter makes at every function exit into one shared, checked `ExitPlan`, produced before target rendering:

```text
ExitPlan {
  outgoing_values,
  transferred_credits,
  released_credits,
  carried_borrow_dependencies,
  closed_regions,
  continuation   // ReturnSuccess | ReturnError | TailTransfer
}
```

For every exit edge the checker verifies credit conservation with unique credit identities (not count matching):

```text
incoming frame credits  ==  transferred ⊎ released
```

Tail transfer order is fixed: evaluate args and establish outgoing ownership → transfer outgoing credits → release every remaining frame credit, locals and params alike → transfer control. Module membership cannot alter accounting; intrinsics carry explicit borrow/retain summaries, `prim.*` gets no blanket exemption.

## Landing

Shadow mode first: derive the plan from the current emitter's decisions and diff it against what the emitter actually does, on the full `spec/` corpus and the wasm_cross gate. Every mismatch is either a known leak or a new one. Then flip the emitters to consume the plan.

## Why

#1988 (return_call releases owned params but not owned locals) and #1990 (module-space fns skip the tail-site param release) are the same defect class: the release set is computed per emitter path, so each new path can forget a subset. A per-exit conservation check makes an incomplete exit unrepresentable instead of a bug to find.

Bug-class work, so it lands before the `scoped` surface. Ratified 2026-09-07 as step 2 of the contract-relative survival arc.

O6lvl4:
Field report before the design, from today's reading of every frame exit the structural emitter writes (grep of `.return_()` / `.return_call(` in the frame-emitting files, runtime helpers excluded):

| exit edge | site | release before this work |
|---|---|---|
| ReturnSuccess (fall-through) | func.rs epilogue | owned locals + droppable params |
| TailTransfer (`return_call`, named / indirect) | calls.rs `emit_tail_param_release` | params only (#1988) → + locals (#1989) → module space excluded (#1988) → included (#2000) |
| TailTransfer, self (loop-converted) | same | params only, correctly |
| ReturnError (`f()!` err, `!` none in effect / Option fn, raised `err(..)`) | data.rs ×4 | **none** — #2001, fixed on branch `fix-err-exit-release` |
[...]

PR #2003, opened 2026-09-07
The error-propagation and guard exits release the frame like the epilogue

### LINKED ISSUE #2001: Error-propagation exits skip the frame release: f()! and a raised err(..) return over the owned locals and the droppable params
On the structural leg, the ERROR exits of a function — `f()!` propagating its err, a raised `err(..)` in an effect fn, `!` on none in an Option fn — emit a bare `return` (data.rs `lower_try_unwrap` / `lower_err_raise`) without the release the fall-through epilogue performs: every rc_owned local and every droppable param stays at its count. The #1995 exit class (same defect as #1988 / #1990: the release set is computed per emitter path, and this path had none).

Measured with the credit harness (`crates/almide-wasm/tests/call_result_credit.rs`, `an_error_exit_releases_the_frame_like_the_ok_exit`): an effect fn `work(xs: List[Int], n)` with `let y = [n, n, n]` and `boom(n)!` grows 224 B per call on the err path against 96 B on the ok path — 128 B = the four leaked blocks (xs, y in `work`; xs, y in `raise`).

Fix: `Emitter::emit_error_exit_release` mirrors the epilogue (owned locals, then droppable params not among them) at the four error-exit sites; the err block already shares its payload (`rc_share_guard` in `lower_sum`), so the release is safe by the epilogue's own argument.

O6lvl4:
Closes #2001. Refs #1995 (the exit class; field report and route proposal there).

Stacked on #1999 (base `feat-1696-witness-b1-clean`, itself on #2000); retarget to `develop` as the stack lands.

**Bug.** The ERROR exits — `f()!` propagating its err, `!` on none (effect fn → `err("none")`, Option fn → none), a raised `err(..)`, and the `guard c else v` early returns (value and Unit forms) — emitted a bare `return` with none of the epilogue's releases: every rc_owned local and every droppable param stayed at its count. Measured 224 B per call on the err path against 96 B on the ok path (the four leaked blocks: xs, y in `work`; xs, y in `raise`).

**Fix.** `Emitter::emit_error_exit_release` mirrors the epilogue (owned locals, then droppable params not among them; `rc_frame_params` now carries the frame's droppable params regardless of the raw-address rule) at the four data.rs sites and the two guard sites. The guard's value form applies the tail's ret-inc rule first (a borrowed value takes +1 before its owner is released). Safe by the epilogue's own argument: the returned err block already shares any borrowed payload (`rc_share_guard` in `lower_sum`). The witness is not branch-aware, so an armed recorder is poisoned rather than fed a one-path stream.

**Evidence.** `an_error_exit_releases_the_frame_like_the_ok_exit` in `tests/call_result_credit.rs`: `work` (`!`), `raise` (`err(..)`), `guarded` (`guard`) — the err path's per-call growth must not exceed the ok path's (was 224 vs 96; now equal). Full sweep green: `almide test spec/` 436/436, the whole `almide-wasm` crate with ledgers regenerated (one alloc watermark down, none up), cross-target + interp ledger + examples parity.


O6lvl4:

Issue #2005, opened 2026-09-07
Linked stdlib impls leak inside their own bodies: string.trim 96 B and list.repeat 64 B per call on the structural leg


O6lvl4:
Structural leg, measured with the credit harness (`crates/almide-wasm/tests/call_result_credit.rs`, N=1000 vs 8000, per-call growth of the high-water mark, every value BOUND so the #2004 argument-temporary class is out of the picture):

| body | growth |
|---|---|
| `let s = " x" + int.to_string(i); let t = string.to_upper(s)` | 16 B/call (only the #2004 `to_string` temp) |
| same with `string.to_lower` | 16 B/call |
| same with `string.trim` | **112 B/call** — 96 B inside `string_trim` |
| `let s = list.repeat(i, 3)` | **64 B/call** inside `list_repeat` |

A user wrapper around each (`fn wrap(s) = string.trim(s)`) adds nothing — the wrapper's frame is clean (#1995 ExitPlan); the leak is inside the linked impl's own body: prim-tier bodies keep every release on the epilogue under the raw-address rule, and something in these two bodies (an intermediate buffer, a `prim.alloc_*` result not owned by a bind, a scratch list) never reaches it. Same audit as #1990's follow-up: disassemble `string_trim` / `list_repeat` (`wasm-tools print`), count credits per block, find the one without a dec. The other VERIFIED impls measured clean (`to_upper`, `to_lower`, `int_to_string` apart from #2004).

O6lvl4:
Measured against the generated ownership matrix (`tests/native_result_ownership.rs`, 107 rows once generic signatures are instantiated at `Int`): 30 growing rows → 8. Every row was probed to a named shape (dev probe: `ALMIDE_PROBE_SRC=<file> cargo test -p almide-wasm --test heap_probe -- --nocapture`, prints the heap high-water mark):

| shape | rows | fix (PR #2007, second batch) |
|---|---|---|
| prim-tier body ending in a tail call: the raw-address rule kept every release on the dead epilogue | `float.to_string` ×3 (4 KB/call), `float.to_fixed` | `tail_transfer_ok`: a raw-rule frame that holds credits keeps the call in non-tail form so the epilogue runs; credit row `a_prim_body_with_owned_locals_releases_them_after_its_tail_call` |
[...]

Issue #2004, opened 2026-09-07
A fresh call-result temporary handed to a non-consuming native op is never released (string.len(int.to_string(i)) leaks 16 B per call)


O6lvl4:
Structural leg. A droppable value produced by a call and consumed directly as an ARGUMENT of a native (non-table) op is never released: nobody owns the temporary. Measured with the credit harness (`crates/almide-wasm/tests/call_result_credit.rs`, N=1000 vs 8000, per-call growth of the high-water mark):

| body | growth |
|---|---|
| `string.len(int.to_string(i))` | 16 B/call |
| `string.len(" x" + int.to_string(i) + " ")` | 80 B/call |
| `string.len(string.trim(" x" + int.to_string(i) + " "))` | 208 B/call |

Binding the same values first (`let s = …; let t = wrap(s); string.len(t)`) is flat: the Bind route owns and releases them. The frame-exit class (#1995 / #2001) is separate — this is the argument-temporary class: the credit a fresh result carries is handed to a callee that does not take ownership (a native op, or a table fn under the callee-owned convention? — the table path is fine: `rc_arg_guard` skips certainly-fresh args and the callee's epilogue consumes the credit; the natives have no epilogue).

Fix shape: at a native-op argument position, a `rc_owned_result` argument is held in a scratch local and released after the op (the temporary's one credit is spent there), mirroring what the Bind route does for the bound case. Belongs to the #1995 arc (`outgoing_values` / `transferred_credits` at call sites).

O6lvl4:
The mechanism was rebuilt after the review: no hand list of fresh results or reader ops. Every arm returns `Lowered { ty, own: Owned | View | Scalar }` and lowers each argument through `lower_arg(.., ArgMode::Borrow | Retain)` (arm.rs); `scripts/check-arm-args.sh` refuses an undeclared argument, and `tests/native_result_ownership.rs` is the generated matrix over every droppable-result stdlib fn (ledger `native-own-known-leaks.txt`, shrink-only, 30 rows left — all internal leaks of arm bodies, the #2005 class). Borrowed temporaries get their own local pool: sharing the scratch pool let `fs.rename` overwrite the temporary's slot and release its own Result block (found by the corpus burn-up on the first sweep). PR #2007.

Issue #1996, opened 2026-09-07
E-OWN-LOWERING: the build validates emitted exit operations against the checked ExitPlan and names a mismatch as a compiler defect


O6lvl4:
## What

Validate that the emitted release / close / transfer operations of each function exit implement its checked `ExitPlan`. A mismatch is a build failure with its own diagnostic, distinct from any source error:

```text
error[E-OWN-LOWERING]: tail exit leaves an ownership credit outstanding
  --> helper.almd:7:3
   = value: local `x`
   = expected: transfer or release before tail transfer
   = emitted: neither
   = compiler contract failure; no artifact emitted
```

Scope: narrow validation of the generated Rust subset and the wasm control-flow ops for functions that carry an ExitPlan. Downstream rustc / wasmtime assumptions stay in `docs/contracts/proven-vs-trusted.md` as trusted.

## Why

A checked plan does not help if an emitter can omit its cleanup: that reproduces the trust gap that produced #1988/#1990. And the diagnostic must never instruct an LLM to rewrite a valid program into a shape the emitter happens to handle. A valid source that exposes a compiler ownership bug is a compiler defect, and the text says so.

Depends on the ExitPlan issue. Ratified 2026-09-07 as step 3 of the contract-relative survival arc.

PR #2006, opened 2026-09-07
A checked ExitPlan owns every function exit: one producer, one renderer, and a source gate that refuses an unplanned return

O6lvl4:
Closes #1995 (route A + B as proposed on the issue; C — branch-aware credit identities — is #1996's).

Stacked on #2003 (base `fix-err-exit-release`); the full stack is #1992 → #1993 → #2000 → #1999 → #2003 → this. Retarget to `develop` as the stack lands.

**A — one producer, one renderer.** `crates/almide-wasm/src/exit_plan.rs`: `Continuation { ReturnSuccess, ReturnError, GuardReturn, TailTransfer { replaces_frame } }`; `exit_plan(c)` derives from the frame state at the site the partition of the frame's credits (rc_owned locals ⊎ droppable params) into `released ⊎ carried`; `emit_exit(&plan)` is the ONLY writer of a frame's release instructions and carries the witness attribution (success/tail record their decs and the frame replacement; error/guard poison the straight-line recorder). The epilogue, `emit_tail_param_release` and `emit_error_exit_release` are gone; `rc_droppable_params` folds into `rc_frame_params`. The carried set is non-empty on exactly two edges: the loop-converted self tail call (locals live on) and a raw-address-rule frame's tail site. Module membership does not enter the plan; `prim.*` bodies get only the epilogue-vs-tail placement rule, never a blanket exemption from release.

**B — the checked half, shadow mode.** `scripts/check-exit-sites.sh` (CI `checks` job, `proofs/gate-verification.toml` NEGATIVE_TESTED): every `return` / `return_call` / `return_call_indirect` in a frame-emitting file must sit within four lines after `emit_exit(`; hand-written helper bodies (allocator, string primitives, display/eq/scan, Value equality, decoders, map probes) and the deterministic-budget unwind are an explicit allow-list. The credit-identity mirror for straight-line fns is the B1 witness in #1999 through the kernel-proven checker.

**What the gate found on its first run.** A THIRD tail site — `lower_linked_call`'s `return_call` (the registry-table path) — released nothing: a user fn whose tail is `string.to_upper(s)` replaced its frame with `s` at rc 1 on every call. Fixed through the plan; pinned by `a_linked_tail_call_releases_the_frame`.

**Two more classes surfaced while probing, filed, not in scope here:** #2004 (a fresh call-result temporary handed to a native op is never released, 16 B/call for `string.len(int.to_string(i))`) and #2005 (linked impls leaking inside their own bodies: `string_trim` 96 B, `list_repeat` 64 B per call).

**Evidence.** Byte-identical alloc ledger (no fixture reaches the linked tail site; sizes regenerated for the added releases); `tests/call_result_credit.rs` 9 rows green; witness floor and call-boundary certs unchanged; full sweep green (`almide test spec/` 436/436, whole `almide-wasm` crate, cross-target + interp ledger + examples parity); the gate's negative test recorded in `proofs/gate-verification.toml`; codopsy A (91) held.


O6lvl4:

PR #2007, opened 2026-09-07
Every native arm declares the ownership of its result and of each argument as a type, with a source gate and a generated ownership matrix in place of the hand lists

O6lvl4:
Closes #2004. Closes #2005.

Stacked on #2006 (base `feat-1995-exit-plan`); the stack is #1992 → #1993 → #2000 → #1999 → #2003 → #2006 → this. Retarget to `develop` as the stack lands.

**The ownership of every value that crosses a native arm boundary is DECLARED at the site, as a type — never inferred from a hand list.** The reference shape is Swift OSSA's `@owned` / `@guaranteed` on the signature and Koka / Lean's owned-result convention with declared borrows; here the callee is a native arm of the structural emitter, and its signature is what it returns and what it says about each argument (`crates/almide-wasm/src/arm.rs`).

1. **Results.** Every arm returns `Lowered { ty, own: Owned | View | Scalar }` (`ArmResult`); `Scalar` asserts non-droppable in debug builds. The module-call wrapper is the one reader: it marks the call node owned (pointer identity, `owned_call_marks`) exactly when the arm said `Owned` and the type is droppable, so the bind / assign / return / argument routes take no borrow +1. The registry-table path hands over the callee's credit as `Owned`. `NATIVE_FRESH_RESULTS` and the completion-order stamp are gone.
2. **Arguments.** Every argument an arm lowers goes through `lower_arg(expr, want, ArgMode::Borrow | Retain | Raw)` — 341 sites across 23 arm files. Under `Borrow` an owned temporary (`string.len(int.to_string(i))`) is parked and released by the enclosing `arm_scope` once the op's result is on the stack; under `Retain` its credit moves into the arm's result and a block that already has a holder takes the share +1 there (the declaration carries the guard — `value.str` stored a Var's string bare and only an over-borrow elsewhere kept it alive); `Raw` is the prim floor alone (a raw address read out of the block, the frame's raw-address rule owns the release) and the gate refuses it anywhere else. Shared probes take the mode from their caller (`set_scan(s, x, x_mode)`, `map_scan(m, key, key_mode)`: `insert` / `set` retain, `contains` / `get` / `remove` borrow — both found as cross-target divergences on the first run). `println` / `eprintln` and the slice / map-access / map-set syntax entries are their own scope. A borrowed temporary lives in its OWN local pool (`BORROW_POOL`), disjoint from the scratch holds an arm releases and re-acquires mid-op — sharing the scratch pool let `fs.rename` land a hold on the temporary's slot and decrement its own Result block (fs_write_errno, caught by the corpus burn-up and the stock-wasmtime gate on the first sweep). The `reader_op` list is gone; `arg_temps.rs` keeps only the language-construct half (BinOp / ForIn / Match / IndexAccess / StringInterp operands).
3. **Table path** (`rc_arg_guard`, unchanged from the first cut): an owned argument moves into the callee with no +1; the witness mirrors it.

**Gates.**
- `scripts/check-arm-args.sh` (CI, `proofs/gate-verification.toml` NEGATIVE_TESTED): an arm file may lower an argument only through `lower_arg`; the undeclared form is a failed build. A borrow with no scope is the existing hold-balance BUG wall at the frame's end, never a silent leak.
- `tests/native_result_ownership.rs`: the GENERATED matrix — every stdlib fn with a droppable result from the `docs/stdlib` signature indexes, flat-across-N at 1000 vs 8000 — with the shrink-only ledger `tests/golden/native-own-known-leaks.txt` (30 rows left, all INTERNAL leaks of arm bodies, the #2005 class; `ALMIDE_UPDATE_NATIVE_OWN=1` regenerates). A `Retain` on a read block shows as a growing row; an `Owned` on a view as a divergence.
- `tests/call_result_credit.rs`: 12 rows, incl. the argument move, the born-here consumer shapes, and `eprintln(int.to_string(i))` (16 B per call with the bare lowering, verified by A/B).

**Second batch — the ownership matrix driven from 30 growing rows to 8** (each probed to a named shape; the 8 left are the non-droppable shapes: `List[String]` inputs, the tuple a codepoint decoder returns, the regex engine's sums — the rc-for-tuples/sums arc, not arm defects):
- a prim-tier body ending in a tail call handed its owned locals to the dead epilogue (`float.to_string`: 4 KB per call) → `tail_transfer_ok` keeps the call in non-tail form when the raw-rule frame holds credits (credit row);
- a conditional tail whose arms are all owned took the return +1 (`uint64.to_string`) → `rc_owned_result` looks through `If` / `Match`;
- a loop-form self tail call under the raw rule took +1 per iteration on a pass-through param (every `bytes.read_*_array`) → the param moves;
- arm intermediates never released: `error.chain`, `string.capitalize`, `string.from_bytes`, `list.unique_by`'s seen-keys scratch;
- `list.repeat` / `list.update` allocated but declared `View`; `list.filter` shrank the result's CAP to the kept count so `$free` filed it in the wrong class;
- the matrix now instantiates generic signatures at `Int` (that is what surfaced the last three);
- `string.trim*`'s codepoint decoder returns one packed Int instead of a tuple per scanned codepoint (the tuple is a graveyard block until #2010) — the trim rows are flat, #2005's two named leaks (trim, repeat) are both gone; 5 matrix rows remain, all `List[String]` / sum shapes for #2010.

**Also in this PR.** `bytes.new` zero-fills its block (freed blocks are reused; the `$alloc` reuse path is Coq-pinned, so constructors zero, not the allocator). Emit-time diagnostic knob `ALMIDE_RC_TRAP_DOUBLE_FREE=1` traps a dec at rc 0 (off by default; helper hashes unchanged when off).

**Evidence.** A local differential fuzz campaign against this branch (`xtarget-fuzz run --minutes 8 --jobs 4 --family all --seed 20260907`): 1,655 programs, 1,605 clean, 0 findings (the 8 walls are the known unlinked-twin frontier). Full sweep green: whole `almide-wasm` crate with every ledger regenerated (alloc watermarks only DOWN), `almide test spec/` 436/436, `almide-spine` parity gates, `regex_engine` / `map_set_index_threshold` / `examples/lisp.almd` byte-identical native vs wasm; witness floor and call-boundary certs unchanged; both source gates green.


O6lvl4:

Issue #1998, opened 2026-09-07
Contract-relative survival: declared contract, checked realization, adversarial edit family, one record


O6lvl4:
> **状態** — 追跡・前提待ち: #1997
> **一行で** — 7 段のアークの追跡 issue。step 1-4（#1994 / #1995 / #1996 / #1991）は着地済みで、残りは step 5 = #1997、step 6 = almide-dojo 側の bank、step 7 = #1589。
>
> **やること**
> 1. step 5: #1997（`scoped` の二箇所形）を着地させる。両ターゲットが同じ fragment を honour してから有効化する、が本文の凍結条件。
> 2. step 6: almide-dojo#3 の 8 family bank を 400 cluster pilot まで持っていく（公開比較は 800 × 4 × 3）。almide-dojo#4 が bank を seed 済み、resource 側 4 family は almide-dojo#5 の resource oracle 待ち。
> 3. step 7: #1589（explicit conformance 付きの generic protocol、protocol-as-type は自分のゲートの後ろ、preregistered threshold はそこに記録）。
> 4. L2/L3 の名前は最初の proof obligation が書かれた時に採用する（#2009）。**ツリー確認と食い違う点**: `docs/specs/edit-locality.md` は現在 L1 だけでなく L2（cross-target agreement）と L3（diagnostic locality）も命名済みで、本文の「names L1 only」と、提案されている L2/L3 の割り当てはこの綴りと衝突する。
>
> **完了条件** — step 1-7 の 7 件すべてが閉じ、各 contract が `declaration` / `refusal` / `challenge` の三方向リンクを持ち（#1994 のゲート）、bank が飽和せず（`INCONCLUSIVE_BANK_SATURATED` を出さず）に MSR を測った記録が一つ存在すること。
> **根拠** — 2026-09-21 の確認: #1994 / #1995 / #1996 / #1991 は CLOSED、#1997 と #1589 は OPEN。`docs/contracts/contracts.toml` のヘッダ 60-69 行に `declaration` / `refusal` / `challenge` の欄が実在する。`docs/specs/edit-locality.md:27,33` が L2 / L3 を別の主張に使っている。step 6 の進捗は almide-dojo 側（コメント 2026-09-07）。MSR への効果は **未測定**（本文自身が「No MSR improvement from any of the above until the bank measures it」と書いている）。
[...]

PR #2011, opened 2026-09-07
The build validates every emitted exit against its checked ExitPlan and names a mismatch as compiler defect E083, never a wall

O6lvl4:
Closes #1996 (step 3 of the contract-relative survival arc #1998, ratified 2026-09-07).

Stacked on #2007 (base `fix-2004-arg-temporaries`); the stack is #1992 → #1993 → #2000 → #1999 → #2003 → #2006 → #2007 → this. Retarget to `develop` as the stack lands.

**The build reads every function's bytes back and checks each exit window against its checked ExitPlan; a mismatch is a compiler defect with its own diagnostic, never a wall.**

- `emit_exit` (the one writer of releases, #1995) now also RECORDS: the plan and the function-body byte offset its releases start at (`ExitRecord`, `exit_ledger`). After the function's final `end`, `validate_exits` (`crates/almide-wasm/src/exit_plan.rs`) re-encodes the body, parses it with wasmparser, and for every record — latest start first, so a nested early exit claims its own window — checks: each released credit is `local.get; call $dec_flat`'d in the window; nothing the plan carries is; nothing outside the plan is; the transfer is the instruction the continuation names (`end` / `return` / `return_call*`); and a returning or frame-replacing edge leaves no credit outstanding (the issue's own example: `tail exit leaves an ownership credit outstanding`). It validates wasm, not the emitter's intent.
- `EmitError::OwnershipLowering(OwnDefect)` is a new variant. The CLI (`--target wasm`) prints it as `error[E083]` and stops — it does NOT reroute to the incumbent leg (that would ship a program the checked plan says leaks) — and the text says "compiler contract failure" and never asks the writer to change a valid program. `lower_fn`'s callers treat it as fatal for the whole program (a leak in an unreachable function is still a defect); the gauntlet / burn-up / fuzz harnesses panic on it rather than counting it as a wall.
- `docs/diagnostics/E083.md` (`almide explain E083`, embedded like every code) with the fix-it verdict `not-fixable`; the diagnostic-coverage gate exempts it beside E081/E082 (no source program can trigger it) and names its pin.
- **Negative evidence.** `tests/exit_validation.rs`: the honest emitter passes; with the emitter's test switch `test_omit_first_release` the validator names `fn helper` / `local xs` / `expected: release before the epilogue return` / `emitted: none`. Mutant `ci/mutations/043-exit-release-dropped.patch` (every exit release dropped) is killed by the validator on the first function. `docs/contracts/proven-vs-trusted.md` carries the validator's claim and non-claim row.

**Evidence.** Full sweep green (whole `almide-wasm` crate with ledgers regenerated, `almide test spec/`, `almide-spine`, the explain-docs and E081 root tests, the diagnostic-coverage gate); the validator ran over every function of the corpus with zero findings — which is the standing evidence that #1995's plans and #2007's exits agree byte for byte.

### DIFF
--- a/ci/mutations/043-exit-release-dropped.patch
+++ b/ci/mutations/043-exit-release-dropped.patch
@@ -1,0 +1,12 @@
+diff --git a/crates/almide-wasm/src/exit_plan.rs b/crates/almide-wasm/src/exit_plan.rs
+--- a/crates/almide-wasm/src/exit_plan.rs
++++ b/crates/almide-wasm/src/exit_plan.rs
+@@ -127,7 +127,7 @@ impl Emitter<'_> {
+             if omit_first && k == 0 {
+                 continue;
+             }
+-            self.f.instructions().local_get(idx).call(F_DEC_FLAT);
++            let _ = idx; // [MUTANT: exit releases dropped]
+         }
+         match plan.continuation {
+             Continuation::ReturnSuccess => {
--- a/crates/almide-wasm/src/display.rs
+++ b/crates/almide-wasm/src/display.rs
@@ -499,6 +499,7 @@ fn build_one_scan_helper(
             rc_owned: std::collections::BTreeSet::new(),
             owned_call_marks: Default::default(),
             borrowed_temps: Vec::new(),
+            exit_ledger: Vec::new(),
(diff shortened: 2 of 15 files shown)


O6lvl4:

Issue #2010, opened 2026-09-07
Drop glue for every heap shape: tuples, sums, lists of blocks, Map and Set leave the bump graveyard


O6lvl4:
> **状態** — 実装・判断待ち
> **一行で** — 型駆動 drop glue は Str / Bytes / List / Option / Result / tuple / record / variant / Map / Set まで着地し、graveyard に残っているのは **closure env（+ C-319 cell）と Value 型の payload**、それに call 形 self-assign skip と known-leak 台帳 3 行。
>
> **やること**
> 1. closure env の解放形を裁定する。capture layout は lambda ごとで Fn 型ごとではないので `dec_fn_of(SliceTy::Fn)` は型駆動にできず、closure 限定の runtime dispatch（funcref テーブルの第二列 `slot → drop fn`、または env ヘッダの drop スロット）が要る — 本文の ○×（全形状 static glue）への明示的な例外なので、ここは ruling が先。
> 2. C-319 の cell を refcount する: 今は bind 時に F_ALLOC され、アドレスが各 env にコピーされ、旧占有者は解放されず cell 自身も free されない。capturing env ごとに +1、`DropCell{ty}` glue、cell local の frame-exit 規則。
> 3. `rc_droppable(Fn) = true` を lambda index に鍵を持つ env glue で入れる。受入は PR #2048 の説明にある probe（closure env が 64 B/round）。
> 4. call 形の self-assign skip（`s = set.insert(s, x)` / `s = string.to_upper(s)`）を arm-aware にする（param を consume する table fn には正しく、borrow だけする native arm には leak）。
> 5. Value 型の payload（Map の値が Value、Value payload）に typed drop を入れる。
> 6. known-leak 台帳の残り 3 行（`regex.replace` 16 / `regex.replace_first` 16 / `regex.split` 64）を落とす。
>
> **完了条件** — `rc_droppable_ty`（`crates/almide-wasm/src/rc_ownership.rs:299`）の `_ => false` 側に heap 形状が残らず、`crates/almide-wasm/tests/golden/native-own-known-leaks.txt` が 0 行、#2048 の probe の closure env が flat、alloc 台帳が all-down で再生成されること。
[...]

PR #2014, opened 2026-09-07
Every flat-payload block shape is droppable: tuples, Option and Result of scalars leave the bump graveyard

O6lvl4:
Refs #2010 (stage 1 of the drop-glue arc — the flat shapes, the ○× ruled ○ by "どんどん進めていいよ").

Stacked on #2013; retarget to `develop` as the stack lands.

**Every block shape with no heap interior is droppable now, and `$dec_flat` alone releases it.** `rc_droppable` (rc_ownership.rs) admits a tuple, an `Option`, a `Result`, and a `List` whose payload slots are all non-Str scalars (Unit and fn values count as flat); the IR pre-pass `arg_temps.rs` binds the same shapes when a call or a constructor hands them to a language consumer. No glue is emitted and no arm needs a field audit: a flat block has no shared field to dangle. Records and variants wait for stage 1b — the `mut` param move-mode rewrite (C-132) writes a record back through the Assign route, whose ownership is not yet audited for a droppable record (`mut_param_effect_never_err` double-freed the `Tally` on the first run).

**What the widening exposed, and the rule it taught.** Four sums arms hand their INPUT block back as the result on one branch (`result.map` / `map_err`'s pass-through side, `option.or_else` on some, `option.filter` when kept). Declared `Borrow` in / `Owned` out, the scope released a born-here `some(30)` before the bind could take its share — `f 0` and a double free (alias_combinator_rc, caught by the corpus burn-up with the trap knob). The declaration for a pass-through arm is **`Retain` in, `Owned` out**: a temporary's credit moves into the result; a Var takes the share, and on the branch that builds a fresh block instead the retained share is a leak, never a dangle (the per-branch identity is #1996's C item). `none` joins `rc_certainly_fresh`: it is NULL_ADDR, so an `if c then some(x) else none` tail is owned as a whole.

**Evidence.** Credit rows `a_flat_tuple_result_is_released` / `a_flat_option_result_is_released` (were 32 B per call on the graveyard); the ownership matrix now measures `Option[Int]` / `Int?` / `(Int, Int)` / `Result[Int, Int]` results — 124 rows, 116 flat, the same 5 growing rows as before (all `List[String]` / regex sums, stages 2–3); corpus burn-up and stock-wasmtime gate clean with `ALMIDE_RC_TRAP_DOUBLE_FREE=1`; full sweep green; 23 alloc watermarks moved, 22 down and one up by a fixed 32 B (`binary_search_duplicate_keys`: a born-here `some` block is now bound for the consumer and lives to the frame exit — a constant, not per call).


O6lvl4:

PR #2015, opened 2026-09-07
A callback lambda inside a map literal lowers on the structural leg: the local-collection walk gains its MapLiteral arm

O6lvl4:
Refs #1423 (stage 5: the campaign's "walled but declared available" class).

The 2026-09-08 local campaign dumped four walled programs; every one reduced to the same shape: a callback lambda inside a MAP LITERAL's key or value (`["k0": (if list.all(xs, (x) => …) then … else …)]`). The HOF arm inlines the callback, so its params must be locals — but `collect_binds_data` had no `MapLiteral` arm, the params were never collected, and the bind walled as `bind:unmapped`; the program then fell to the incumbent, whose typed twins (`list.zip_with_x`, `list.drop_while_heapelem`, `list.scan_x`, `result.flatten_x`) are unlinked — the four "unlinked stdlib/runtime call" wall reasons of the campaign, all one defect.

Evidence: the four dumped programs print identically on native and the forced structural leg; `tests/map_literal_callback.rs` pins key- and value-position callbacks; corpus burn-up green; `almide test spec/wasm_cross/` green.

### DIFF
--- a/crates/almide-wasm/src/collect.rs
+++ b/crates/almide-wasm/src/collect.rs
@@ -103,6 +103,17 @@ pub(crate) fn collect_binds_data(
             }
             Ok(())
         }
+        // A map literal's keys and values are expressions like a list's
+        // elements: a callback lambda inside one (`["k": (if list.all(xs,
+        // (x) => …) then …)]`) needs its params as locals too — the
+        // `bind:unmapped` wall the 2026-09-08 campaign dumped (#1423).
+        IrExprKind::MapLiteral { entries } => {
+            for (k, v) in entries {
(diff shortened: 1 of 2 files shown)


O6lvl4:
Re-measured with the same seed as the campaign that dumped the walls (`xtarget-fuzz run --minutes 5 --jobs 4 --family all --seed 20260907`): 945 programs, 925 clean, **0 findings, 0 walls** — the first campaign had 8 walls (all four dumped programs were this shape) on the same seed.

O6lvl4:

PR #2016, opened 2026-09-07
value.object retains the pairs spine the Value keeps, instead of borrowing it

O6lvl4:
Refs #2010 (stage 2, the first finding of the spine-sharing audit).

`value.object(pairs)` declared its argument `Borrow` while the Value keeps that list spine and reads its entries through it forever. Today nothing frees a list of tuples, so the borrow was invisible; the moment list spines become droppable (the stage-2 experiment), a born-here pairs literal was released under the object and every codec fixture printed `{"":null}`. With `Retain` — the same declaration `value.array` already carries — all 20 codec fixtures of the experiment are identical again. No observable change on the shipped set; the declaration is now honest, which is what the arm-args discipline is for.

### DIFF
--- a/crates/almide-wasm/src/value.rs
+++ b/crates/almide-wasm/src/value.rs
@@ -84,7 +84,12 @@ impl Emitter<'_> {
             // insertion order IS the block, exactly the interp's ordered
             // object model.
             ("object", [pairs]) => {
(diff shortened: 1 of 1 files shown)


O6lvl4:

Issue #1909, opened 2026-09-04
matrix.from_bytes_f32_le with an offset near i64::MAX passes the bounds check through addition overflow and reads outside the buffer on the structural leg


O6lvl4:
`matrix.from_bytes_f32_le(f, 9223372036854775807, 1, 2)` over an 8-byte buffer: native answers the all-zero matrix (OOB → zeros); the structural leg'\''s in-bounds test `offset + r*c*width <= len` wraps and passes, and the read returns a subnormal (`0.0000000000000000000000000000000000000000001793662034335766`). Finding of the 60-minute dispatch 33858798267 (seed 541740772278 index 10760). Expected: an overflow-safe check (`offset <= len && r*c*width <= len - offset`), zeros on the far side.

O6lvl4:
Fixed in #1913 (landed on develop). Fixture: spec/wasm_cross/matrix_from_bytes_offset_domain.almd, contract in docs/contracts/contracts.toml.

PR #2017, opened 2026-09-07
The matrix byte loaders zero their cells explicitly instead of trusting fresh memory

O6lvl4:
Refs #2010 (stage 2 audit, second finding) and #1909's OOB→zeros edge.

`matrix.from_bytes_f32_le` / `_f16_le` answer an out-of-range offset with the all-zero matrix by allocating the output and filling nothing — which was correct only while nothing was ever freed. `$alloc` hands back reused blocks unzeroed (the Coq-pinned reuse path), so once list spines are released the "zero" matrix printed subnormal garbage (the stage-2 experiment's last diverging fixture, `matrix_from_bytes_offset_domain`). The constructor now `memory.fill`s its cell region, exactly as `bytes.new` learned to in #2004. No observable change on the shipped set.


O6lvl4:

PR #2019, opened 2026-09-07
Every List spine is droppable: a list of any element type leaves the bump graveyard, elements follow with the typed glue

O6lvl4:
Refs #2010 (stage 2a of the drop-glue arc).

Stacked on #2017; retarget to `develop` as the stack lands.

**A List of ANY element type is droppable now: `$dec_flat` releases its spine, and the elements keep the credits they hold today.** `rc_droppable` admits every `List`, and the IR pre-pass binds any born-here list handed to a language consumer. Releasing the elements is stage 2b (typed glue).

**How it was made safe.** Widened blindly, the corpus burn-up diverged 40 fixtures. Run in a side tree with `ALMIDE_RC_TRAP_DOUBLE_FREE=1`, every one of the 40 reduced to two defects, each now its own PR beneath this one: `value.object` borrowed the pairs spine the Value keeps (#2016, `Retain`), and the matrix byte loaders trusted fresh memory to be zero for their OOB→zeros edge while `$alloc` hands back reused blocks unzeroed (#2017, an explicit fill — the `bytes.new` lesson). With those two, the burn-up is green with every spine droppable.

**Evidence.** Credit row `a_list_of_strings_spine_is_released` (every `List[String]` return was 32 B per call on the graveyard); the ownership matrix now measures `List[String]` / `List[Bytes]` / `List[List[Int]]` / `List[(Int, Int)]` / `List[Int?]` results — 138 rows, 119 flat; the spine rows (`list.join`, `string.join`) are gone and the 16 growing rows left are exactly the element leaks inside a fresh list (`string.chars` / `split` / `lines`, `list.enumerate` / `zip` / `chunk` / `window`, `bytes.chunks` / `split`, `regex.find_all` / `split`), stage 2b's worklist. Corpus burn-up and stock-wasmtime gate clean with the trap armed; full sweep green: 175 alloc watermarks moved, 169 down, 6 up by fixed 32–224 B (born-here lists now bound for their consumer and living to the frame exit — constants, not per call).


O6lvl4:

PR #2020, opened 2026-09-07
Release list elements through typed drop glue and give every copied spine its own element credits (#2010 stage 2b)

O6lvl4:
Stacked on #2019 (stage 2a). Closes the element half of #2010 for List spines whose elements are heap handles (Str / Bytes / nested List).

## Mechanism

- `Helper::DropList { elem_dec }` — the typed drop glue: the spine's credit down; at zero every element handle is released through `elem_dec` (`$dec_flat` for a leaf, the inner list's own glue for a nested one), then the spine freed. `dec_fn_of(ty)` chooses it; every owned local / droppable param records its type (`rc_own(idx, ty)`, `owned_ty`) and `emit_exit`, the bind-route rebinds and the borrow-pool release all go through it. The E083 validator accepts the glue indices as releases.
- `Helper::IncElems` / `CopyElems` / `CowElems` — every duplication of a spine's handle slots takes element credits: the native arms' `memory_copy` sites (slice / take / drop / take_end / drop_end / remove_at / chunk / windows / window / reverse / insert / unique / unique_by / filter / concat / rest patterns / map.keys / map.values / pop), `$block_copy` (set / update / swap / sort / sort_by / global init) and `$cow` (index assignment, append route, mut-var read).
- A handle stored into a block that has no typed drop yet (Option payloads of `list.get/first/last/find/min/max/reduce`, enumerate and `map.to_list` pairs) takes +1 — leak-not-dangle until stage 2c. The arms that allocate their Option block now declare it `Owned`.
- A `View` result of an arm whose scope releases a borrowed temporary is promoted before the release (`promote_escaping_view`): `list.get(list.chunk(xs, 2), 1)` read a row of the freed chunk list.
- The bind / assign share applies to EVERY droppable shape, not only List / Str / Bytes: `let n1: Int? = n ?? none; let n2: Int? = n ?? none` owned the nested option's payload twice and double-freed it (pre-existing, stage 1 — the earlier binary traps on it too).

## Evidence

- Corpus burn-up + gauntlet green **under `ALMIDE_RC_TRAP_DOUBLE_FREE=1`** (new CI step): five fixtures only showed their dangling reads with the trap armed (`list_chunk_windows`, `list_total_order`, `map_set_in_place`, `closures_and_variants`, `path_extension_hidden`), and `unwrap_or_source_survives` / `unique_non_flat_element` trapped.
- Ownership matrix: known-leak ledger 16 → 5 rows (bytes.chunks/lines/split, list.chunk/window/windows, regex.find_all, string.chars/lines/split, url.encode_component now flat).
- Allocation watermarks: 65 fixtures down, 2 up — `map_index_string_key_shapes` +32 B (one Option payload share) and `unwrap_or_option_payload` +64 KB: its loop leaks a `Result[Float?, String]` per iteration (a nested payload block, not droppable until 2c); the old code freed the inner option out from under the leaked Result.
- Two new credit rows (`a_list_of_strings_releases_its_elements`, `a_copied_spine_holds_its_own_element_credits`); exit validation, heap probe, arm-args gate green; codopsy A 91 unchanged.

## Left for stage 2c

Typed drops for Option / Result / tuple / record blocks with handle payloads (turning today's leak-not-dangle +1s into releases), and the 5 remaining matrix rows.


O6lvl4:

Issue #1963, opened 2026-09-06
The thesis has no current measurement: MSR is 147 days stale and MiniGit saturates at 20/20 for every language — build a task set that discriminates


O6lvl4:
> **状態** — 計測器・ゲート・判断待ち
> **一行で** — プロジェクトが唯一の指標と宣言している軸に、今の数字が無い（README の scorecard は 2026-04-12、lang-bench の MiniGit snapshot は全言語 20/20 で飽和して弁別できない）。ハーネス半分は着地済みで、残りは task 半分と走行の許可。
>
> **やること**
> 1. task 半分を almide-dojo に作る（4 タスク × 6 編集の順序列、`seed/<lang>/` と `edits/0k-*/{prompt.md, scope.txt, visible.sh, hidden.sh, solution/, wrong/}`。hidden oracle はツリーに入れない）。repo 境界により MSR の task 素材は Dojo 側。
> 2. `research/benchmark/lang-bench/runner_edits.rb --tasks <dojo>/tasks/xlang` の `--dry-run` bank ゲートを 5 言語ぶん通す（編集前ツリーが落ちる / solution が通る / wrong は visible 通過・hidden 落ち / overlay が `scope.txt` の内側）。
> 3. 走行の go-ahead（spend と model pin）を確定させる。コメントに見積りが載っている: N=5 で 600 セッション ≈ $540、N=10 で 1,200 セッション ≈ $1,085。
> 4. 台帳を publish する。`(task, language)` の全升を出し、状態は `measured | not-run | harness-limitation | inconclusive-saturated` で型付ける（負けの升と未計測の升も空欄にしない）。
>
> **完了条件** — `(task, language)` の全升が埋まった台帳が publish され、MSR-seq が Wilson 95% CI つきで出て、かつ `INCONCLUSIVE_BANK_SATURATED` が出ない（= task set が飽和していない）こと。
> **根拠** — 本文の実測: README の MSR scorecard は 2026-04-12（コメント 2026-09-19 の時点で 160 日）、lang-bench MiniGit snapshot（2026-07-15、20 trials）は全言語 20/20。2026-09-21 のツリー確認: ハーネス半分は着地済み — `research/benchmark/lang-bench/runner_edits.rb`、`aggregate_edits.rb` が存在し、`runner_multi.rb:87` に 5 つ目の toolchain エントリがある。task 半分と走行は未着手。
[...]

PR #2024, opened 2026-09-07
The lang-bench runner takes the model from the environment and documents the Claude 5 roster

O6lvl4:
The MiniGit runner had `claude-sonnet-5` hard-coded, so the measurement plan in #1963 / #1617 could not name the current family. `MODEL` (and `MODEL_TAG`, derived) now come from the environment with the same default, and the header lists the roster to measure: claude-fable-5-1, claude-opus-5, claude-sonnet-5, claude-haiku-4-5-20251001.

Refs #1963, #1617.


O6lvl4:
Withdrawn: the request was meant for another session.

Issue #1675, opened 2026-08-29
Derived `Codec.decode` errors carry no path and no received type — three different failures all say `expected Str`


O6lvl4:
`type X: Codec` is the language's answer to "validate this JSON" — the type *is* the schema, and `X.decode` is generated. The happy path and the round-trip are correct. The failure path returns a bare fixed string with **no path and no received type**, so three different failures in three different places are indistinguishable.

## Repro

```almide
type Address: Codec = { city: String, zip: String }
type User: Codec = { name: String, address: Address, tags: List[String] }
```

| input | `User.decode` today | zod 4.5 `safeParse`, same schema |
|---|---|---|
| `address.city` is a number | `expected Str` | `[address.city] invalid_type: expected string, received number` |
| `address.zip` absent | `missing field 'zip'` | `[address.zip] invalid_type: expected string, received undefined` |
| `tags[1]` is a number | `expected Str` | `[tags.1] invalid_type: expected string, received number` |
| `name` is null | `expected Str` | `[name] invalid_type: expected string, received null` |

Three of the four collapse to the same six characters. `missing field 'zip'` is the only one that names anything, and it still omits the `address.` prefix — at depth two that is already ambiguous, and a config with a repeated field name at two depths is unresolvable.

## Where it comes from

`runtime/rs/src/value.rs` (mirrored to `crates/almide-codegen/src/generated/rust_runtime.rs`):

```rust
pub fn almide_rt_value_as_string(v: &Value) -> Result<String, String> {
    match v { Value::Str(s) => Ok(s.clone()), _ => Err("expected Str".to_string()) }
}
pub fn almide_rt_value_field(v: &Value, key: &str) -> Result<Value, String> {
    ...  Err(format!("missing field '{}'", key))
}
```

The generated decode threads these straight through with `?`:

```rust
pub fn User_decode(_v: Value) -> Result<User, String> {
    let _f_name: String = (almide_rt_value_as_string(&(almide_rt_value_field(&_v, "name"))?))?;
    ...
}
```

so there is no frame anywhere that could accumulate a path. The decoder *knows* the field name at every step — it is a literal in the generated code — and throws it away.

## Why it matters here specifically

[...]

PR #2051, opened 2026-09-08
Derived Codec.decode pays its per-field error frame only on the error path (#2050)

### LINKED ISSUE #2050: Derived Codec.decode pays the #1675 error-path wrap on the SUCCESS path: match _r.clone() plus a key .to_string() per field doubles decode time
## What

The per-field error-wrapping frames #1675 added to derived `T.decode` (`T___erratw_<field>`) run on the **success** path too: each field lookup `match`es a clone of the `Ok` payload and allocates the key (`"name".to_string()`) before it is known that anything failed. For the 8-field `User` record of the new `decode` perf row (#2047: nested `Address`, a 3-element `List[String]`) that is 11 frames per decode.

## Measured (M4 Pro, 2M decodes, floor-subtracted, quiet box; PR #2047)

| variant | ns/op |
|---|---:|
| native, as emitted today (slot hint included) | 412 |
| native, the `___erratw_*` frames collapsed by hand in the emit | 205 |
| hand-written Rust reference (`rust-ref/decode.rs`) | 205 |

The whole 2.0× of the row is these frames; the slot hint (#2047 PR4) bought ~10 ns. `research/benchmark/perf/decode` is the ratchet row (REPORTED, baseline ≈ 2.4 whole-process).

## Where

- `crates/almide-codegen` derive: the frame is emitted around every `field(_v, "k")?` of a derived decode. The success path should be a plain `?` on the borrowed lookup; the path/received-type prose is built only in the `Err` arm.
- The key string must not be allocated on the success path — the literal is already `&'static str`; only the error message needs an owned copy.

## Done when

- the `decode` row reads ≤ 1.1 against the reference (native), with the ratchet baseline lowered in the same PR;
- #1675's error text is byte-identical (its spec fixtures and C-084 `spec/wasm_cross/codec_decode_field_order.almd` unchanged);
- the wasm leg is untouched or improves (verify with the row's wasm build once #2046 closes).

Found by the #1679 decode-corpus work; the ADR-0018 `Value::Str(AlmideStr)` change is orthogonal (it removes the payload copies, not these frames).

O6lvl4:
## Why

#1675 gave every field of a derived `T.decode` a path frame: the field's `Result` goes through a per-type worker `T___erratw_<mangle>(r, "key")` that passes an `Ok` through and splices the key into an `Err`. On the native leg that frame is paid on **every** field, failed or not — the walker renders the pass-through match as `match _r.clone() { Ok(_) => _r, … }` (a clone of the decoded payload) and the call site hands the worker an owned `"key".to_string()`. On the 8-field `decode` row (#2047) that is eleven frames per decode and the whole 2x against the hand-written reference.

## What

A native-only nanopass, `DecodeErrFramePass` (`crates/almide-codegen/src/pass_decode_err_frame.rs`), after BuiltinLowering and before NormalizeRuntimeCalls. It rewrites each `T___erratw_M(<res>, "key")?` into

```rust
<res>.map_err(|_we| almide_rt___err_at(_we, "key".to_string()))?
```

so the success path is the plain `?` on the field's own borrowed lookup, the key stays a `&'static str` literal until an error needs an owned copy, and the error bytes are the ones the worker produced (`almide_rt___err_at` is what the worker called). Workers no call site names any more are dropped from the program, so no dead `T___erratw_*` fns are emitted. The frontend IR and the wasm leg are untouched — the worker is still the shape both legs lower; the pass runs after BorrowInsertion so the lookups inside the frame keep their borrow decoration (BorrowInsertion treats `InlineRust` args as opaque).

Before / after for the `User` record of the row:

```rust
let _f_name: String = (User___erratw_string(almide_rt_value_as_string(almide_rt_value_field_ref(_v, "name")?), "name".to_string()))?;
let _f_name: String = (almide_rt_value_as_string(almide_rt_value_field_ref(_v, "name")?).map_err(|_we| almide_rt___err_at(_we, "name".to_string())))?;
```

## Numbers

`research/benchmark/perf/decode` (from #2047, not merged yet), M4 Pro, 2M decodes, floor-subtracted, min of 5 interleaved runs, built the way the row builds (`almide build --release`); the box was shared with sibling agents, so these are mins, not medians:

| variant | ns/op |
|---|---:|
| native before (installed 0.62.0) | 518 |
| native after (this PR) | 224 |
| hand-written reference `rust-ref/decode.rs` | 188 |

2.75x → 1.19x. The derived decode body now matches the reference line for line; compiling the emitted Rust with the reference's own rustc flags reads 181 ns/op against the reference's 182, so the remaining ~35 ns of the row is in the `almide build --release` profile, not the emit — a separate follow-up if the row is to be anchored.

**Ratchet:** the `decode` row and its baseline live on #2047's branch, not on develop, so this PR changes no `scripts/check-perf-ratio.sh` / `perf-ratio-baseline.txt` line. Lowering the row's baseline (2.46 whole-process on the day it landed) depends on #2047 landing first; it should be re-read on that branch with this pass applied.

## Gates

- `tests/codec_decode_success_path_test.rs` (new, in the mold of `codec_decode_borrows_input_test.rs`): every field frame is a `.map_err` on the borrowed lookup, no `.clone()` and no key `.to_string()` outside the error closure in the decode bodies, no `___erratw_` fn emitted, and the error text (`, at id` / `, at address.city` / `missing field 'zip', at address`) unchanged
- `cargo test --release --test codec_decode_success_path_test --test codec_decode_borrows_input_test --test codegen_snapshot_test`: green, no snapshot changes (none covers a derived Codec)
- `almide test spec/stdlib/` (162 files) and `spec/lang/` (211 files): green on the default leg; `spec/stdlib/` also green with `--target native`
- C-084 `codec_decode_errors.almd`, C-085 `codec_float_int.almd`, the four C-087 json fixtures, and #2047's `codec_decode_field_order.almd`: byte-identical across native (this PR), wasm (untouched leg, the oracle) and the installed 0.62.0 native binary
- A module-defined Codec (`import geo` / `geo.parse`) decodes and reports `at y` identically on both legs (the module-prefixed worker name path)
- `bash scripts/check-contracts.sh`, `bash scripts/check-readme-numbers.sh`: green; `cargo clippy --release -p almide-codegen`: no warnings in the new pass

## Found on the way (pre-existing, not this PR)

A record with an `Option[T]` field where `T` has a defaulted field emits invalid Rust on the native leg (`almide_rt_value_decode_option_custom(_v, …)` — the by-value option driver handed a borrowed `_v`; rustc E0308). Reproduces identically on the installed 0.62.0. Repro:

```almide
type Addr: Codec = { city: String, zip: String = "0" }
type Rec: Codec = { name: String, alt: Addr? }
```

Closes #2050

### DIFF
--- a/crates/almide-codegen/CLAUDE.md
+++ b/crates/almide-codegen/CLAUDE.md
@@ -18,7 +18,7 @@ PatternLiteralGuard → ResolveCalls → RegionWindow → BoxDeref → LICM → 
 MatrixShapeSpec → ConstFold → IntrinsicLowering → BorrowInsertion →
 TailCallOpt → CaptureClone → CloneInsertion → MatchSubject → EffectInference →
 StdlibLowering → AutoParallel → ResultPropagation → BuiltinLowering →
-DecodeSlotHint → Peephole → RustLowering → FanLowering → NormalizeRuntimeCalls →
+DecodeSlotHint → DecodeErrFrame → Peephole → RustLowering → FanLowering → NormalizeRuntimeCalls →
 IrLinkFlatten → RangeCountingVars → TopLetStorage
 
 - Each pass: `impl NanoPass { fn run(&self, program, target) -> PassResult }`
--- a/crates/almide-codegen/src/lib.rs
+++ b/crates/almide-codegen/src/lib.rs
@@ -31,6 +31,7 @@ pub mod pass_borrow_inference;
 pub mod pass_box_deref;
 pub mod pass_builtin_lowering;
 pub mod pass_decode_slot_hint;
+pub mod pass_decode_err_frame;
 pub mod pass_capture_clone;
 pub mod pass_shared_cell_borrow;
 pub mod pass_clone;
--- a/crates/almide-codegen/src/pass_decode_err_frame.rs
+++ b/crates/almide-codegen/src/pass_decode_err_frame.rs
@@ -1,0 +1,174 @@
+//! DecodeErrFramePass (#2050): the per-field error frame of a derived
+//! `T.decode` costs nothing on the success path.
+//!
+//! #1675 gave every field of a derived decode a path frame: the field's
+//! `Result` goes through a per-type worker `T.__erratw_<mangle>(r, "key")`
+//! that passes an `Ok` through and splices `key` into an `Err`. The worker
+//! is the shape BOTH legs lower (a top-level fn matching on its own param —
+//! see `wrap_err_at` in the frontend), but on the native leg it is paid on
+//! EVERY field, failed or not: the walker renders the pass-through match as
+//! `match _r.clone() { Ok(_) => _r, … }` — a clone of the decoded payload —
+//! and the call site hands the worker an owned `"key".to_string()`. On the
+//! 8-field `decode` perf row that is eleven frames per decode and 2× the
+//! hand-written reference.
+//!
+//! This pass rewrites each `T.__erratw_M(<res>, "key")?` on the native leg
+//! into `<res>.map_err(|_we| almide_rt___err_at(_we, "key".to_string()))?`:
+//! the success path is the plain `?` on the field's own result, the key is
+//! the `&'static str` literal until an error actually needs an owned copy,
+//! and the error bytes are the ones the worker produced (`almide_rt___err_at`
+//! is the runtime the worker called). A worker no call site names any more
+//! is dropped from the program, so the emitted Rust carries no dead
+//! `T___erratw_*` fns.
+//!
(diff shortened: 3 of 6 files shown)


O6lvl4:
Stacked on #2047 (the slot hint) so the queue can take both: `lib.rs`/`target.rs` keep both passes, DecodeSlotHint before DecodeErrFrame, and the frame test now reads the slot-indexed lookup (`almide_rt_value_field_ref_at(_v, "id", 0)?` wrapped in the `.map_err`). Both passes compose on the C-084 fixture (both legs byte-identical). The `decode` row's REPORTED baseline is left as #2047 sets it; lower it on a quiet machine once both are on develop.

Issue #2045, opened 2026-09-08
Stream fusion of |> chains never fires on the Rust target: IterChain is not produced, every stage materialises its list


O6lvl4:
> **状態** — 実装・着手可
> **一行で** — `|>` チェーン自体は Rust ターゲットで融合されるようになった（`pass_stream_fusion.rs`、ablation は `ALMIDE_STREAM_FUSION_OFF`）。残っているのは **head** — `list.range` が今も `Vec<i64>` を建てるので、fold が一度消費するだけの range まで確保される。
>
> **やること**
> 1. range SOURCE を `(start..end)` として融合し、`Vec` を建てない（`codegen/templates/rust.toml:473` が exclusive、`:476` が inclusive）。
> 2. 上限だけは落とさない: 今その観測を作っているのは `try_reserve_exact`（C-197 の `Error: out of memory` + exit 1）で、`proofs/domain-edges.toml:281-293` の `list.range` セル群がそれに対して宣言されている。融合形はループ前の明示的な count guard として同じ上限を運ぶ。
> 3. 本文 Acceptance の残り: `listbuild-comb` が両 runner class で `listbuild-append` 以下、Rust の iterator reference に勝つなら `VICTORY` 行へ、出力は両レッグでバイト一致。
> 4. 本文 Ask の (3)（ablation の env knob）は着地済み — `ALMIDE_STREAM_FUSION_OFF`、idiom ゲートがそれを使っている。
>
> **完了条件** — `list.range` を一度だけ消費するチェーンの Rust 出力に `Vec` 確保が無く、`proofs/domain-edges.toml` の `list.range` セルが同じ観測を保ち、`listbuild-comb` が両 runner class で `listbuild-append` 以下（Ir 比が `IDIOM_CEILING` 以下）、出力が両レッグでバイト一致すること。
[...]

PR #2053, opened 2026-09-08
Stream fusion runs on the RuntimeCall shape after StdlibLowering, so a pure pipe of list combinators lowers to one iterator expression

### LINKED ISSUE #2054: Native/interp: egg list rules substituted a map callback's body per downstream stage, duplicating its side effects (map(f) |> filter(p) printed f's println three times per element)
## What

The egg saturation list rules (`EggSaturation`, before #2053) rewrote `xs |> list.map(f) |> list.filter(p)` by substituting `f`'s body into the filter predicate and into the result, so an effectful `f` ran up to three times per element. Observable divergence: with a `println` in `f`, native and interp printed

```
m1 f1 m1 m1 m2 …
```

while the wasm leg printed the spec order `m1 m2 m3 f1 f2 f3`.

## Found / fixed

Found by the probe in the #2045 stream-fusion work; fixed in PR #2053 (egg now lifts only `matrix` calls; the dead intercept is deleted; `spec/lang/stream_fusion_test.almd` pins side-effect order as values through a captured-`var` log for map→filter, map→fold, map→take, map→any, take-then-map, and the fallible `!` form, on both legs).

## Ledger

Cross-target divergence class; the fixture is in `spec/lang` (3-way oracle). If a `spec/wasm_cross` contract row is wanted for the ordering law, file it under the evaluation-order contract in the follow-up, not here.

Closes with #2053.

O6lvl4:
`map(f) |> filter(p)` could duplicate and reorder `f`'s side effects because egg substituted its body into downstream stages. Meanwhile, the intended stream-fusion intercept never ran: it matched module calls after intrinsic lowering had already turned them into runtime calls.

This change removes the egg list rewrites and introduces StreamFusion after StdlibLowering. A literal callback becomes a typed iterator stage without an Rc callback allocation. Adjacent stages merge only when their evaluation order is unobservable: callbacks must be proven pure, at most one stage may be partial, and stages before short-circuiting must be proven total. Otherwise the intermediate collection remains and preserves eager evaluation.

Termination uses a least fixed point, so recursive functions and their callers are never treated as terminating merely because they have no I/O. For example, `[1] |> list.map((x) => diverge(x)) |> list.take(0)` keeps the eager map; it must not silently become an empty list. Unknown Rust templates, effectful/unknown runtime modules, mutable borrows and unproven function-value callbacks also cannot justify fusion. Purity checks are split into disjoint expression families, with unknown nodes rejected.

`ALMIDE_STREAM_FUSION_OFF=1` disables the pass. The existing flat-map array optimization and typed collector callbacks remain supported. The update preserves develop's DecodeErrFrame pass and resolves the pipeline-documentation conflict without rewriting branch history.

Validation on the current branch:

- Default and native Almide suites: **438/438 files each**.
- Codegen crate: **56 unit + 10 integration tests**.
- Nanopass, codegen snapshots and end-to-end rendering: **34 / 13 / 2**.
- New regressions reject direct/mutual recursion as total, propagate acyclic proofs, and check opaque Rust, unknown/effectful calls, callback references and mutable borrows. The existing stream suite pins both values and callback side-effect order.
- Clippy completes, with no warning on the fusion file; pass roster, file discipline and diff checks pass.
- Codopsy 2.2.0: fusion file **100/A**; the existing codegen aggregate remains **89/B**. No scoring rules were relaxed.

Pinned compiler references and the reviewed evaluation semantics are recorded in `docs/stream-fusion-safety-review.md`. Performance timings have not been remeasured after the termination-proof tightening; #2045 remains open for its broader performance requirements.

Refs #2045
Closes #2054

Namespace-safety follow-up: a pure module function no longer contributes its bare name to the root-function purity/totality proof sets. A real imported-module regression reproduced suppressed `println` calls under `map |> take(0)` before the fix and now preserves output `0, 1, 2, 0`. Details are recorded in the safety review.

## Merge-queue integration

Fixes #2054.


O6lvl4:
`bash scripts/check-perf-ratio.sh` on this branch (rebased onto develop's victory-row gate), M4 Pro, shared machine 20 min after a reboot (Spotlight/Xprotect indexing plus another agent's `cargo test`, load 12–30):

| row | run 1 | run 2 | band |
|---|---|---|---|
| fasta | 1.179 | 1.328 | baseline 1.152 +40% ok |
| fft | 1.165 | 1.261 | baseline 1.176 +40% ok |
| nbody | 0.869 | 0.999 | baseline 0.997 ok |
| spectralnorm | 0.990 | 0.999 | baseline 1.001 ok |
| listbuild / -append / -comb | 1.576 / 1.589 / 1.612 | 1.525 / 1.586 / 1.687 | reported |
| listbuild-idiom | 1.021× | 0.994× | reported (Ir gates it; valgrind absent here) |
| binarytrees / treealloc | 0.154 / 0.301 | 0.208 / 0.218 | victory ok; ablations 3.4–4.6 ≥ 1.30 ok |
| ablation/fasta, fft, nbody, spectralnorm | 0.688, 0.891, 0.878, 0.890 | 0.541, 0.584, 1.022, 1.122 | **floor 0.90 — red both runs** |

The `ablation/*` rows are `ALMIDE_DISABLE_OPT=1` (almide-optimize's fold/DCE/propagate), not this pass, and the gate measures the two legs minutes apart. Interleaved on the same binary, 5 runs each, two rounds, native min/median (s):

| | fasta | fft |
|---|---|---|
| optimized, fused | 0.4110/0.4155, 0.4120/0.4306 | 0.1788/0.1800, 0.1793/0.1808 |
| `ALMIDE_DISABLE_OPT=1`, fused | 0.4109/0.4137, 0.4181/0.4217 | 0.1770/0.1809, 0.1792/0.1840 |
| optimized, `ALMIDE_STREAM_FUSION_OFF=1` | 0.4146/0.4152, 0.4167/0.4180 | 0.1780/0.1803, 0.1786/0.1810 |
| both off | 0.4287/0.4356, 0.4186/0.4234 | 0.1780/0.1789, 0.1796/0.1798 |

All four combinations sit within ±2% of each other — ablated/optimized ≈ 1.00 for both benches, with or without this pass — so the 0.54–0.89 gate readings are the machine's load moving between the two legs, and the stream-fusion knob has no effect on these rows either way. The CI runner will give the ratchet its reading; nothing in this PR touches those rows' pipeline.

PR #2055, opened 2026-09-08
fan runs on a thread per core natively for pure callbacks over Send-safe scalars: fan.map and list ops under a fan block reach the parallel runtime twins, ablated with ALMIDE_FAN_SEQUENTIAL

O6lvl4:
Explicit `fan.map` and list operations inside `fan { ... }` use the native parallel runtime only when their callback and captured types satisfy Send checks and transitive purity. Implicit list pipelines remain sequential and can fuse.

This revision reconciles current develop and #2053, preserves DecodeErrFrame, and removes the retired AutoParallel registration. Plain functions can contain effects, so a declaration flag is insufficient: the routing now checks reachable helper bodies. Closed helpers with scalar parameters may mutate invocation-local state; the analysis retains every effectful operand and does not infer termination. Module-qualified function names cannot accidentally prove a same-named root function pure. StreamFusion leaves explicit fan operations available for routing.

Validation on this revision: all 438 Almide test files pass on native and default (426 structural WASM, 12 native fallbacks); codegen tests and the local-write/effect regression pass. File discipline and pass roster pass. Codopsy 2.2.0: local-state analysis and shared purity analysis each 100/A with unchanged thresholds. All 41 nanopass tests and 13 codegen snapshots pass. GitHub CI is pending.

Fannkuchredux emits the parallel list call and matches the sequential ablation and handwritten Rust output. Seven interleaved n=10 measurements give 54.16 ms parallel, 179.27 ms sequential, and 172.47 ms Rust (native/Rust 0.314, ablation 3.310). These are local results; CI retains its performance gate. Raw evidence and pinned Rust/Gleam research are in `docs/fan-parallel-safety-review.md` and `research/benchmark/perf/results/2026-09-09-fan-repair.json`.

This is an increment toward #2044, not its closure: Bytes-returning mandelbrot callbacks remain outside the Send-safe subset. Nested callback boxing is handled by existing #2064. Parallel speculative work can still encounter a later trap beyond the first returned Err; that pre-existing limitation remains documented in the runtime.

## Merge-queue integration


O6lvl4:

PR #2064, opened 2026-09-08
A closure literal inside a fan callback keeps the uniform Rc<dyn Fn> boxing, so a list combinator nested in fan.map builds on native

### LINKED ISSUE #2059: Native: a list combinator nested inside a fan.map callback emits Rust that fails E0282 (type annotations needed); wasm runs it
## Repro (0.62.0 and develop, native)

```almide
effect fn main() -> Unit = {
  let rows = [1, 2, 3]
  let out = fan.map(rows, (r) => [r, r + 1] |> list.map((x) => x * 2))
  println(int.to_string(list.len(out)))
}
```

`almide run` on native fails in rustc with E0282 (type annotations needed) for the inner `list.map` closure inside the `fan.map` callback. The wasm leg runs it and prints `3`. Same with `list.filter` or `list.fold` nested in the callback.

## Where

The fan callback is emitted as a closure whose element type is left for rustc to infer, and the nested combinator's own closure parameter then has no inferable type. The outer `list.map` (non-fan) path annotates; the fan path does not. `crates/almide-codegen` fan lowering (`pass_rust_lowering_fan.rs` after PR #2055 / `pass_fan_lowering.rs`) and the walker's lambda parameter rendering.

## Done when

- the repro compiles and prints `3` natively, byte-identical to wasm, under a `spec/wasm_cross` fixture (C-004 family);
- matrix: nested {map, filter, fold, flat_map} inside {fan.map, fan {} block} all compile natively — extend the fan fixture matrix, not a point test.

Found while landing #2044's first increment (PR #2055); not fixed there.

O6lvl4:
A list combinator nested inside a fan callback failed to compile natively because closure boxing exempted the entire fan subtree. Nested closure literals now receive the ordinary Rc<dyn Fn> representation; only the callback consumed by the fan runtime is re-tagged or unboxed for that API.

Fixes #2059.

The branch is reconciled with current develop and the reviewed #2053/#2055 changes. The fixture covers map, filter, fold, and flat_map in fan.map and fan blocks, plus fan.any. Native and structural WASM produce identical output. The six legacy renderer ledgers retain their existing baselines and add only the new fixture row.

Validation: 43 nanopass tests and 13 codegen snapshots pass on the final code. All 438 default test files pass on the final compiler (426 WASM, 12 native fallbacks). All 438 native files pass on the integrated compiler before the behavior-preserving module split; the final native fixture produces the same output. Contract ledger, file discipline, and pass roster pass. Statement rewrites are separated from closure boxing to keep both files below 800 lines; Codopsy 2.2.0 reports 100/A for both and 90/A for codegen overall with unchanged thresholds.

Research and parallel-routing safety evidence are recorded in docs/fan-parallel-safety-review.md, using the Rust and Gleam compilers cloned under ../almide-references. GitHub CI must pass before the merge queue admits this revision.

## Merge-queue integration


O6lvl4:

PR #2074, opened 2026-09-09
Borrow native comparisons and list fields without redundant copies

### LINKED ISSUE #2066: native: String field comparison clones both operands (t.kind == want emits t.kind.clone(); a literal emits "...".to_string())
## Summary

On the native backend, `==` between a `String` record field and another string clones both
operands to compare them. The receiver is already borrowed, and comparison needs no ownership.
When the right side is a literal, a fresh `String` is allocated on every evaluation.

## Repro

```almide
type Tok = { kind: String, text: String }

fn same(t: Tok, want: String) -> Bool = t.kind == want
fn is_newline(t: Tok) -> Bool = t.kind == "newline"
```

## Generated Rust (almide 0.62.0)

```rust
pub fn same(t: &Tok, want: &str) -> bool {
    almide_eq!(t.kind.clone(), want)
}

pub fn is_newline(t: &Tok) -> bool {
    almide_eq!(t.kind.clone(), "newline".to_string())
}
```

`almide_eq!` expands to `($a) == ($b)`, so both clones are pure waste. `t` is `&Tok` and
`want` is already `&str`. The expected output is `t.kind.as_str() == want` and
`t.kind == "newline"`, neither of which allocates.

Note the asymmetry in `same`: the parameter is lowered to `&str` correctly, but the field
access is not.

## Why it matters

This sits on the hot path of any tokeniser or matcher. In O6lvl4/gramide the parser compares
a token's `kind` and `text` against a grammar terminal on every match attempt, tens of times
per token. Replacing exactly these clones by borrows in the generated Rust, and nothing else,
took a 140,287-byte Go file from 100 ms to 82 ms per parse (rustc -O, 10 iterations, best of 3).
The remaining clone removals in the same function are filed separately.

## Environment

almide 0.62.0, rustc 1.96.1, macOS 26.3 arm64. The repro uses `process.exec` only to force the
native codegen leg under `almide test`; the wasm leg is not what this report is about.

### LINKED ISSUE #2068: native: for over a list field clones the list and then clones every element (h.kids.clone().iter().cloned())
## Summary

`for x in rec.list_field { ... }` clones the whole list and then clones every element while
iterating, even when the record is borrowed and the body only reads.

## Repro

```almide
type Hit = { kids: List[Int], pos: Int }

fn total(h: Hit) -> Int = { var s = 0; for k in h.kids { s = s + k }; s }
```

## Generated Rust (almide 0.62.0)

```rust
pub fn total(h: &Hit) -> i64 {
    let mut s: i64 = 0i64;
    for k in h.kids.clone().iter().cloned() {
        s = (s).wrapping_add(k);
    }
    s
}
```

Two copies of the data for a read-only loop. `for k in h.kids.iter()` (with `k` bound by
reference, or `.copied()` for `Copy` element types) has the same meaning here.

Where the collection is owned and dead after the loop, `for k in h.kids` should move instead.

## Why it matters

In O6lvl4/gramide the PEG engine appends a matched rule's children into an accumulator with
exactly this loop, once per sequence element. Since the element type is a tree node that owns
its own children, `.iter().cloned()` deep-copies each subtree. This is part of the same 3.3x
that #2067 covers; the three clone classes were measured together.

## Environment

almide 0.62.0, rustc 1.96.1, macOS 26.3 arm64. The repro uses `process.exec` only to force the
native codegen leg under `almide test`.

O6lvl4:
Native comparisons and loops unnecessarily copied borrowed record fields. String equality and ordering now borrow stable operands, read-only list-field loops borrow their collections and elements, and the final use of a proven owned field transfers consumed elements with into_iter().

Fixes #2066. Fixes #2068.

Comparison operands that can mutate each other retain snapshot evaluation. Loop bodies that write their parent retain the input snapshot. A positive ownership set distinguishes owned value bindings and parameters from borrowed parameters; the borrowed-parent regression caught an invalid provisional move before it was fixed.

This is the next commit after the reviewed linear stack #2053 → #2055 → #2064, because develop's merge queue uses REBASE. The new change itself is the final commit; the preceding commits belong to those existing PRs. Ownership decisions stay in nanopasses, with a consumed-loop annotation read by the renderer. Projection and loop-render helpers are separated to stay under the file-size limit.

Validation: all 439 Almide test files pass on native and default (427 WASM, 12 native fallbacks); 43 nanopass tests, 13 snapshots, 11 loop tests, and one native/structural string-comparison matrix pass. Both original allocation regressions fail before their fix. Codopsy 2.2.0 reports codegen 90/A with unchanged thresholds. Pass roster and diff checks pass.

Pinned Rust compiler research, safety boundaries and test evidence are in docs/native-read-borrow-review.md. Reference source is cloned under ../almide-references/rust.


O6lvl4:

PR #2075, opened 2026-09-09
Reclaim structural WASM decoder scratch storage

### LINKED ISSUE #2046: Derived T.decode in a loop retains ~130-800 bytes per call on the wasm leg; 1M decodes of an 8-field record peak at 786 MB, 5M is out of memory
Found while landing the #1679 decode benchmark corpus (`research/benchmark/perf/decode/decode.almd`): the wasm leg of a loop of derived `T.decode` calls grows without bound; the native leg does not.

## Measured (arm64, wasmtime, `/usr/bin/time -l` peak RSS)

| program | iterations | native | wasm (worktree, develop 8e384213b) | wasm (v0.62.0 installed) |
|---|---:|---:|---:|---:|
| 8-field `User` (nested `Address`, 3-element `List[String]`), `acc + weight(u)` | 1M | 1.7 MB | 786 MB | 1118 MB |
| same, 5M | 5M | — | **out of memory** (3.7 GB, exit 1) | — |
| same record, loop body reduced to `ok(u) => u.id` | 1M | — | 798 MB | 1118 MB |
| `type Flat: Codec = { id: Int, name: String }`, `ok(u) => u.id` | 100k / 1M | — | 21.8 MB / 139 MB | — / 200 MB |

So the slope is ~130 B per decode for a 2-field record and ~800 B for the 8-field one — roughly the decoded record (its strings and list) plus intermediates, never released. The output is byte-identical across legs on every run that finishes, so this is a lifetime bug, not a value bug.

Repro (the body is the whole loop; `json.parse` runs once, outside it):

```almide
type Flat: Codec = { id: Int, name: String }
let doc = """
  {"id": 42, "name": "Ada Lovelace"}
  """
effect fn main() -> Unit = {
  let v = json.parse(doc)!
  var acc = 0
  for i in 0..<1000000 {
    acc = acc + (match Flat.decode(v) { ok(u) => u.id, err(_) => -1 })
  }
  println("acc: ${acc}")
}
```

`almide build f.almd --target wasm -o f.wasm && /usr/bin/time -l wasmtime run f.wasm` — compare against `almide build f.almd` native.

Consequence for the benchmark row: `decode` runs its wasm leg only at `bench.py`'s verify arg (1000); the timing arg (5M) is native/rust only until this is fixed — the `bench.py` SUITE entry carries the note.

O6lvl4:
Repeated derived decoding retained 128 bytes per successful two-field decode on structural WASM and exhausted a 4 MiB heap at 100000 calls. The same workload now completes within that cap.

Bind temporary Result/Option operands at extraction sites so existing frame exits release their wrappers. Preserve tail carrier transfers and mutable-call ABI handling. Normalize value-match arms to one owned credit, and release the intermediate missing-field error string. This also fixes retention on type-error and missing-field paths.

Validation:
- Six regression tests cover flat and nested decoding, two error paths, mutable argument writeback, and 200000 tail calls. Decoder heap watermarks are identical at 1000 and 8000 calls.
- All 439 Almide test files pass (427 WASM, 12 native fallback).
- Call-result ownership, native-result ownership, exit validation, witness call boundaries, and structural witness floor pass.
- Allocation ledger passes in normal mode after regeneration: 79 watermarks decrease, none increase, no new structural refusals.
- Codopsy WASM source score: 91/A, existing thresholds unchanged.
- Koka ownership research and design details: docs/wasm-decoder-reclaim-review.md.

This is stacked on #2074 (which includes #2053, #2055, and #2064); only the final commit is this fix. Keep the dependency order when merging.

Fixes #2046.


O6lvl4:

Issue #2084, opened 2026-09-10
almide test reports 'All N test file(s) passed' when zero tests ran: the summary and --json count files, never tests, so a zero-match --run and a file with no test block are both indistinguishable from a real pass


O6lvl4:
Found on develop @ dff9a458f while exercising `almide test` on shebang scripts.

`almide test` never reports how many *tests* ran. Its summary counts *files*, so a run in which zero tests executed is byte-identical to a run in which every test passed.

## Two reproductions

**A zero-match `--run` filter:**

```
$ almide test fib.almd --run "zzz-no-such-test" --json
{"file":"fib.almd","status":"pass","exit_code":0,"failures":[]}
$ almide test fib.almd --run "fib" --json
{"file":"fib.almd","status":"pass","exit_code":0,"failures":[]}
```

Identical, and both exit 0. (`fib.almd` holds one `test "fib"` block.)

**A file with no `test` block at all:**

```
$ cat notest.almd
fn main() -> Unit = { println("no tests here") }
$ almide test notest.almd
1 via WASM, 0 via native fallback, 0 failed (of 1 files)
All 1 test file(s) passed        # exit 0
```

A file that cannot possibly pass a test is counted as a passing test file.

## Cause

`src/cli/commands.rs:201` and `src/cli/commands.rs:757`:

```rust
err(&format!("\nAll {} test file(s) passed", test_files.len()));
```

`test_files.len()` is the number of files the walker collected. Nothing in the summary — or in `--json`, whose object carries `file` / `status` / `exit_code` / `failures` and no count — is derived from the number of test blocks actually executed. `-v` adds pipeline notes, not a count.

## Why it matters

`cargo test` prints `running 0 tests` and `N filtered out`, so a filter typo is visible. Here the two runs cannot be told apart by a human reading the terminal or by a script reading the JSON. A CI step written as `almide test --run "integration"` goes green forever after the tests it names are renamed, and a `.almd` file that loses its `test` block during a refactor keeps reporting as a passing test file.

**Not currently reachable through this repo's CI** — `git grep 'test --run'` over `.github`, `scripts` and `Makefile` is empty; CI runs the bare `almide test` / `almide test --target wasm` over directories that do contain tests. So this is a latent footgun rather than an active hole, which is why I have not asked for a priority above P-low. The exposed surface today is `almide-pkg-ci.yml:30` (`out=$(almide test 2>&1)`) for a package that ships no tests.

## Precedent: this failure class has bitten once already

[...]

Issue #1166, opened 2026-08-10
The local wasm test harness silently converts runtime traps into native fallback, hiding miscompiles CI's Test WASM catches


O6lvl4:
Found while landing #1165 (the #1134 Shape 2 burn-down): `cmd_test_fast`'s fallback routing (`src/cli/commands.rs`) treats **every** wasm outcome the leg can't pass — emitter gap, `wasm:skip`, compile error, **and a runtime trap** — as "fall back to native", reported as `N via native fallback` with zero distinction:

```rust
WasmTestOutcome::Fail { file, .. }        // <- a runtime TRAP lands here
| WasmTestOutcome::CompileError { file, .. }
| WasmTestOutcome::Skip { file, .. } => fallback.push(file),
```

Consequence: the `indirect call type mismatch` trap in `fallible_hof_test` (the voided fallible-each callback, fixed in #1165's second commit) ran **silently native for its whole life locally** — `almide test` said "1 via native fallback, 0 failed" — while CI's `Test WASM` job (a separate, honest runner) failed the PR. The local ladder was structurally unable to see the class of bug the burn-down series exists to flush out; the developer's first signal was a red on the PR.

A wall (compile-time refusal) falling back is the designed contract — honest incompleteness. A **trap** falling back is different in kind: the wasm leg *compiled* the file, claimed it, and produced wrong behavior at runtime; masking that behind the native leg's green is exactly the silent-fallback class #782 retired at the emitter level.

Fix direction (small): split `WasmTestOutcome::Fail` (ran, trapped/failed) from the benign classes in the fallback accounting — still fall back so the suite stays authoritative, but print the file + trap on stderr and count it in the summary line (e.g. `343 via WASM, 20 via native fallback (1 after a wasm TRAP), 0 failed`). Optionally an `ALMIDE_TEST_STRICT_WASM=1` mode that fails instead, matching CI's Test WASM semantics for pre-push runs.

Evidence: #1165's Test WASM red vs the same tree's local `0 failed`; the trap repro is pinned in that PR's comment thread.

PR #2086, opened 2026-09-10
The wasm test leg honours --run instead of dropping it at the signature

### LINKED ISSUE #2085: almide test --run is discarded on the wasm leg: cmd_test_wasm drops the filter at its signature, so --target wasm fails on excluded tests and the default run reports a fabricated 'after a wasm TRAP' divergence
Split out of #2084, where the corrected evidence lives (that issue keeps the separate "the summary counts files, never tests" defect).

`almide test --run <pattern>` selects tests on the native legs and is discarded entirely on the wasm leg, so the two legs run different test sets from the same command line.

## Reproduction

```almide
fn main() -> Unit = { println("x") }
test "alpha passes" { assert_eq(1, 1) }
test "beta fails"   { assert_eq(1, 2) }
```

```
$ almide test filt.almd                             ; echo $?
… 1 failed (of 1 files)                             1     ← correct: beta fails
$ almide test filt.almd --run alpha                 ; echo $?
0 via WASM, 1 via native fallback (1 after a wasm TRAP), 0 failed
All 1 test file(s) passed                           0
$ almide test filt.almd --run alpha --target wasm   ; echo $?
  trapped at: test: beta fails ...
  assertion failed: left == right
0 passed, 1 failed (of 1 files)                     1     ← FALSE RED
```

`--run alpha` excludes `beta fails`. The wasm leg runs it regardless.

Measured on develop @ dff9a458f, `almide 0.62.0` built from that commit, wasmtime on PATH.

## Cause

`src/cli/commands.rs:518` drops the parameter at the signature:

```rust
pub fn cmd_test_wasm(file: &str, _run_filter: Option<&str>) {
```

The native paths (`cmd_test`, `cmd_test_fast`, `cmd_test_update_snapshots`, `cmd_test_json`) all build harness argv through `test_harness_args` (`src/cli/test_report.rs:47`):

```rust
pub fn test_harness_args(run_filter: Option<&str>) -> std::sync::Arc<Vec<String>> {
    let mut args = vec!["--nocapture".to_string()];
    args.extend(run_filter.map(|f| f.to_string()));
    std::sync::Arc::new(args)
}
```

`cmd_test_wasm` never calls it, and `compile_and_run_wasm_test` takes no argv, so the module is invoked with none.

## Two user-visible consequences, opposite in sign

- **False red.** `--target wasm` with `--run` fails on a test the filter excluded (exit 1 above). CI's Test WASM lane is the leg most likely to be pointed at a filtered subset.
- **False green plus a fabricated divergence.** On the default dual run the wasm leg traps on the excluded test, the native fallback honours the filter and passes, and the summary prints `1 after a wasm TRAP`. That counter is meant to mean "the targets disagree on observable behaviour"; here it is an argument-plumbing artifact. `ALMIDE_TEST_STRICT_WASM` (#1166) turns that artifact into a hard failure, so the strict lane can be failed by a bug that has nothing to do with target parity.

The second is the worse one: it puts noise into the exact signal the repo uses to detect real native/wasm divergence.

## Fix

The plumbing already exists — `crates/almide-wasm-run/src/host.rs:596` offers `run_wasm_real_stdin_args(bytes: &[u8], args: &[String])`. So:

1. `cmd_test_wasm` accepts `run_filter` (drop the leading underscore) and builds `test_harness_args(run_filter)`, exactly as the native paths do.
2. `compile_and_run_wasm_test` takes that argv and passes it down.
3. The wasm invocation uses `run_wasm_real_stdin_args` instead of the argless entry point.

Check the synthesized `__test_runner` (`crates/almide-mir/src/pipeline_test_runner.rs`) reads its filter from program args on the wasm target the same way it does natively — if it does not, that is the real work and step 3 is only the delivery.

## Acceptance

- `almide test <f> --run <p> --target wasm` runs exactly the tests `--run <p>` selects.
- For the same `<f>` and `<p>`, the native leg, `--target wasm`, and the default dual run agree on the verdict and the exit code.
- The default dual run of the filtered file prints no `after a wasm TRAP` note, and `ALMIDE_TEST_STRICT_WASM=1` does not fail it.
- A fixture pins filter parity across the three legs — a file with one passing and one failing test, filtered to the passing one — so the next refactor cannot silently unplumb the argument again. A signature that ignores a parameter is invisible to every existing gate, which is why this survived.
- `--run` with no match behaves identically on all three legs (whatever #2084 settles that behaviour to be).

O6lvl4:
Closes #2085.

`almide test --run <pattern>` selected tests on the native legs and was discarded on the wasm leg, so one command line ran two different test sets.

```
$ almide test filt.almd --run alpha --target wasm   ; echo $?
  trapped at: test: beta fails ...                    1     ← before: FALSE RED
$ almide test filt.almd --run alpha --target wasm   ; echo $?
                                                      0     ← after
```

`cmd_test_wasm` dropped the parameter at its signature (`_run_filter`), and the wasm leg shells out to the wasmtime CLI with no program args, so there was nowhere for a pattern to arrive. The native paths build harness argv through `test_harness_args`.

### Where the fix had to go

Not argv. The synthesized `__test_runner` (`crates/almide-mir/src/pipeline_test_runner.rs`) reads no program arguments at all — it emits one unconditional call per `test` fn. So the wasm leg selects at **synthesis time**: non-matching test fns are dropped from the IR before the runner is built.

That makes the predicate the parity risk. The native leg's answer comes from the Rust test harness substring-matching the *emitted* fn name, so `rust_safe_fn_name` moves from almide-codegen (which almide-mir cannot see) to `almide_base::names`, re-exported so every existing caller keeps its spelling. `test_name_matches_filter` is now the only admitted way to ask "does this test run under this pattern", and both legs call it.

The drop happens after the wall checks, not before, so `--run` cannot quietly change which files reach the wasm leg.

### Consequences fixed

- **False red.** `--target wasm` with `--run` failed on a test the filter excluded.
- **False green with a fabricated divergence.** On the default dual run the wasm leg trapped on the excluded test, the native fallback honoured the filter and passed, and the summary printed `1 after a wasm TRAP` — an argument-plumbing artifact in the counter that is supposed to mean "the targets disagree on observable behaviour". `ALMIDE_TEST_STRICT_WASM` (#1166) turned that artifact into a hard failure.

Verified: `--run alpha --target wasm` → 0, `--run alpha` default → 0 with no TRAP note, `ALMIDE_TEST_STRICT_WASM=1 --run alpha` → 0, `--run beta --target wasm` → 1, unfiltered → 1.

### The gate

A signature that ignores a parameter is invisible to every existing gate — running MORE tests than asked still passes when they all pass, so a green suite proved nothing. `tests/wasm_test_filter_parity_test.rs` asserts by **module identity** rather than by grepping labels (test labels lower to runs of `i32.store8`, not to data strings): filtering a two-test file down to one test must render byte-for-byte the module a file containing only that test renders. A/B against the pre-fix behaviour: 3 of its 4 tests fail, all 4 pass after.

`docs/specs/cli.md` now states the matching rule, which was undocumented: a case-sensitive substring match against the emitted name, so `test "beta fails"` is selected by `beta` and `beta_fails` but not by `beta fails`. That is native's existing behaviour; this PR makes the wasm leg surprising in the same way rather than differently.

### Test status

- `cargo test --workspace`: green except `charge_probe_gate`, which fails identically on unmodified `origin/develop` (`spec/wasm_cross/fuel_bounded_boundary.almd` asserts `expected: 2, found: 1` on both binaries) — pre-existing, untouched here.
- `almide test`: 440/440 files pass (428 via WASM, 12 via native fallback).
- No output text changed, so the summary-parsing callers (`scripts/ci-acceptance-ring.sh`) are unaffected.

#2084 (the summary counting files rather than tests) follows on top of this branch.

### DIFF
--- a/crates/almide-base/src/lib.rs
+++ b/crates/almide-base/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod intern;
 pub mod span;
 pub mod diagnostic;
+pub mod names;
 pub mod profile;
 
 // Re-export commonly used items at crate root
--- a/crates/almide-base/src/names.rs
+++ b/crates/almide-base/src/names.rs
@@ -1,0 +1,79 @@
+// Symbol spelling shared by the Rust emitter and the wasm test-runner synthesis.
+//
+// `rust_safe_fn_name` lives here — not in almide-codegen where it is emitted —
+// because the two test legs must agree on ONE spelling. The native leg compiles
+// `test "…"` to a Rust fn whose name is this function's output and hands
+// `--run <pattern>` to the Rust test harness, which substring-matches THAT name;
+// the wasm leg synthesizes its runner in almide-mir and must select exactly the
+// same tests. When the two spellings drifted apart the filter was simply dropped
+// on the wasm side and nothing noticed (#2085), so the rule now has one home and
+// `test_name_matches_filter` is the only admitted way to ask the question.
+
+/// Sanitize an IR function name into a Rust-emittable identifier.
+///
+/// Spaces and punctuation fold to `_`; parentheses vanish; operator characters
+/// spell out. A name containing any non-ASCII character also gets an FNV-1a
+/// suffix, because the fold above maps every non-ASCII byte to `_` and two
+/// distinct names would otherwise collide on one skeleton.
+pub fn rust_safe_fn_name(raw: &str) -> String {
+    let s = raw
+        .replace([' ', '-', '.', ',', ':', '[', ']'], "_")
+        .replace(['(', ')'], "")
+        .replace('+', "_plus_").replace('/', "_div_").replace('*', "_mul_")
+        .replace('=', "_eq_").replace('!', "_bang_").replace('?', "_q_")
+        .replace('<', "_lt_").replace('>', "_gt_")
+        .replace('|', "_pipe_").replace('&', "_amp_").replace('%', "_mod_");
+    let mut safe: String = s.chars().map(|c| if c.is_ascii_alphanumeric() || c == '_' { c } else { '_' }).collect();
+    if raw.chars().any(|c| !c.is_ascii()) {
+        let mut h: u64 = 0xcbf2_9ce4_8422_2325;
+        for b in raw.as_bytes() {
+            h ^= *b as u64;
+            h = h.wrapping_mul(0x0000_0100_0000_01b3);
+        }
+        safe.push_str(&format!("_{:08x}", (h >> 32) as u32 ^ h as u32));
+    }
+    safe
+}
+
+/// Does the test whose IR name is `ir_name` run under `--run <pattern>`?
+///
+/// The native leg's answer is the Rust test harness's: a case-sensitive
+/// substring test against the EMITTED fn name, so the pattern is compared
+/// against `rust_safe_fn_name(ir_name)` and not against the `test "…"` label.
+/// A pattern spelled with the label's spaces therefore selects nothing on both
+/// legs — surprising, but identically surprising, which is the property the
+/// wasm leg has to reproduce.
+pub fn test_name_matches_filter(ir_name: &str, pattern: &str) -> bool {
+    rust_safe_fn_name(ir_name).contains(pattern)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{rust_safe_fn_name, test_name_matches_filter};
+
+    #[test]
+    fn same_skeleton_nonascii_names_stay_distinct() {
+        let a = rust_safe_fn_name("__test_almd_a: 値を取る旗に値が無ければ断る");
+        let b = rust_safe_fn_name("__test_almd_a: 旗の値は位置引数に混ざらない");
+        assert_ne!(a, b);
+    }
+
+    #[test]
+    fn ascii_names_are_untouched_by_the_hash_rule() {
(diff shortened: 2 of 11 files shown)


O6lvl4:
Correction to this PR body: it claimed `charge_probe_gate` was a pre-existing failure on `origin/develop`. That was wrong.

The three local failures I saw (`charge_probe_gate`, `kernel_conformance_native_stdout_matches_kernel_trace`, and `wasm_cross_target_spec` with 72 cross-target divergences) were all one stale 1.6 GB native build cache under `$TMPDIR/almide-run`. Clearing it turns all three green:

```
$ almide run spec/wasm_cross/assert_abort_msg.almd     # stale cache
Error: assertion failed
  at: line 9
  expected: 2
  found: 1
$ rm -rf "$TMPDIR/almide-run" "$TMPDIR/almide-test"
$ almide run spec/wasm_cross/assert_abort_msg.almd     # matches the wasm leg
Error: assertion failed: one exceeds two
  at: line 8
```

The A/B I ran to "prove" it was pre-existing compared this branch's binary against an unmodified `develop` build — but **both read the same cache**, so identical output proved nothing about the code. `cargo test --workspace --no-fail-fast` is green on a cold cache.

No change to the code in this PR; only the claim in its body was wrong.

PR #2166, opened 2026-09-13
Judge every entry under src/ when almide check is given no file inside a package

### LINKED ISSUE #2165: almide check with no arguments reports No errors while a sibling entry point in the same package does not type-check
## Summary

`almide check` with no arguments type-checks only the package entry (`src/mod.almd`). Sibling entry points in the same package are never looked at, and it still prints `No errors found`. `almide test` does not reach them either.

## Repro

A package with three entry points that share one module:

```
almide.toml
src/mod.almd     # the shared logic, package entry
src/main.almd    # a native CLI  (effect fn main)
src/wasm.almd    # a wasm entry  (effect fn main)
spec/*.almd      # tests against mod.almd
```

Break a signature in `mod.almd` and update only `wasm.almd`, leaving `main.almd` calling the old arity:

```
$ almide check
No errors found

$ almide test
All 1 test file(s) passed
43 tests in 1 file

$ almide check src/main.almd
error[E004]: board.discord_body() expects 5 argument(s) but got 4
error[E005]: argument 'logo_url' expects String but got List[board.Change]
error[E005]: argument 'changes' expects List[board.Change] but got String
3 error(s) found
```

Confirmed by breaking each file in turn and running bare `almide check`:

| broken file | bare `almide check` |
| --- | --- |
| `src/mod.almd` | 6 error(s) found |
| `src/main.almd` | **No errors found** |
| `src/wasm.almd` | **No errors found** |

## Why it matters

This is not hypothetical — it cost me four consecutive red CI runs. I changed a function signature in `mod.almd`, updated the wasm entry and the tests, and never touched the CLI entry. Locally `almide test` was green every time. CI ran `almide check src/main.almd` explicitly and was red every time, and I did not look until someone told me.

`No errors found` is the part that makes it expensive. If it said nothing, or named what it checked, the gap would be visible. As it stands, the most natural local command reports success over code that does not compile.

## Suggestion

Any of these would close it:

- bare `almide check` walks every `.almd` reachable in the package (or at least every file with a `main`), not just the package entry
- it names what it checked: `Checked src/mod.almd — No errors found`
- `almide test` type-checks the package before running the spec

The first is what I expected as a user: "check this project" rather than "check one file I did not name".

## Workaround

Listing the entry points explicitly, which means knowing the gap exists:

```bash
almide check src/main.almd
almide check src/wasm.almd
almide test
```

## Version

- v0.62.0 (released)
- develop `112df3ed8` (2026-09-13)

O6lvl4:
Closes #2165.

`almide check` with no file inside a package checked `src/mod.almd` alone and printed `No errors found` while `src/main.almd` in the same package did not type-check. This makes the bare form walk every `.almd` under `src/` (`mod.almd`, `main.almd`, then the rest in path order), name each on its own line, and summarize:

```
src/mod.almd: ok
src/main.almd: ok
src/wasm.almd: ok
Checked 3 files — No errors found
```

A broken sibling exits through the same path a single-file check does, after its diagnostics, so the reader sees how far the walk got.

- The single-file form is byte-for-byte unchanged (`No errors found`), so every existing reader of that line still matches.
- `--json` / `--effects` are per-file reports with no multi-file shape; they keep the single-entry resolution.
- `resolve_file` is untouched — `build`, `run` and the rest still pick the first entry as before.

Tests: `tests/check_package_entries_test.rs` (walk + naming + order, broken sibling turns it red, single-file output unchanged, `--json` keeps single entry). Related check tests still pass.


O6lvl4:

Issue #2167, opened 2026-09-13
A record or variant that derives Ord sorts on native and walls on both wasm legs, while C-053 claims they are byte-identical


O6lvl4:
Found 2026-09-13 while closing #2154 (the compound sort-key divergence). Measured on `almide` built from `112df3ed8`. All three are the SAME diagnostic, and none is a cross-target divergence — both legs agree. What they are is a promise nothing keeps.

## 1. The hint names `records` and the checker rejects them

```almide
type K = { a: Int, b: String }

fn main() -> Unit = {
  let ks = [K { a: 2, b: "x" }, K { a: 1, b: "y" }]
  for k in ks |> list.sort { println("${k.a} ${k.b}") }
}
```

```
error[E030]: type 'K' has no ordering — cannot be used with list.sort
  hint: Ordering needs Int, Bool, String, Float, or lists/tuples/records of those. …
```

The hint tells the reader the exact thing they just tried. The same applies through a key: `list.sort_by(xs, (x) => K { … })` is rejected by the same rule with the same hint.

This is the LLM-facing surface rule (#1483) in its sharpest form: a model that reads the hint writes the record spelling next, and gets the same error again.

## 2. C-053 claims variant ordering the compiler refuses

C-053's statement reads, verbatim: "…for EVERY totally-ordered element type — Int, Bool (false<true), String, Tuple (lexicographic), nested List (lexicographic with shorter-first tiebreak), Option (none<some), **and variants (tag order)**".

```almide
type Level = | Low | Mid | High
fn main() -> Unit = { for l in [High, Low, Mid] |> list.sort { println("${l}") } }
```

is E030 on both targets. The ledger asserts a behaviour that does not exist; no fixture covers the cell, which is why it survived.

## 3. The hint's own text is malformed

It renders as one line with runs of interior spaces:

```
  hint: Ordering needs Int, Bool, String, Float, or lists/tuples/records of those.                  Map, Set, and function values have no order; Float inside a compound                  element has none either (compare via an explicit key instead).
```

A multi-line source string is being joined without collapsing its indentation.

## What the fix has to decide

Which side moves. Either:

- **implement** record (field order) and variant (tag order) ordering — the emitter side is nearly there, `emit_val_cmp` already chains tuple fields and would chain record fields the same way, and `SliceTy::Named` reaches its `unsup` arm today; or
- **withdraw** both claims — correct the hint to the set the checker actually admits, and issue a ratified correction on C-053 the way C-054 carries one.

[...]

Issue #2155, opened 2026-09-13
The idiomatic word-count shape cannot build for stock WASI: a tuple sort key walls the structural leg and list.group_by is unlinked on the incumbent


O6lvl4:
Found 2026-09-13 by the stdlib-linked arena (docs/roadmap/active/arena-breakthroughs.md §2.4). Source: `../almide-references/bench-sources/stdlib-regime/wordfreq.almd` — the spelling CLAUDE.md and the cheatsheet steer a model toward (`list.map |> list.group_by |> map.map |> list.sort_by`).

```
$ almide check wordfreq.almd --target wasm
wall: unlinked stdlib/runtime call(s) with no wasm definition: list.group_by_x — rendering them would emit a dangling `(call $…)` (invalid wasm).
wall (structural leg, the default): list-sort-by-key:Tuple(4)
error[E082]: both wasm legs refused this program
```

Two independent holes, both on the recommended path:
1. **`list.group_by` has no wasm definition on the incumbent** (`list.group_by_x` unlinked) — register it in `self_host_registry.rs` or route it to the structural leg.
2. **A tuple sort key walls the structural leg** (`list-sort-by-key:Tuple(4)`). `list.sort_by` takes a key, not a comparator, so `(0 - c, w)` is the only clean spelling of "count desc, word asc". Either lower tuple keys on the structural leg or give `list.sort_by` a comparator twin so the documented idiom has a wasm-legal spelling.

The wall is honest (not a miscompile); the sibling issue is the incumbent accepting the same key and trapping. Workaround measured: a padded String key (`../almide-references/bench-sources/stdlib-regime/wordfreq_strkey.almd`) builds on the structural leg — and then hits the `group_by` constant in the next sibling issue.

O6lvl4:
Sibling: #2154 is the incumbent leg accepting the same tuple key and trapping; #2156 is the group_by constant once the key is a String.

O6lvl4:
Closing: the state this issue reports no longer exists on the default leg, verified by re-running the issue's **own source** on 0.63.0 rather than by reading the fix.

The issue's headline was `error[E082]: both wasm legs refused this program`. Today, on `bench-sources/stdlib-regime/wordfreq.almd` — the file this issue names:

```
almide check wordfreq.almd --target wasm   → No errors found
almide build wordfreq.almd --target wasm   → Built (structural leg, trusted)
almide run   wordfreq.almd  vs  --target wasm → byte-identical
```

[...]

PR #2168, opened 2026-09-13
Order a compound sort_by key on the structural leg and refuse it on the incumbent

### LINKED ISSUE #2154: A tuple sort key builds on the incumbent wasm leg as "verified" and traps at run time with an indirect call type mismatch, while native prints the answer
Found 2026-09-13 by the stdlib-linked arena (docs/roadmap/active/arena-breakthroughs.md §2.4). Reproduced by hand on almide 0.62.0, wasmtime 47.0.2. Source: `../almide-references/bench-sources/stdlib-regime/probe_imp_1k.almd` (1,000 draws; the 2M version dies as `out of memory` instead, which hides the real trap).

```almide
let top = map.entries(counts)
  |> list.sort_by(((w, c)) => (0 - c, w))   // Map[String, Int] entries, tuple key
  |> list.take(10)
```

```
$ almide check probe_imp_1k.almd --target wasm
No errors found
$ almide build probe_imp_1k.almd --target wasm -o p.wasm
Built p.wasm (11157 bytes, incumbent v1 leg, verified — …)
$ wasmtime p.wasm
error while executing at wasm backtrace:
    0:   0x1a3e - __sbr_init
    1:   0x1b29 - list.sort_by_rc
    2:    0x77c - main
wasm trap: indirect call type mismatch
$ almide run probe_imp_1k.almd
canjbfa 2
ciya 2
…
```

The structural leg walls this shape (`wall (structural leg, the default): list-sort-by-key:Tuple(4)`, see the sibling issue), so the build falls to the incumbent, which accepts it and emits a `call_indirect` whose type does not match the sort-by-key callee. Native and wasm disagree on observable output (stdout empty + non-zero exit vs the answer): `I-divergence`. Two extra faults ride along:

1. `almide check --target wasm` says "No errors found" for a program the build route cannot run — the check-time verdict (#1922) does not cover the incumbent fallback's mis-render.
2. The artifact is labelled **verified**. Whatever the certificate covers, it did not cover the sort-by-key call table.

Same program with a String key (`"${string.pad_start(int.to_string(n_words - c), 8, "0")}|${w}"`) routes to the structural leg and runs correctly: `../almide-references/bench-sources/stdlib-regime/wordfreq_imp_strkey.almd`, 8,210 B, 201 ms.

O6lvl4:
Closes #2154.

## What was wrong

`entries |> list.sort_by(((w, c)) => (0 - c, w))` — count descending, word ascending, the word-count idiom and the shape an LLM writes first — type-checked, ran on native, and **walled the structural leg** (`list-sort-by-key:Tuple(4)`). The build therefore fell to the incumbent, whose cached-key twins declare `f: (Int) -> Int` and compare the cached key as a raw i64. The key closure hands back an i32 **handle**, so the sort's `call_indirect` was an `indirect call type mismatch` trap at run time — out of an artifact `build --target wasm` had just reported `verified` — while native printed the answer.

Had the two signatures lined up, it would have been worse: the sort would have ordered by the key's **address**.

C-053 already promised `sort_by` was byte-identical native == wasm "for any key type the callback returns". The element half was gated; the key half was not.

## The defect family

Every heap key traps, not just the tuple the issue named. Measured on 0.62.0, before/after, same programs:

| key | element | before | after |
|---|---|---|---|
| `(Int, String)` | heap | trap | ✅ agrees |
| `(Int, Int)` | heap | trap | ✅ agrees |
| `(Int, Int)` | scalar | trap | ✅ agrees |
| `List[Int]` | heap | trap | ✅ agrees |
| `Option[Int]` | heap | trap | ✅ agrees |
| `Bool` | heap | agreed | ✅ agrees |

The issue's own repro now builds on the **structural leg** at 3,358 B — down from the incumbent's 11,157 B — and prints the answer.

## The fix

**One source of truth for the key domain.** `lower_list_sort_by` restated a scalar key list (`INT | FLOAT | STR | BOOL`) that had drifted from what the emitter could actually order. It no longer restates anything: it hands the key type to `emit_val_cmp` — the *same* type-directed comparator `list.sort` uses for a compound **element** — so a key is orderable exactly when an element of that type is, and a key shape with no order walls `list-cmp-elem:<ty>` out of that one place. `emit_merge_sort_by` was already generic over the key type; only the gate in front of it was not.

**The cached keys are released.** The key buffers are raw blocks and the winner was freed flat, so every compound key leaked. Measured over 200 sorts of 2,000 elements: **146 MB → 10.7 MB** peak RSS. The winner takes the typed `$drop_list` for its key type now; the loser holds only stale duplicates of the same handles and stays flat.

**The incumbent refuses instead of mis-rendering.** A heap key routes to `list.sort_by_x` (an unlinked render wall) — the same refusal it already makes for a non-scalar `unique_by` key (C-147). The reported failure was not that this leg lacked a compound-key route; it was that it rendered one anyway. That is now a build-time refusal with no artifact, so the trapping `verified` build is gone rather than merely unreachable.

String, Float, Int and Bool keys are untouched on all three legs (verified directly).

## Evidence

- `spec/wasm_cross/sort_by_compound_key.almd` (new, `// @contract: C-053`) — `(Int,String)`, `(Int,Int)`, `(String,Int)`, `(Bool,Int)` with an `i64::MIN` field, `List[Int]` and `Option[Int]` keys over both heap and scalar elements; equal-key stability; the empty and single-element lists; and the 200-round release check. Byte-identical native == wasm.
- `tests/sort_by_compound_key_test.rs` (new, 4 tests) — the three agreement checks also assert the wasm build took the **structural leg** (an answer that agrees only because the build silently fell back is the state this issue was), plus the kill-check that forcing `ALMIDE_WASM_INCUMBENT=1` refuses at build time, names `list.sort_by_x`, and leaves no artifact.
- C-053's statement carries the `EXTENDED (#2154, 0.62.1)` clause; ledger indexes regenerated, stamped counts untouched.

Gates run locally: `check-contracts` OK (348 contracts, 705 fixtures, links symmetric), `check-als-pin` OK, `check-fmt-gate` OK, the 3-way interp oracle over all of `spec/wasm_cross/` OK, `almide test spec/lang/` 213/213, `almide test spec/stdlib/` 162/162.

## Not in scope

The structural leg's `list.group_by` cost (#2156) and the incumbent's unlinked `group_by` (#2155) are the other half of the word-count shape and are untouched here. A record key is still rejected at check time on both targets — but E030's hint says "lists/tuples/**records** of those", which a named record is not; filed as #2167 — along with C-053's own claim of variant ordering, which is refused the same way.

https://claude.ai/code/session_01Pi5SisFrjpvYtzoM8sZzV4


O6lvl4:
## Why "one domain, two positions" and not a widened key list

The change above deletes a key-type list rather than extending it. That is the load-bearing decision, so I checked it against the reference implementations in `../almide-references` before committing to it. **No implementation in the set defines a separate, narrower "sortable key" domain** — and the two that actually ship a key-sort make the key domain identical to the element domain *by construction*, not by coincidence:

- **Rust** — `sort` is `where T: Ord` (`library/alloc/src/slice.rs:131`), `sort_by_key` is `where K: Ord` (`:247`): same trait, same `stable_sort` driver, same `lt`. Decisively, `sort_by_cached_key` (`:312`) builds a `Vec<(K, u32)>` and calls `indices.sort_unstable()` — **the key is sorted as an ordinary element, inside a tuple.** It could not typecheck if the two domains differed. Almide's self-hosted `sort_by` is explicitly modelled on `sort_by_cached_key` (`stdlib/list_sortby.almd`'s header says so), so this is the direct ancestor of the code being changed.
- **CPython** — `list.sort` always sorts `lo.keys`; with no `key=`, *the elements are installed as the keys* (`Objects/listobject.c:2975-3003`). One array, one domain; `key=` only decides what goes in it. There is no key type test that can fail — `safe_object_compare` is a bare `PyObject_RichCompareBool(v, w, Py_LT)` (`:2765`), and the type dispatch above it is pure specialisation with a fallback.
- **Go, Zig, Gleam** have no key-sort at all — they take a comparator, so the key never appears in a signature and cannot carry a bound. (Go's *element* domain is the narrower one: `cmp.Ordered` is a closed union excluding structs and arrays, `src/cmp/cmp.go:18`.) **Koka** resolves one structural `cmp` with lexicographic tuple instances (`lib/std/core/tuple.kk:60-86`) and has no second, restricted notion.

The corollary the references also carry, and which this PR depends on: the composite type used for a multi-field key must itself be in the element domain with a lexicographic order. Rust implements `Ord` for tuples of arity 1–12 (`library/core/src/tuple.rs:111`, `tuple_impls!` at `:238`), which is exactly what makes `sort_by_key(|x| (a, b))` work; CPython's `tuple_richcompare` (`Objects/tupleobject.c:728`) is the same shape, which is why `key=lambda x: (a, b)` is *the* Python idiom. Almide already had that half — `list.sort` has ordered compound elements since C-053. Only the key position was gated, which is the defect.

So the fix is the one with precedent, and a widened key list would have been the one without: it would have re-created the same two-copies-of-one-truth that produced this bug.

(Checkout note for anyone re-running this: `rust/`, `cpython/` and `go/` in `../almide-references` are sparse, so the cited files need `git show HEAD:<path>` from inside each repo; `zig/` and `koka/` are on disk. Gleam's stdlib is not vendored there.)

O6lvl4:
## Follow-up commit: the manifests, and one thing the ledger found

`df44dfb0b` registers the new fixture with the oracle-pinned goldens and re-pins one allocation row. Three of the four are bookkeeping; the fourth is a finding.

**The allocation ledger caught a second leak, already in the corpus.** `spec/wasm_cross/list_total_order.almd` dropped from 72,224 to 72,208 B without me touching it. The cause is line 67 of that fixture — `list.sort_by([3, 1, 20], (x) => int.to_string(x))`, a **String** key. String is a handle, so the same flat free that leaked compound keys had been leaking the cached String keys on the structural leg since the sort was written. Nothing pointed at it because a String key never trapped; it just never gave the blocks back. The typed release in this PR covers both, which is what a type-directed release is for. Ratified with `ALMIDE_UPDATE_ALLOC=1`, one row, downward.

**The fixture sits outside the oracle-refereed sweeps, on purpose.** `spec/wasm_cross/sort_by_compound_key.almd` is in `scripts/lib/run-oracle-exclusions.txt`: the run manifest's reference is the a877d2138 oracle's wasm leg, and on this fixture that leg *traps on the very subject the fixture pins*. Its recorded row would be the defect, not an answer — the `closure_captured_list_mutation` situation exactly. The register row states the cost: holding it out takes it off the exercised-surface, allocation and size sweeps, and that costs nothing here because `list.sort_by` stays in the manifest through `sort_by_call_count`, `sort_by_str_key_heap` and `list_total_order`. (The register I want is not the *stale* one: that keeps the row in the manifest, and a row needs an answer the oracle actually produced. Fabricating a refusal row into a generated file is not a thing I'm willing to do.) It re-enters when the port catches up.

The check and AST manifests take ordinary rows — `almide check --json` is clean (empty stdout, exit 0, the `e3b0c442…` every clean fixture carries) and the AST hash is recomputed and verified against a known row's recipe first.

Full `commissioned-wasm-gates` job reproduced locally and green: carried gates across the six crates, the release-only root tests, the `ALMIDE_RC_TRAP_DOUBLE_FREE=1` corpus burn-up + gauntlet.

PR #2171, opened 2026-09-13
Complete and promote every llms.txt fence to almide check, and have the surface gate report checked/total per file

### LINKED ISSUE #2144: llms.txt has 0 of 8 code fences under almide check while CHEATSHEET has 12: the promotion arc moved in one file only
Surveyed 2026-09-13 against `develop` @ 51b9e80d7.

## The count

`scripts/check-llm-surface.sh` scans both files named in `CLAUDE.md` as the LLM-facing surface (`FILES=("$ROOT/llms.txt" "$ROOT/docs/CHEATSHEET.md")`) and runs `almide check` on every fence labelled ```` ```almide check ````. Current state:

| file | ```` ```almide check ```` | plain ```` ```almide ```` | checked |
|---|---:|---:|---:|
| `docs/CHEATSHEET.md` | 12 | 9 | 57 % |
| **`llms.txt`** | **0** | **8** | **0 %** |

All eight `llms.txt` fences are fragments — none carries a `fn main`, so none is promotable as written. The gate is behaving exactly as designed: plain means fragment, fragment means unchecked.

## Why it is worth a row anyway

`CLAUDE.md` states the rule as *"anything that appears there is a promise"*, naming both files, and #1483 built the promotion arc to make the promise checkable one fence at a time. That arc has moved **twelve fences in one file and zero in the other**. The file with zero is the one whose entire reason to exist is being read by a model — a cheatsheet is also read by people who can notice a broken example, `llms.txt` is not.

The eight fragments are small (2–11 lines). Nothing suggests they are wrong; nothing has ever checked that they are right.

## Ask

Two things, neither large:

1. **Complete and promote the eight**, the same way CHEATSHEET's twelve were — wrap each fragment in the minimum that makes it a program, promote the marker, let CI run `almide check` on it. If a fragment genuinely cannot be completed (it demonstrates a deliberate ✗-example, or a form that only appears inside a larger construct), leave it plain and say which in a comment, so the remainder is a real zero rather than an unexamined one.
2. **Make the imbalance visible.** `check-llm-surface.sh` already walks both files and already knows both counts. Having it print `checked/total` per file turns "the arc moved in one file only" from something someone has to go count into something the gate says every run. The abstain-class gate is the model: `abstain-classes OK — corpus 704, voting 682 (96.9%), abstaining 22 (HEAP_BOUNDARY=11, BRIDGEABLE=11); near-horizon target 693 (98.4%)`. **It reports the size of its own hole.**

## What this is not

Not about the gate's correctness — it scans the right files and applies the right rule. Not about the perf or size corpora (#2141, #2142, #2143). This is coverage of the documented surface only.

O6lvl4:
Closes #2144.

**1. The eight fences.** Every plain ```` ```almide ```` fence in `llms.txt` is now a complete program under ```` ```almide check ````, so CI runs `almide check` on each. Each was wrapped in the minimum that makes it checkable while keeping what it teaches — the ✗/✓ pair stays a pair (two named fns), the `!` / `?` / `??` trio stays three lines with the same comments, the error-handling stances stay three stances. None was left plain: none of the eight is a deliberate ✗-example or a form that only exists inside a larger construct.

Two fragments were quietly wrong as written and are now honest: the head/tail recursion used `list.first(xs)!` in a pure fn that does not return Result/Option (E0xx at check), now `?? 0`; the operator trio named `option.get` / `result.get` / `parse_int`, which are not stdlib, now `value.field` / `value.as_int` / `int.parse`.

```
llm-surface: llms.txt checked 8/8 fence(s) (100%), 0 plain
llm-surface: CHEATSHEET.md checked 12/21 fence(s) (57%), 9 plain
```

**2. The gate reports the size of its own hole.** `scripts/check-llm-surface.sh` now prints `checked/total` per file (labeled, all, percent, plain) — the lines above — so "the arc moved in one file only" is something the gate says every run instead of something someone has to count. The allowlist and the failure path are untouched.

`docs-gen --check` clean.


O6lvl4:

Issue #2160, opened 2026-09-13
wasm export mode: a pub fn that matches a locally produced variant and returns a String walls (remainder of #2110)


O6lvl4:
Follow-up to #2110. The six walls from that report are gone and the module now builds as a library. One wall remains, and its shape is now clear enough to split out.

## Symptom

In export mode (`ALMIDE_WASM_INCUMBENT=1`, a main-less library), a `pub fn` walls when

- it `match`es a variant that was **produced inside that function**, and
- the result of the `match` is a **heap value (String)**

```
wall: exported `pub fn f` is outside the MIR-lowering subset (the wasm module
must carry its export): heap-result `match` outside the executable subset
cannot be faithfully returned in this brick (would move out an empty deferred
heap value)
```

## Minimal repro (3 lines)

```almide
type T = | A(String) | B

local fn make(s: String) -> T = if s == "" then B else A(s)

pub fn f(s: String) -> String = match make(s) { A(x) => x, B => "b" }
```

```
ALMIDE_WASM_INCUMBENT=1 almide build src/mod.almd --target wasm
```

## What triggers it and what doesn't

| shape | result |
| --- | --- |
| `match make(s)` → String | **wall** |
| `let t = make(s)` then `match t` → String | **wall** (a let-binding does not help) |
| `match t` where t is a **parameter** → String | builds |
| `match make(s)` → Int | builds |
| `match n { 0 => "a", _ => "b" }` (matching an Int) | builds |
| `match o { some(_) => "a", none => "b" }` (matching an Option) | builds |
| payload-less variant taken as a parameter, matched → String | builds |

So it needs both halves: a **locally produced variant** and a **heap result**. Either one alone is fine.

Native (`almide check`, native build) accepts it. Only wasm export mode walls.

## Versions

- v0.62.0 (released) — wall
- develop `112df3ed8` (2026-09-13) — wall

## Why it matters

We run the judgement half of a Cloudflare Workers deployment as an Almide wasm module. Exporting `pub fn`s would let the JS host call it directly. Because of this wall the boundary is still `main` + stdin/stdout, which costs a hand-written 95-line WASI shim on the host side — and that shim is where #2114 became a production incident in the first place.

The function that hits it classifies a GraphQL response:

```almide
pub fn parse_page(body: String) -> String = match parse_board(body) {
  RateLimited   => "rate_limited",
  BoardNotFound => "not_found",
  FetchError(e) => e,
  Page(page)    => json.stringify(...),
}
```

[...]

PR #2173, opened 2026-09-13
Hand a walled test-free main file to the structural leg in almide test --target wasm, as the build already does

### LINKED ISSUE #2121: `build --target wasm` reports `Built … verified` for an entry whose dependency walled; the artifact OOMs at runtime while `test --target wasm` correctly reports the wall
`almide build --target wasm` reports success for an entry whose dependency walled, and the artifact fails at runtime with `out of memory`. The same entry, run through `almide test --target wasm`, correctly reports the wall and skips.

#782 retired the v0 emitter so that "a wall is an honest error instead of a silent fallback". On this path the build is not honest: it prints `Built … verified` and hands back a module that cannot run.

almide 0.62.0, macOS arm64.

## Reproduction

Public repo, three commits, nothing else needed: https://github.com/O6lvl4/emet

```
git clone https://github.com/O6lvl4/emet && cd emet
```

**1. The build says it worked.**

```
$ almide build src/worker.almd --target wasm -o /tmp/e.wasm
Built /tmp/e.wasm (67042 bytes, structural leg, verified — wasm-opt skipped; pass --wasm-opt for a smaller, non-verified build)
```

**2. The test runner, on the same entry and the same target, finds the wall.**

```
$ almide test src/worker.almd --target wasm --verbose
SKIP src/worker.almd (v1 wall — no verified wasm rendering)
0 passed, 0 failed, 1 skipped (of 1 files)
```

**3. Running it fails with an allocator error rather than the wall.**

```
$ almide run src/worker.almd --target wasm < examples/worker-in.json
Error: out of memory

$ almide run src/worker.almd < examples/worker-in.json      # native: correct output
{
  "tool": "emet",
  "version": "0.1.0",
  "total": 6,
  …
```

The input is 842 bytes and the largest string in it is a 40-character HTML
fragment, so this is not a real allocation. The native run of the same entry on
the same input is correct.

**4. Building the dependency directly does surface the wall.**

```
$ almide build src/judge.almd --target wasm -o /tmp/j.wasm
error: this program shape is not yet supported by the verified wasm renderer

  exported `pub fn verdict` is outside the MIR-lowering subset (the wasm module
  must carry its export): heap-result `match` outside the executable subset
  cannot be faithfully returned in this brick (would move out an empty deferred
  heap value)
```

So the information exists in the pipeline. `build` on the entry that *imports*
`judge` does not consult it.

## The two things I think are separable

**(a) The build/test divergence.** Whatever `test --target wasm` checks before it
decides to skip is not being checked by `build --target wasm`. If it were, step 1
would have printed step 4's diagnostic. This is the part that cost the most time
here: the artifact existed, was labelled `verified`, and only failed at the point
where the failure carried no information.

**(b) The wall shape itself.** `judge.verdict` returns a variant whose cases carry
`List[String]`, from a `match` on a payload-free variant, and is public so the
module has to export it. Related to the closed family #1547 / #1564 / #1581 /
#1583 / #1676.

## Minimisation attempts, and what they rule out

I could not reduce (b) to a small file — which is itself a clue, so here is what
does **not** reproduce it. Every one of these builds and runs on wasm:

- a list-carrying variant returned from a flat `match`
- the same from a *nested* `match`, with block-bodied arms containing `let`s
- a nested closure capture in the returned payload
  (`list.filter(figs, (f) => list.any(pages, (p) => …))`)
- a hand-copied single-file version of `verdict` with all five variant cases,
  the `local type Early: Eq` scrutinee, and the `best_source` var-loop helper
- the same split across two modules in a package, with the types in a third

What the real `judge.almd` still has that those copies lost: calls into a
*sibling* module from inside the walling function (`tokens.figures`,
`claim.ok_page`), the record parameter it reads fields off (`c.text`,
`c.citations`), and `test` blocks in the same file that call the public function.

One shape I did confirm independently: returning `Option[Verdict]` — an **empty**
option of a heap type — walls on its own, with the same "would move out an empty
deferred heap value" wording. Replacing it with a payload-free
`local type Early = NoSource | NoFigures | MustRead` cleared that wall but not the
one on `verdict`.

## Why this one matters beyond the shape

The program is a CLI whose judging is a pure function on purpose, so that the same
`verdict` can run behind a Cloudflare Worker where the host's `fetch` reads the
sources and Almide only decides. That is a fair chunk of what native==wasm buys,
and today the failure mode for it is a module that builds, claims to be verified,
and OOMs.

## Unrelated, same session, in case it is useful

A run over 115 live URLs died in:

```
thread 'main' panicked at src/main.rs:1223:39:
end byte index 681 is not a char boundary; it is inside '\u{fffd}' (bytes 679..682 of string)
```

on a page with invalid UTF-8, and did not reproduce on the next run of the same
input. `string.index_of`, `string.slice` and `string.drop` were checked against
multibyte input and agree on char indices, so the offset does not look like it
originates in user code. I do not have a deterministic reproduction; filing
separately would need one. Happy to open it if the line number is enough to go on.

O6lvl4:
Closes #2121.

On `develop` the artifact half of the report is already fixed — `almide build --target wasm` ships `src/worker.almd` on the structural leg and it runs byte-identical to native (129 lines, checked against the repro repo). What remained was the divergence: `almide test --target wasm` on the same entry still said `SKIP (v1 wall — no verified wasm rendering)`, because the test runner asked only the incumbent and never took the reverse handover the build takes (`render_wasm_module_routed`). A program that builds and runs reported SKIP — and a structural-leg defect in it could hide behind that SKIP.

The runner now mirrors the build's route: when the incumbent walls a `main`-carrying, export-free entry **with no test blocks**, it lowers on the structural leg, validates, audits the host ops against the p1 surface wasmtime serves, applies the same `to_wasi` rewrite the build ships, and runs it. `ALMIDE_WALL_REASON=1` names each stage that declined, or prints `[handover] … structural leg took the test`.

A file that declares tests keeps the honest skip — the structural leg has no test mode, so running its main alone would report a pass over nothing (or a FAIL over a main that wanted stdin, which is exactly what the repro's `worker.almd` with its 52 tests did on a first cut). The skip now says so:

```
SKIP src/worker.almd (v1 wall — no verified wasm rendering (1 test block(s) route to native; the structural leg has no test mode))
```

and the default harness reports `1 via native fallback` for it with no wasm TRAP.

The wall shape itself (a pub fn matching a locally produced variant and returning a String) is the #2160 remainder and is not touched here.

Tests: `tests/test_wasm_structural_handover_test.rs` — a walled sibling under a test-free main runs (`1 passed`, no SKIP, build precondition asserted); a main with a test block keeps the skip and names the reason. Existing wasm runner tests green; `almide test` 4082/440 green.


O6lvl4:

PR #2174, opened 2026-09-13
Borrow a list forwarded by a wrapper into another module instead of cloning it at every call

### LINKED ISSUE #2164: native: a one-line wrapper forwarding a List to another module's fn takes it by value and clones per call — quadratic on a single linear pass
## Summary

A **one-line forwarding wrapper** that passes a `List[T]` to a function in **another module** makes the
callee take the list **by value and clone it at every call**. The same wrapper forwarding to a function in
the *same* file is borrowed, and calling the other module's function *directly* is borrowed too.

Only the cross-module hop through a wrapper regresses. A read-only accessor called once per element then
costs O(n) per call, so a single linear pass over a list becomes quadratic.

This is the same defect class as #2040 (mutually recursive params by value) and #647 / #2070, but a
different trigger, and it is present in **0.62.0**. `--release` does not change it.

## Repro

`src/other.almd`

```almide
type Tok = { kind: String, text: String, line: Int }

fn get_far(ts: List[Tok], i: Int) -> Tok = {
  var out = Tok { kind: "", text: "", line: 0 }
  match list.get(ts, i) {
    some(t) => { out = t },
    none => (),
  }
  out
}
```

`src/main.almd`

```almide
import process
import self.other

// the same accessor, in this file
fn get_here(ts: List[other.Tok], i: Int) -> other.Tok = {
  var out = other.Tok { kind: "", text: "", line: 0 }
  match list.get(ts, i) {
    some(t) => { out = t },
    none => (),
  }
  out
}

fn via_far(ts: List[other.Tok], i: Int) -> other.Tok = other.get_far(ts, i)   // <-- SLOW
fn via_here(ts: List[other.Tok], i: Int) -> other.Tok = get_here(ts, i)       // fast

fn scan(ts: List[other.Tok], n: Int, mode: String) -> Int = {
  var acc = 0
  var k = 0
  while k < n {
    var t = other.Tok { kind: "", text: "", line: 0 }
    if mode == "far-wrap" then { t = via_far(ts, k) }
    else if mode == "here-wrap" then { t = via_here(ts, k) }
    else if mode == "far-direct" then { t = other.get_far(ts, k) }
    else { t = get_here(ts, k) }
    if t.kind == "id" then { acc = acc + 1 } else ()
    k = k + 1
  }
  acc
}

fn build(n: Int) -> List[other.Tok] = {
  var xs: List[other.Tok] = []
  var i = 0
  while i < n {
    list.push(xs, other.Tok { kind: "id", text: "t" + int.to_string(i), line: i })
    i = i + 1
  }
  xs
}

effect fn main() -> Unit = {
  let args = process.args()
  let n = int.parse(list.get(args, 1) ?? "1000") ?? 1000
  let mode = list.get(args, 2) ?? "far-wrap"
  let ts = build(n)
  println(mode + " " + int.to_string(scan(ts, n, mode)))
}
```

```
almide build
./repro 10000 far-wrap
./repro 10000 here-wrap
```

## Measured (almide 0.62.0, macOS 14.5 arm64, default build)

**The call count is identical in all four rows. Only the shape of the hop differs.**

| shape | n=10 000 | n=20 000 | n=40 000 |
|---|---|---|---|
| `via_far` — wrapper → **other module** | **3.64 s** | **13.30 s** | **54.55 s** |
| `via_here` — wrapper → same file | 0.00 s | 0.00 s | 0.00 s |
| `other.get_far(ts, k)` — called directly | 0.00 s | 0.00 s | 0.00 s |
| `get_here(ts, k)` — called directly | 0.00 s | 0.00 s | 0.00 s |

The slow row is ×4 per doubling of n = quadratic in list length for a **single linear pass**.
`--release` measures the same on the same repro (3.29 s / 12.06 s at n=10 000 / 20 000), so this is a
codegen decision, not an optimization level.

## Why it matters

Found while profiling [teastia](https://github.com/Aid-On/teastia), a static analyser whose token scanners
all go through exactly this shape:

```almide
// src/tools.almd
fn tok_at(ts: List[lexer.Tok], i: Int) -> lexer.Tok = tokens.tok_at(ts, i)   // <-- the wrapper
```

Consequences in that real program:

- one scan of a 10 560-token file: **3.6 s** (the lexer that produced those tokens takes 0.023 s)
- `tools` runs 5 such scans → 18 s for a 33 KB source file
- the layer above re-scans per tool, so a file with N tool definitions is **O(N³)**:
  10 tools 0.82 s → 20 tools 5.63 s → 40 tools 41.6 s → 80 tools **321.9 s**
- the same 80 tools split across 80 files: **0.39 s**

Isolation that led here (same binary, same call count each time):

| | |
|---|---|
| the lexer alone on the same 33 KB file | 0.023 s |
| a bare loop calling the wrapper once per token | 3.62 s |
| the same loop, same call count, passing a **1-element** list | 0.02 s |
| the same loop with `list.get(ts, k)` written inline | 0.02 s |

So the cost tracks the **size of the list handed to the wrapper**, not the number of calls.

## Expected

A wrapper that only forwards a read-only `List[T]` should borrow it, the same as the direct call and the
same-file wrapper already do — presumably `pub fn via_far(ts: &[Tok], i: i64) -> Tok`.

O6lvl4:
Closes #2164.

A one-line wrapper forwarding a `List[T]` to a fn in another module took the list **by value** and cloned it at every call; the same wrapper into its own file, and the other module's fn called directly, were both borrowed. A read-only accessor called once per element went O(n) per call — quadratic over one linear pass (3.64 s / 13.30 s / 54.55 s at n = 10k / 20k / 40k in the report).

Cause: the ownership walk (`check_needs_ownership_call_user_named`) consulted the borrow snapshot only for a `CallTarget::Named` callee. A `CallTarget::Module` callee — a sibling user module — fell through to the pessimistic fallback, where any use of the parameter as an argument means "needs ownership". The snapshot already holds the module fn's signature under `module::func`, the key the `RefMut` walk (`check_needs_refmut_call`) reads; the ownership walk now reads it too, including the pending-fn optimism for forward references. Bundled stdlib modules never reach this branch (answered earlier by `check_needs_ownership_call_bundled_module`).

Emitted Rust for the report's repro, before → after:

```
pub fn via_far(ts: Vec<almide_rt_other_Tok>, i: i64)   →   pub fn via_far(ts: &[almide_rt_other_Tok], i: i64)
```

Measured on the repro: `far-wrap` n=10 000 8.11 s → 0.00 s; n=40 000 0.01 s — the same as `here-wrap`.

Tests: `tests/cross_module_wrapper_borrow_test.rs` (emit-shape over a two-module package: the cross-module wrapper takes `&[`, not `Vec<`, and forwards without a clone). The ten borrow/ownership test files stay green; `almide test` 4082/440 green.


O6lvl4:

PR #2175, opened 2026-09-13
Wall a heap sort_by key on the incumbent wasm leg instead of emitting a call_indirect that traps

### LINKED ISSUE #2154: A tuple sort key builds on the incumbent wasm leg as "verified" and traps at run time with an indirect call type mismatch, while native prints the answer

O6lvl4:
Closes #2154. (#2155 — lowering the tuple key so the idiom builds — stays open; this removes the divergence, not the wall.)

`list.sort_by(map.entries(counts), ((w, c)) => (0 - c, w))` routed to the incumbent (the structural leg walls `list-sort-by-key:Tuple`), which accepted it, labelled the module **verified**, and emitted a `call_indirect` whose type did not match: `sort_by_rc` takes `f: (Int) -> Int`, a scalar key fn, and a tuple key returns a heap handle. wasmtime trapped (`indirect call type mismatch` in `__sbr_init`); native printed the answer. `almide check --target wasm` said `No errors found`.

The incumbent's `sort_by` router now sends a heap key that is not a String (tuple, record, list) to the `_x` twin — the same honest-wall spelling `list.sort_x` already uses — so the link step reports it as unlinked instead of rendering it:

```
$ almide build p.almd --target wasm
wall: unlinked stdlib/runtime call(s) with no wasm definition: list.sort_by_x — …
wall (structural leg, the default): list-sort-by-key:Tuple(3)
error[E082]: both wasm legs refused this program — …
$ almide check p.almd --target wasm
error[E082]: both wasm legs refused this program — …
```

Both extra faults in the report follow from that: no artifact claims to be verified, and `check --target wasm` reports the wall at check time. Native is untouched (`a 3 / b 2 / c 1`). String and Float keys keep their twins.

Tests: `tests/wasm_sort_by_heap_key_walls_test.rs` (native still answers; the wasm build refuses and names `sort_by`, no `Built` line; `check --target wasm` is non-zero). `wasm_both_legs_e082_test` green; `almide test` 4082/440 green.


O6lvl4:
Superseded by #2168 (merged 2026-09-13, closed #2154). After rebasing onto develop the only remaining diff is the test file, and its build-must-wall assertion no longer holds: the structural leg now orders a tuple key, so `almide build --target wasm` succeeds and `tests/sort_by_compound_key_test.rs` covers that on both legs. The incumbent-side `_x` refusal this PR added is already in develop from #2168 as well.

Issue #2172, opened 2026-09-13
A recursive type that derives Ord sorts on native and walls the structural leg: ordering has no out-of-line $cmp_<ti>


O6lvl4:
Split out of #2167, which made records and variants orderable on the structural leg. One cell of that lattice stays walled: a RECURSIVE orderable type.

```almide
type Tree: Ord = { v: Int, kids: List[Tree] }

fn main() -> Unit = {
  let ts = [Tree { v: 2, kids: [] }, Tree { v: 1, kids: [Tree { v: 9, kids: [] }] }]
  for t in ts |> list.sort { println("t ${t.v}") }
}
```

```
$ almide run tree.almd
t 1
t 2
$ almide run tree.almd --target wasm
wall (structural leg, the default): cmp-recursive-named:0
```

Native derives `Ord` for it, so it is legal Almide; the structural leg refuses. The refusal is deliberate and is the RIGHT failure — `emit_val_cmp` recurses INLINE, so without the guard #2167 added, the compiler would have blown its own stack on this input rather than emitted anything. A wall the user can read beats a compiler crash. But it is still a divergence: native answers, wasm will not build.

## Why it is not fixed in #2167

Equality already solved this exact cycle, and the solution is not a guard — it is an out-of-line helper. `emit_val_eq` emits `$eq_<ti>` as a real function (`Helper::NamedEq`, built by `build_one_eq_helper` in `display.rs`) and CALLS it, so a recursive type becomes a recursive call instead of infinite inlining. Ordering needs the same `$cmp_<ti>`.

The reason to do it separately is that the honest version is a REFACTOR, not a copy. `build_one_eq_helper` is ~60 lines of Emitter scaffolding plus its wiring across `work.rs`, `display.rs` and `assembly.rs`; adding a near-identical `build_one_cmp_helper` beside it would plant exactly the second-copy-of-one-truth that caused #2154 and #2167 in the first place. The builder should be parameterized over what it emits (eq | cmp) with one bodies map keyed by (kind, ti), and that is worth its own change.

## Worth having beyond the recursive cell

Out-of-line helpers also stop inlining a comparator per use site. A record with many fields, or a list-of-records sorted in three places, emits the whole field chain three times today. `$cmp_<ti>` is emitted once and called — smaller modules on exactly the shapes #2167 just enabled.

## Done when

- `cmp-recursive-named:` no longer exists as a wall reason
- the `Tree` program above prints identically native == wasm, with a fixture under C-053
[...]

Issue #1521, opened 2026-08-18
Silent-acceptance and diagnostic-scope backlog from the wave-2 fixture sweep


O6lvl4:
Remaining verified findings from the 2026-08-18 wave-2 diagnostic sweep (378 new fixture pairs; the top holes were fixed at 2199ded6d / 56cc44840: guard-Bool, duplicate top-lets, unknown named args, tuple-pattern shape+arity, uppercase assignment targets, fan nested-block capture, E057, E029 variant payloads, E004-coded ctor arity). Each row reproduces on develop.

## check-accepts / build-or-banner-dies
- [ ] `for x in 5` (non-iterable head) → COMPILER BUG banner at codegen; should be a checker error
- [ ] `3.value` (field access on a scalar) → COMPILER BUG banner; should be E013-family
- [ ] lambda-VALUE arity: `f(1)` where `f = (a, b) => …` passes check (arity checks are named-fn only); unify tolerates param-count mismatch
- [ ] `..=` retired range inside an INTERPOLATION segment bypasses E031 entirely (`"${list.len(1..=4)}"` runs)

## diagnostic scope gaps
- [ ] E052 deprecation not fired through pipe position (`"hi" |> yell`) — warn_if_deprecated is Call-callee only
- [ ] E014 duplicate match arms unseen inside a pipe-position lambda (direct call form fires)
- [ ] E030 orderable-element check skips record elements (`list.max(List[record])` silent)
- [ ] E024 sized-literal range: computed constants (`0 - 200`), branch-result positions, and Map VALUE positions all skip the range check (literal-only today)
- [ ] E016 fn-in-key: Named-record recursion gap (#1518) plus interpolated set-of-closures fully silent
- [ ] E029 unknown types in LAMBDA param annotations surface as phantom-type E001 instead
- [ ] unknown UFCS on a record (`p.zzcompute()`) → misleading E001 "expected P but got fn() -> Int" (should be E002/E013)
- [ ] code-less diagnostics to code: unused import / duplicate import / Result-interp debug warning / "type K is not hashable" / cascade operator errors ("operator '>' is not defined for Result…")
- [ ] `@deprecated` attribute parse errors silently swallowed (#1518 item 1)
- [ ] `o?.z` with a missing field on the inner record: code-less diagnostic (E013 gap)

## noted, by-design (no action)
- `effect fn main() -> Int` is deliberate (wrapper discards the Ok payload — mod.rs comment)
[...]

Issue #1483, opened 2026-08-16
Define the LLM-facing surface as the stable surface, and gate that they match


O6lvl4:
From the stability survey (`../almide-references/RESEARCH-stability-path.md`), gate 4. This is rust's feature gate translated to a language whose users read documentation as training data — no reference compiler has it, because none of them has this user.

## The problem

rustc's stable/unstable partition is enforced by the binary: stable rustc is compiled with `CFG_DISABLE_UNSTABLE_FEATURES` and literally cannot use a gated feature (`compiler/rustc_feature/src/lib.rs:62-85`). 257 unstable features, 211 accepted, 103 removed, all in an append-only ledger with a ~600-line linter demanding a tracking issue and a gate test per row.

Almide has no partition at all. Every feature that exists looks equally blessed. And for our users that is not an abstract risk: **`docs/CHEATSHEET.md` and `llms.txt` are the training and context surface.** What appears there is what models will write. Today nothing connects "documented in the cheatsheet" to "covered by a contract" to "intended to be stable".

## The Almide-shaped version of the gate

Instead of a compile-time feature flag, define:

> **The LLM-facing surface is the stable surface.** Anything in `docs/CHEATSHEET.md` or `llms.txt` is a promise. Anything not intended as a promise does not appear there.

and gate both directions:

1. **Every construct shown in the cheatsheet has executable coverage** — a `spec/` test, and for anything with cross-target meaning, a `spec/wasm_cross/` fixture naming a contract. A syntax we teach models to write and do not test is the worst cell in the matrix.
2. **Nothing marked experimental appears there.** This needs the experimental marker to exist, which is the other half of the work — today the closest thing is the `DECLARED_BUT_UNREAD` list added with E053, and `@dialect` epochs for things that have already changed.
3. **The stdlib names in `docs/stdlib/` match `almide compile --json`** — partially gated already by `tools/gen-stdlib-doc-index.py --check`; the gap is that the *cheatsheet* is unchecked prose.

## Why bother, given nobody else does

Because the failure mode is invisible and expensive. A model reads the cheatsheet, writes the construct, and the construct is under-tested — so the bug it hits is a compiler bug found by a user rather than by us, at the point of maximum cost. Every other project can rely on humans to notice a doc/reality gap; our primary reader cannot notice anything.

## Definition of done

1. A written statement of the rule, in `CLAUDE.md` next to the contract discipline.
[...]

PR #2177, opened 2026-09-13
Order a record or variant that derives Ord on the structural leg, and name the derive in the hint

### LINKED ISSUE #2167: A record or variant that derives Ord sorts on native and walls on both wasm legs, while C-053 claims they are byte-identical

O6lvl4:
Closes #2167. Follows #2168 (#2154) — same defect, one function over.

## What was wrong

```almide
type Inner: Ord = { n: Int, s: String }
type Shape: Ord = | Dot | Line{ len: Int } | Box{ w: Int, h: Int }
```

C-053 has claimed, since 0.24.0, that ordering is byte-identical native == wasm "for EVERY totally-ordered element type — Int, Bool, String, Tuple, nested List, Option, **and variants (tag order)**". Nothing executed the record or variant cell. Both were false:

| | native | wasm (before) | after |
|---|---|---|---|
| record element, `list.sort` | ✅ field declaration order | **both legs wall** | ✅ agrees |
| variant element, `list.sort` | ✅ case order, then payload | **both legs wall** | ✅ agrees |
| `list.min` / `list.max` | ✅ | **wall** | ✅ agrees |
| record or variant as a `sort_by` KEY | ✅ | **wall** | ✅ agrees |

`list-sort-elem:Named(0)` on the structural leg, `list.sort_x` on the incumbent. Native printed the answer; wasm would not build. A wall is honest, but a contract asserting the cell closed is not — and the cell had no fixture, which is exactly why the claim survived six versions.

## The cause is #2154's, verbatim

#2154 was `lower_list_sort_by` keeping its own list of orderable types while `emit_val_cmp`, the real comparator, had moved on. That list had **two more copies** in the same file:

- the `("sort", [xs])` arm — `INT | FLOAT | STR | BOOL | Tuple | List | Option`
- `lower_list_min_max` — the same seven

Both are deleted. `emit_val_cmp` is the single source of truth for what this language can order, in all three positions.

## What `emit_val_cmp` learned

Both shapes fall out of arms it already had, which is why this is small:

- **A record** is the `SliceTy::Tuple` arm's field chain, run over the def's fields. `pack_fields` assigns offsets in input order, so the def's field list *is* declaration order — the order native's derive compares in.
- **A variant** compares the tag, and only if the tags are equal dispatches to that case's payload chain. That is the general form of the `SliceTy::Option` arm's `none < some`.

The **derive stays the gate**, on both targets. A record or variant without `: Ord` is still E030 (#1521 — structural orderability of the fields is deliberately not enough, because the native monomorph needs `T: Ord`), and the test asserts that refusal on both targets rather than only the positive cells.

## One cell stays walled, declared not hidden

A **recursive** orderable type (`type Tree: Ord = { v: Int, kids: List[Tree] }`) builds natively and walls `cmp-recursive-named:` on the structural leg. `emit_val_cmp` recurses *inline*, so without that guard the compiler would exhaust its own stack on this input — a readable refusal beats a crash. Filed as **#2172**, with the reason it is not done here: equality escapes the same cycle with an out-of-line `$eq_<ti>`, and adding a near-identical `build_one_cmp_helper` beside `build_one_eq_helper` would plant exactly the second-copy-of-one-truth that caused #2154 and this issue. The builder wants parameterizing, and that is its own change.

## The hint (the part a reader actually hits)

It read:

```
hint: Ordering needs Int, Bool, String, Float, or lists/tuples/records of those.                  Map, Set, and function values have no order; Float inside a compound                  element has none either …
```

Two faults. The space runs are a multi-line source string joined with its indentation left in. And it told a reader whose record was just rejected that records order — without naming the `: Ord` derive, which is the one thing they had to write. That is the LLM-facing-surface rule (#1483) in its sharpest form: read the hint, write the same spelling, get the same error. It now names the derive, and `docs/diagnostics/E030.md` says the same thing with an example. The test asserts the hint names the derive, so the two cannot drift apart again.

## Evidence

- `spec/wasm_cross/ord_record_variant.almd` (new, C-053) — record elements incl. i64::MIN/MAX fields, nested records, unit variants, record-shaped and tuple-shaped variant cases with a fieldless case among them, min/max over both, record and variant keys, key-equal stability, empty and single-element lists. Byte-identical native == wasm, structural leg.
- `tests/ord_record_variant_test.rs` (new, 6 tests) — four positive cells that also assert the build took the **structural leg**, plus the two negative ones: no derive → E030 on both targets *with the derive named in the hint*, and the recursive type → a wall that names the cycle while native still builds.
- C-053's statement carries the `EXTENDED (#2167, 0.62.1)` clause, including the walled recursive cell and its issue.
- Record keys were also checked against #2168's release path under `ALMIDE_RC_TRAP_DOUBLE_FREE` (60 rounds × 300 elements): right answer, no trap, steady 16.5 MB.

Gates run locally: `check-contracts` OK (348 contracts, 706 fixtures), `check-als-pin` OK, `check-fmt-gate` OK, `check-llm-surface` OK, `explain_docs_test` OK, carried gates across the six commissioned crates OK (allocation ledger unmoved), `almide test spec/lang/` 213/213.

The incumbent wall ceiling rises 29 → 30 for the new fixture, documented at the bump: the incumbent refuses every non-scalar element (C-147) and nothing that emitted before stopped emitting. Measured directly — 676 emitted, 30 walled, 706 total.

https://claude.ai/code/session_01Pi5SisFrjpvYtzoM8sZzV4


O6lvl4:
The 3-way oracle found the same defect one leg further out, and it is worth naming.

`interp_cross_target_spec` went red with a `BOTH-BACKENDS-WRONG` banner on the new fixture:

```
ord_record_variant:
  interp:  exit=1 stderr="Error: internal: list.sort on non-comparable elements"
  native:  exit=0 stdout="rec 1 a\nrec 1 y\nrec 2 x\n…"
  wasm:    exit=0 stdout="rec 1 a\nrec 1 y\nrec 2 x\n…"
```

That banner is the gate working exactly as designed. The reference interpreter — the third judge, which shares no codegen pass with either backend — kept **its own** copy of "what this language can order": `Value::total_cmp_val` handled Int/Float/String/Bool/Tuple/List/Option and returned `None` for a record or variant. The same restated-domain shape as #2154 and #2167, in the crate whose job is to disagree when the other two are wrong.

So the fix is the fix, applied to the third position:

- `TotalOrder` replaces `Value::total_cmp_val`/`total_cmp_seq` and is now the interp's whole ordering domain, used by `list.sort`, `list.min`/`max` and `sort_by`'s key compare alike — a record by **field declaration order**, a variant by **case order then that case's payload**, recursing back through itself for tuples, lists, Options and nested nominals.
- It is a struct rather than a `Value` method because a variant needs one fact a `Value` cannot carry: the case's declaration index. Native's derive compares the discriminant, both wasm legs compare the stored `SUM_TAG` — each IS declaration order, and the interp reads the same fact from the program's type decls. The tag index is built in the **same walk** as `variant_ctors`, keyed by `(type, case)`: the registry that decides a case's identity is the one that decides where it sorts, so those two cannot drift either. Keying by case name alone would let two types that spell the same case swap tags — a wrong third vote, the one thing this crate must never cast.
- A record's fields are already rebuilt in declaration order by `fill_record_defaults`, so the positional walk IS the derive's order; `sort_orders_a_record_written_in_permuted_field_order_identically` pins that a permuted literal compares the same.
- `bytes.Endian` has no decl in `IrProgram`, so its tag lookup returns `None` and the sort abstains rather than guessing.

Verified: the interp now prints the fixture's 42 lines byte-identically to the native==wasm consensus; `list_total_order` and `sort_by_compound_key` (#2154's) still evaluate unchanged; `interp_abstain_ledger` is unmoved (no abstain gained or lost); `cargo test -p almide-interp` 118/118.

Evidence lands with it: five backend-free tests in `crates/almide-interp/tests/eval_test_parts/sort_and_mut.rs` covering the record element, the permuted literal, the variant element, min/max, and a record in the KEY position — milliseconds, so the nominal order is pinned without paying for the corpus build.

Issue #1050, opened 2026-08-02
Implicit unwrap of a user effect-fn result covers let/list/argument positions but not binary-operator operands; the `+` diagnostic never mentions `!`


O6lvl4:
almide 0.50.0, develop (2026-08-02).

## Repro

```almide
import random

effect fn helper() -> Int = random.int(0, 9)

effect fn main() -> Unit = {
  let b = helper()        // ✓ implicit unwrap → Int
  let d = [helper()]      // ✓ implicit unwrap in list-literal element position
  let e = helper() + 1    // ✗ error: operator '+' requires numeric, String, or List types
                          //          but got Result[Int, String] and Int
  println(int.to_string(b + e + list.len(d)))
}
```

## Actual

A user effect-fn result is implicitly unwrapped in let-binding, list-literal, and call-argument positions, but **not** as a binary-operator operand. `helper() + 1` fails; `[helper()]` succeeds. (Presumably `+`'s overload set means no unique expected type, so the coercion never fires.)

## Expected

Either the implicit unwrap covers operator operands too, or — if the coverage hole is intentional — the diagnostic must say so. The current hint ("Use + with numeric types, String, or List") is actively misleading: the fix is `helper()! + 1`, and nothing points there. A writer who has just seen `let b = helper()` work has no model that predicts the `+` failure.

Related surface: #950 (auto-unwrap in interpolation call-argument position, wasm wall) — same implicit-unwrap machinery, different hole.

O6lvl4:
[...]

PR #2202, opened 2026-09-14
Spell a borrowed loop binder's str view as a slice and report an effect call in operand position

### LINKED ISSUE #2194: native on develop: generated code calls .as_str() on a value already &str — E0658 str_as_str on stable rustc
## Summary

`a22cfb7f` fixes both shapes I reported (#2188, #2189) — confirmed, thank you. Building my program
against `develop` went from **3 errors to 1**, and the two `&String` vs `&str` shapes are gone.

The remaining one is new: the generated code calls `.as_str()` on a value that is already `&str`,
which is unstable on stable Rust.

```
error[E0658]: use of unstable library feature `str_as_str`
     --> <generated.rs>:24512:23
      |
24512 |     if almide_eq!(key.as_str(), "agents") { "agent".to_string() } else { "task".to_string() }
      |                       ^^^^^^
      |
      = note: see issue #130366 <https://github.com/rust-lang/rust/issues/130366>
```

`rustc 1.97.1 (8bab26f4f 2026-07-14)` (Homebrew), macOS 14.5 arm64, default `almide build`.

## Source

`key` is the binder of a `for` over a list literal of `String`, inside a larger function. The source
is in a private repo, so the relevant part inline (elided to the shape that matters):

```almide
fn object_graph(units: List[wfproject.WxUnit], u: wfproject.WxUnit, call: WeCall, id: String) -> WeGraph = {
  var out = ...
  var kind = "agent"
  if call.symbol == "crewai:Task" then {
    kind = "task"
  } else ()
  out = add_node(out, name, kind, u.file, call.line)
  var noted = false
  for key in [
    "tools",
    "mcp_servers",
    "mcpServers",
    "handoffs",
    "agent",
    "context",
    "agents",
    "tasks"
  ] {
    ...
    var child_kind = "agent"
    var relation = "uses"
    if key == "context" then {
      child_kind = "task"
      relation = "context"
    } else ()
    if key == "agents" then {          // <-- key.as_str() is emitted here
      child_kind = "agent"
      relation = "contains"
    } else ()
    if key == "tasks" then {
      child_kind = "task"
      relation = "contains"
    } else ()
    ...
  }
  out
}
```

Note the generated line is not the loop body verbatim — it reads like a hoisted
`if … { "agent" } else { "task" }` expression, so the `.as_str()` may be inserted while lifting that
comparison out rather than at the original site.

## What I could not do

I tried to reduce this to a standalone program and **failed** — both of these build cleanly on
`a22cfb7f`:

```almide
// (1) loop over a List[String] parameter
fn kind_of(keys: List[String]) -> String = {
  var out = "task"
  for key in keys {
    if key == "agents" then { out = "agent" } else ()
  }
  out
}
fn other(s: String) -> Int = string.len(s)

// (2) loop over a list literal, two comparisons
fn kind_of() -> String = {
  var out = "task"
  for key in ["tools", "handoffs", "agents", "tasks"] {
    if key == "handoffs" then { out = "handoff" } else ()
    if key == "agents" then { out = "agent" } else ()
  }
  out
}
fn other(s: String) -> Int = string.len(s)
```

So the trigger needs more surrounding context than I have isolated. Reporting the failing generated
line and the source shape rather than sitting on it — happy to bisect further or try more shapes if
that would help.

A cheap unconditional guard would be to never emit `.as_str()` on an expression already typed `&str`
(or to emit `&*x` instead), independent of finding the exact trigger.

### LINKED ISSUE #2196: An effect call in binary-operator operand position still propagates implicitly without a bang
Follow-up to #2182 / PR #2191. The four positions #2182 listed now report E041/E042, but one implicit-propagation position remains: an effect call used as a binary operator operand.

## Repro (released `almide 0.62.0` and the #2191 head behave the same)

```almide
effect fn f() -> Int = error("boom")

effect fn main() -> Unit = {
  let x = f() + 1
  println(int.to_string(x))
}
```

`almide check`: clean. `almide run`: `Error: boom`, exit 1 — the Result was unwrapped and propagated with no `!`.

## Where

The binop operand goes through the same auto-unwrap path the tail/arm/branch leaves used to; `crates/almide-frontend/src/lower/auto_try.rs` (bind arm around lines 746-771) still wraps it, and the checker's new leaf walk in `infer_implicit_tail.rs` does not visit operator operands.

## Fix

Report E041 (the bind form) / E042 (statement form) for an effect call in operand position of a binary or unary operator, with the `!` fix-it; add fixtures `tests/diagnostics/e041-binop-operand/` and `e042-binop-operand-statement/`; then delete the remaining wrapping arm in `auto_try.rs` (once no checker-accepted program can reach it, the file is dead and should go, closing the ADR-0008 loop that #2182 opened).

Found while auditing #2191 (2026-09-14).

O6lvl4:
Closes #2194, closes #2196.

**#2194** — `c.as_str()` (from #2190) assumed the borrowed loop binder is bound `&String`. The branch lift hoists a loop body's `if` into a synthesized fn whose param is that same binder as `&str`, and there `.as_str()` resolved to the unstable `str::as_str` (E0658 on stable rustc). The reduced shape is `for key in ["…"] { let kind = if key == "agents" then "agent" else "task"; … }` — it reaches rustc only on the build route (the v1 native render walls on the string-list literal, the v0 leg lifts the branch), which is why `emit` alone never showed it. The view is now `&c[..]`: `Index<RangeFull>` reaches `str` from `String`, `&String` and `&str` alike through auto-deref, so the spelling no longer depends on how the binder is bound. Pinned by a third case in `tests/native_borrowed_string_flows_test.rs` that runs the lifted shape on native and wasm.

**#2196** — the binary-operand strip (`operand_effect_unwrap`, #1050) handed the operator the payload without a report. It now queues the same E041 the other positions do, with the `!` fix-it (a plain call). Fixtures `e041-binop-operand` and `e041-binop-operand-statement` — the statement form reports E041 too: the sum is an Int the statement may drop, the call's Result is what propagated. The two #1050 spec fixtures that pinned the implicit form now spell `!` and still exercise the operand A-normalization in the lowering.

**The incumbent leg and the explicit `!`** — spelling the two #1050 fixtures with `!` exposed one shape the incumbent v1 leg did not lower: a BARE lifted body `int.parse(s)! + 1` (a block tail, an if branch, a ctor payload and a `let` all lowered already through the MIR's own ANF lift; the auto-inserted `Try` operand had been hoisted by the frontend's BinOp arm and never reached it as a bare root). The frontend now hoists such a root's `!` operands into binds (`{ let t = int.parse(s)!; t + 1 }`), the bind-position shape the MIR proves — measured over the eight sibling shapes on the incumbent, all build; `unwrap_in_callarg` (`ok(h(p)! + h(p+1)!)`) is untouched because the hoist stops at the root and never reaches into a ctor payload (a first attempt that hoisted at every BinOp walled exactly that fixture).

Not done: deleting the `auto_try.rs` BinOp arm. That arm A-normalizes the `Try` operand the two Result-tail shapes still produce, so it is not dead.

Commits keep the AST-manifest re-hash apart from the implementation (the ratchet-commit separation gate).


O6lvl4:
Superseded: the #2194 half landed as #2201 (merged 3f70bfb5e), and the #2196 half is in #2203, which now also carries the root-only hoist of bang operands that kept the err-test fixture on the wasm leg (the one gap this PR had over #2203). Closing to avoid landing the same change twice.

Issue #2254, opened 2026-09-16
An un-annotated Result binding inside a lambda body is not E041


O6lvl4:
Found while working #2250.

```almide
import fs

effect fn e() -> Unit = {
  let xs = [1, 2, 3] |> list.map((x) => {
    let t = fs.read_text("w.txt")
    t
  })
  println("${xs}")
}

effect fn main() -> Unit = e()!
```

`almide check` reports **no diagnostics** (release 0.62.0 and develop), and `almide run` prints a `List[Result[String, String]]` of three `err(...)`. The same `let t = fs.read_text("w.txt")` at fn-body level is E041 (ADR-0008: an un-annotated `let` of a `Result` must say what it does with failure). Moving the shape into a lambda body switches the rule off, so the two positions are judged differently for the same binding.

Expected: E041 on the `let` inside the lambda too (with the same `!` / `??` / `let _ =` hints), or a written rule for why a lambda body is exempt.

Issue #2253, opened 2026-09-16
check --json has no project-wide form: --json forces single-entry resolution


O6lvl4:
Split out of #2250 ("A project-wide `--json` would also remove the need to check files singly").

`almide check` with no FILE inside a package judges every entry under `src/` (#2165), but `--json` (and `--effects`) keep the single-entry resolution — `src/main.rs` says so: "`--json` and `--effects` are per-file reports with no multi-file shape". A harness that wants machine-readable diagnostics for a whole project therefore checks each file singly, and each single-file run re-infers every sibling module it imports.

What would help: `almide check --json` with no FILE emitting one JSON line per diagnostic across all package entries, each row carrying its `file` (the field already exists), in the same entry order the bare form uses. Exit code as today (0 = judged, `level` decides).

Issue #2095, opened 2026-09-10
A call inside `${}` gets a span with no end_col, so no fix-it is ever offered there


O6lvl4:
A call inside a string interpolation gets a span with a start and no end, so the
diagnostic can point at it but not measure it — and a fix-it needs a measured
range, so none is offered.

## Repro

```almide
fn main() -> Unit = {
  let a = list.nope([1])
  println("${list.nope([1])}")
}
```

`almide check --json`:

```
E002 primary=(2, 11, 20)  try_replace={line:2, col:11, end_col:20}  applicability=maybe-incorrect
E002 primary=(3, 14, null) try_replace=null                          applicability=unspecified
```

Same call, same suggestion available (`list.pop`). Outside the interpolation it
comes with a replacement range a fixer can apply. Inside it, `end_col` is absent,
so `with_suggested_fix` is never reached and the reader gets the hint without the
edit.

## Why it matters

`${...}` is the idiomatic way to print anything in Almide — `println(int.to_string(x))`
aside, most output goes through interpolation — so this is not a corner. Every
undefined-call inside a `${}` loses its fix-it, and `almide fix` cannot help
there at all.

## Scope

The start column is correct, so the sub-parse knows where the segment begins; it
is the end that is not carried back. Whatever re-bases the interpolation segment's
spans onto the outer line sets `col` and drops `end_col`.

## Related

The rendering side of this is handled: a span with no `end_col` now draws one
caret instead of guessing a width from the message text. That makes the
missing measurement honest, not present.

O6lvl4:
## Root cause: the interpolation span remap copies `line` and `col` and forgets `end_col`

`crates/almide-syntax/src/parser/primary.rs:565` — `parse_interpolation_expr_part` sub-lexes the `${…}` body and then rewrites the sub-lexer's coordinates into the parent source:

```rust
for t in &mut tokens {
    t.line = str_line;
    // col: sub-lexer 1-based → 0-based offset + parent string position
    t.col = str_col + 1 + expr_col_start + (t.col - 1);
}
```

`Token` carries `end_col` (`crates/almide-syntax/src/lexer.rs:70`) and this loop never touches it, so every token inside an interpolation keeps a parent-space `col` next to a sub-string-space `end_col` — a value SMALLER than its own start. Downstream reads that as no end at all.

## It costs more than the fix-it

Measured on develop, the same typo in both positions:

```
$ almide check b2095.almd
2 |   let a = list.nope([1])
[...]

PR #2255, opened 2026-09-16
Report interpolation holes at their own line and column and offer the E041 fix-it position only where the span closes the expression

### LINKED ISSUE #2250: check --json: suggestions carry line/col that do not match the file when single-file checking cannot resolve imports
## What happens

`almide check --json <file>` emits `suggestions` (and `try_replace`) whose `line`/`col` do not correspond to the file, when that file has imports the single-file check cannot resolve. The positions are unusable for a tool that applies fix-its.

Measured on develop `57ad24cb4` against one real 560-line module that imports 20 sibling modules (`import self.*`), checked as a single file:

```
$ almide check --json src/cli.almd
...
26 E041/E042 suggestions with replacement "!"
  → 19 of them name a column past the end of that line
  →  7 land inside a comment or a string literal
  →  0 land at the end of the expression the diagnostic is about
```

Concrete rows (line content is what is actually at that line in the file):

```
line 230 col 103   but the line is 21 chars:  `    m = "--osv-query"`
line 243 col  65   but the line is 16 chars:  `      ok(r) => {`
line 271 col  46   but the line is empty
line 433 col 128   but the line is 39 chars:  `    let i = string.index_of(m, q) ?? -1`
line 451 col  26   lands mid-expression:      `let inner = m.slice(4!, m.len())`
line 473 col  38   lands inside a comment:    `// …（<!animatetransform> / …`
```

Checking the same project **as a project** (`almide check` with no FILE) reports the same diagnostics at plausible lines, so the position bookkeeping is specific to the single-file path.

## Why it matters

`--json` is documented as "Output diagnostics as JSON (one per line)" and each row carries `suggestions[].{line,col,end_col,replacement}` plus `applicability`, which reads as a machine-applicable fix-it contract. A harness that trusts it corrupts source: applying the 26 suggestions above inserts `!` into comments, into string literals, and in one case into the middle of a multi-byte character when the column is treated as bytes.

`almide fix` reports these as "need manual application", so it does not hit the bug — but it also means the JSON is the only machine path, and it is the one that is wrong.

## What would help

Either make the single-file path report positions that correspond to the file, or suppress `suggestions` when the file could not be fully resolved (the run already emits `E002 undefined function 'cliargs.parse'` etc. for the unresolved imports, so the condition is known at that point). A project-wide `--json` would also remove the need to check files singly.

## Column unit (for the record)

Where positions ARE correct, `col` is a 1-based **character** count, not bytes: a 66-character line containing CJK (98 bytes, display width 82) reports `try_replace.col = 67` for end-of-line insertion. Worth stating in the docs — a harness that assumes bytes corrupts UTF-8.

## Version

develop `57ad24cb4` (233 commits past v0.62.0), built from source, macOS arm64.

O6lvl4:
Closes #2250

## Two mechanisms behind the wrong positions

1. **A `${…}` hole inside a heredoc** (`parse_interpolation_expr_part`): every sub-lexed token was stamped with the string's FIRST line and a column counted over the whole decoded template — a heredoc has its blank first line skipped and its common indent stripped, so a call on the heredoc's third line landed on the `"""` line at a column past its end (the "19 of 26 name a column past the end of that line" rows). Fix: the hole's text is verbatim in the literal's source (`Token.raw`), so `locate_hole` finds it there, from the previous hole on, and counts lines; a hole that itself spans lines shifts its continuation tokens by the indent the decoder stripped (`continuation_shift`).
2. **A call whose `)` sits on a later line**: a `Span` has no end line, so `parse_postfix_call` carries the span of `(` alone, and the E041/E042 `!` fix-it was placed at that end column — right after `(`, mid-call. Fix (`post_solve_validation.rs`): a positional fix is offered only when the span's text on its line closes the expression (`fix_anchor_ends_expression`: ends in `)` with balanced parentheses outside string literals, or a bare name); otherwise the `!` stays in `try` as display only, with no `try_replace` / `suggestions` row. No position beats a wrong one for a harness that applies them.

## Evidence

`tests/fixit_span_on_line_test.rs` (4): a heredoc hole is reported at its own line and column (6:7, end 28); three holes across two heredoc lines each map to their own occurrence (a repeated spelling included); a multi-line call carries no positional fix; a one-line call still gets its insertion right after `)`. `tests/interpolation_span_parity_test.rs` (#2095) stays green. Locally: syntax + frontend crate tests and 236 root test binaries green (the wasm runtime interp ledger was still running when this PR was opened; CI decides).

## Docs

`docs/specs/cli.md` and `docs/mcp.md`: `col` / `end_col` are 1-based **character** counts (not bytes, not display width), `end_col` exclusive — the unit the issue asked to have written down.

## Not in this PR

A project-wide `almide check --json` is #2253 (split out of the issue's "what would help"). While probing shapes, an un-annotated `Result` `let` inside a lambda body turned out to raise no E041 at all: #2254.

codopsy: almide-syntax A 95 / 10 warnings, almide-frontend A 91 / 27 — both unchanged.


O6lvl4:
CI's fmt gate flagged `spec/wasm_cross/list_unique_by_nonscalar_key.almd`: its `println("int=${list.unique_by([ … ], (x) => x % 3)}")` is a call spanning lines INSIDE an interpolation hole. With the tokens now on their real lines, the formatter lays the call out the way it lays out any call that spans lines (one argument per line, the list expanded), where before — every token stamped onto the string's first line — it kept the source layout. That is a formatter behaviour change for this one shape, in the direction of consistency with the same call outside a string; the fixture is reformatted in its own commit, the output is idempotent, both targets still print the same bytes, and the parity goldens are unchanged (the manifests hash the normalised AST, not the text).

Issue #2288, opened 2026-09-17
Infer whether a fn-typed parameter escapes and render the non-escaping slot as a borrowed callable so a literal lambda handed to a user higher-order function is a scope, not a heap closure


O6lvl4:
Goal series "a closure has no ABI in Almide" — part 2 of 3 (#2287 → this → fusion as specialization). Residue of #2278 ask 1 / #2069 after PR #2286: a literal lambda handed to a **stdlib** combinator is now a scope (no `move`, no `__cap`, captures borrowed); the same lambda handed to a **user** higher-order function is still a heap closure.

## Reproduction

```almide
fn apply_all(xs: List[Int], f: (Int) -> Int) -> List[Int] = xs |> list.map((x) => f(x))
fn scaled(xs: List[Int], k: Int) -> List[Int] = apply_all(xs, (x) => x * k)

fn apply_str(xs: List[String], f: (String) -> String) -> List[String] = xs |> list.map((x) => f(x))
fn suffixed(names: List[String], tag: String) -> List[String] = apply_str(names, (n) => n + tag)

fn keep(f: (Int) -> Int) -> (Int) -> Int = f
```

## What is emitted (develop e73b33846, `--target rust`)

```rust
pub fn apply_all(xs: Vec<i64>, f: std::rc::Rc<dyn Fn(i64) -> i64>) -> Vec<i64> {
    (xs).into_iter().map(|x| (f.clone())(x)).collect::<Vec<_>>()
}
pub fn scaled(xs: Vec<i64>, k: i64) -> Vec<i64> {
    apply_all(xs, (std::rc::Rc::new(move |x: i64| (x).wrapping_mul(k)) as std::rc::Rc<dyn Fn(i64) -> i64>))
}
pub fn apply_str(xs: Vec<String>, f: std::rc::Rc<dyn Fn(String) -> String>) -> Vec<String> {
    (xs).into_iter().map(|x| (f.clone())(x)).collect::<Vec<_>>()
}
pub fn suffixed(names: Vec<String>, tag: String) -> Vec<String> {
    apply_str(names, {
        let __cap_4: String = tag;
        (std::rc::Rc::new(move |n: String| AlmideConcat::concat(n, __cap_4.clone())) as std::rc::Rc<dyn Fn(String) -> String>)
    })
}
pub fn keep(f: std::rc::Rc<dyn Fn(i64) -> i64>) -> std::rc::Rc<dyn Fn(i64) -> i64> {
    f.clone()
}
```

[...]

Issue #2098, opened 2026-09-10
The documented idiom is 1.4-4.3x slower than the shape the docs say to avoid: captured lists are cloned per closure, and enumerate materializes tuples


O6lvl4:
Measured 2026-09-11. almide 0.62.0 @ f73403eaa, Apple M4 Pro, macOS 26.3, median of 9–11, all outputs bit-identical (`1.2742241500970801`). Sources and emits: `../almide-references/bench-sources/idiomatic/`.

## Summary

**The shape `CLAUDE.md` tells you to write is 1.4×–4.3× slower than the shape it tells you to avoid** — on spectralnorm. The doc says *"Prefer list combinators over var + for"* and *"Prefer recursion over var + while"*; the benchmark corpus is written the way the doc says not to, and that is the version that is fast.

This is a doc/reality gap of exactly the class §"The LLM-Facing Surface Is the Stable Surface" exists to prevent, except the promise here is performance rather than compilability, so no gate sees it.

## The measurement

spectralnorm n=1200, four spellings of the same program:

| spelling | native | vs A | wasm | vs A | wasm B |
|---|---:|---:|---:|---:|---:|
| **A** `var` + `while` + `out[i] = s` (what the corpus ships) | **38.4 ms** | 1.00× | **80.6 ms** | 1.00× | 11,657 |
| **C** `enumerate \|> map \|> fold` (what CHEATSHEET shows) | 84.7 ms | 2.21× | 342.2 ms | **4.25×** | 7,855 |
| **D** `range \|> map \|> fold`, indexed | 90.7 ms | 2.36× | 143.4 ms | 1.78× | 7,741 |
| **E** `range \|> fold`, indexed | 62.9 ms | 1.64× | 109.4 ms | 1.36× | 7,733 |

**binarytrees is the control, and it is clean**: `var`+`while`, tail recursion, and `range |> map |> sum` all measure 49.6 / 49.6 / 49.7 ms native and 64.3 / 64.8 / 64.4 ms wasm. The region window fires in all three (`almide_region_window(|| __rgn_check(__rgn_make(19)))`), TCO turns the recursion into a loop, and the pipeline fuses to `.into_iter().map(…).fold(…)` with no intermediate list. **So combinators are not inherently slow — these two mechanisms are.**

## Mechanism 1 — a captured list is cloned once per closure construction

[...]

PR #2290, opened 2026-09-17
Decide an iteration's source ownership from what its body does with the element, not from the combinator's consume slot

### LINKED ISSUE #2287: Decide a fused chain source's ownership from what the step does with the element, not from the combinator's consume annotation
Goal series "a closure has no ABI in Almide" — part 1 of 3 (this → #2288 escape inference for fn-typed params → #2289 fusion as specialization). Residue of #2278 ask 1 / #2069 after PR #2286.

## Reproduction

```almide
fn mapped(ns: List[Int]) -> List[Int] = ns |> list.map((n) => n * 2)
fn total(ns: List[Int]) -> Int = ns |> list.fold(0, (acc, n) => acc + n)
fn shouted(ws: List[String]) -> List[String] = ws |> list.map((w) => w + "!")
fn twice(xs: List[Int]) -> Int = {
  let a = mapped(xs)
  let b = mapped(xs)
  list.len(a) + list.len(b)
}
```

## What is emitted (develop e73b33846, `--target rust`)

```rust
pub fn mapped(ns: Vec<i64>) -> Vec<i64> {
    (ns).into_iter().map(|n| (n).wrapping_mul(2i64)).collect::<Vec<_>>()
}
pub fn total(ns: Vec<i64>) -> i64 {
    (ns).into_iter().fold(0i64, |acc, n| (acc).wrapping_add(n))
}
pub fn shouted(ws: Vec<String>) -> Vec<String> {
    (ws).into_iter().map(|w| AlmideConcat::concat(w, "!".to_string())).collect::<Vec<_>>()
}
pub fn twice(xs: Vec<i64>) -> i64 {
    let mut a: Vec<i64> = mapped(xs.clone());
    let mut b: Vec<i64> = mapped(xs);
    ...
}
```

`mapped` and `total` read `Copy` elements and never need the vector; `shouted` consumes each `String` but the caller has to give up the whole vector for that. Every caller that still needs its list pays a full `Vec` clone (`twice`), and the alloc ledger counts it.

## Why it happens

The list slot of every list combinator is declared consumed in `stdlib/list.almd` — 30 functions carry `@consume(xs)` (`map`, `filter`, `fold`, `find`, `flat_map`, `filter_map`, `sort_by`, `partition`, `reduce`, `group_by`, `scan`, `zip_with`, `window`, …). `StreamFusion` reads that slot through `seed_intrinsic_sigs` (`pass_stream_fusion.rs`, `source_of` / `slot_consumes`) and hands the param to `BorrowInsertion` as consumed, so the param is owned no matter what the step lambda does with the element. `ChainSourceBorrowPass` (PR #2286) recovers only the two cases that are already borrowed by construction: a `Clone{source}` and a bare `&[T]` param. The remaining rule is the combinator's annotation, i.e. the **unfused runtime twin's** signature (`almide_rt_list_map(xs: Vec<A>, …)`) decides the **fused** form's ownership.

## What the references do

- The Zig rewrite of Roc keeps its `List.map` list slot owned (`list_map_prepare_reuse` has `consume_args = {0}`, `src/base/LowLevel.zig:884-886`; `arc_solve.zig:2258-2270` turns that bit into a demand) — but only because it reuses the allocation in place when the runtime refcount is 1 (`src/builtins/list.zig:334-336`, `listMapCanReuse`). That is a refcount decision; the native leg here has no refcount, ownership is static, and an owned `Vec` param forces the *caller* to clone.
- The **element's** mode there is a separate, per-callback decision: the generic path reads each element through `list_get_unsafe`, whose row is `RcEffect.retainsResultBorrowingArgs(argMask(&.{0}))` (`src/base/LowLevel.zig:923-926`), so the element is a borrow into the list and is retained only when an occurrence in the callback demands ownership (the demand list at `src/lir/arc_solve.zig:1630-1647`, `:2320-2365`). The in-place path moves the element out (`list_map_extract_unsafe`, `RcEffect.none()`, `:897-900`). Same body in Roc source: `src/build/roc/Builtin.roc:4122-4145`.
- Lean 4 (`InferBorrow.lean`) and Koka (`Parc.hs`) likewise decide a param's mode from its occurrences, never from the callee's declaration.

That is the rule to port: **the source's mode follows what the step does with the element, not what the combinator declares.**

## Asked

1. **Source mode from element consumption** (native, fused chains). For a chain rooted at a param `xs: List[T]`:
   - `T` is `Copy`, or the step lambda(s) never consume the element (read, compare, interpolate, hand to a borrowed slot) → `xs` is borrowed: `xs: &[T]`, source `xs.iter().copied()` / `.cloned()`.
   - the step consumes a non-`Copy` element (returns it, concatenates it, builds it into a constructor, hands it to an owned slot) and `xs` is dead after the chain → `xs` stays owned, source `xs.into_iter()` (today's shape; `shouted` above).
   - the step consumes a non-`Copy` element but `xs` is live afterwards → borrowed source with `.iter().cloned()`: one clone per element instead of one clone of the whole vector at the call site.
   The decision belongs to `StreamFusion` (it already owns `borrowed_lambda_params`), read by `BorrowInsertion` through the existing annotations; `ChainSourceBorrowPass` then only renders.
2. **`@consume(xs)` stops being the source of truth for fused chains.** The annotation stays for the runtime twin (the unfused path when the callback is a bound closure value) — that path is what #2289 retires. `slot_consumes` in `pass_stream_fusion.rs` is replaced by the element rule; the intrinsic sigs are no longer consulted for a fused source.
3. **Certifier**: C4 owned-never-consumed currently *grants* ownership to a chain source at a consume slot. After this change a param whose only occurrences are chain sources with non-consuming steps is a C4 violation. `scripts/check-ownership-certifier.sh` and its negative twin carry a fixture for both directions.
4. **Fallible twins** (`__fallible_*`, `list.map((f) => g(f)!)!`) follow the same rule.

## Acceptance

- The program above emits `mapped(ns: &[i64])`, `total(ns: &[i64])`, `shouted(ws: Vec<String>)`, and `twice` calls `mapped(&xs)` twice with no `.clone()`.
- `tests/native_borrow_oracle_test.rs`: the alloc ledger for the six pinned programs does not rise; a seventh program (the one above) is pinned at zero list clones.
- Corpus signature diff (spec/lang, spec/wasm_cross, spec/stdlib, research/benchmark/perf, examples; 1320 files): `pub fn` lines with `Vec<` params go down, `.clone()` count does not go up, `move |` / `__cap_` unchanged (`docs/specs/codegen.md` records the numbers).
- `almide test --target rust`, `proofs/output-parity.sh`, `scripts/check-ownership-certifier.sh` (+ `-negative`), `python3 research/benchmark/perf/bench.py --legs native --quick --runs 5` within noise (A/B interleaved twice on an idle machine), `spectralnorm/compare.py` unchanged, clippy ratchet flat.
- `docs/specs/codegen.md` states the rule in one paragraph; `docs/ARCHITECTURE.md` row 13b names it.

## Order

Land this before the escape-inference goal (#2288): once fused sources are borrowed by the element rule, the only remaining owned list slots are runtime twins, and part 2 can measure closure ABI cost on its own.

O6lvl4:
Closes #2287 — part 1 of the "a closure has no ABI" series (#2287 → #2288 → #2289).

## What changes

A fused chain's source (and a `for` loop's iterable) is borrowed or consumed by **what the body does with each element**, not by the combinator's `@consume(xs)` slot. The list slot's annotation described the unfused runtime twin; it now decides nothing for the fused path.

- **One rule, three readers.** `element_reads_only` (use_kind.rs) says whether an iteration only reads its element: a `Copy` element always; a heap element when every occurrence is a borrowed slot, a shared borrow, a clone, a field read, a nested borrowed iteration, or — for `String` — an interpolation part or a `&str` comparison, plus any consuming occurrence the clone pass clones anyway (a later occurrence in the same iteration, the E0505 guard). The borrow verdict reads it through the signature oracle before `BorrowInsertion`, the clone pass reads it on the final IR, and the ownership certifier's C4 reads the same `Iterable::consumed` bit.
- **Source mode written back.** `StreamFusion` starts every source consumed; after the fixed point `commit_chain_source_modes` flips the node to borrowed when no receiving lambda needs the element owned, so the clone pass, `ChainSourceBorrow` and the renderer see the mode the param verdict was built on. `source_element_receivers` (almide-ir) is the one place that says which lambdas the source element reaches (steps up to the first producing step, then the collector; `Collect` / `Find` / `Enumerate` mean the element leaves).
- **`&T` binders, no per-element clone.** When the source is borrowed and every receiver only reads, the clone pass marks the element binders like `for` binders (`borrowed_loop_vars`): the renderer iterates `.iter()`, BorrowLowering spells `w.as_str()` / naked `w`, and a filter-family step (which Rust hands `&&T`) is rebound `let w = *w` instead of `let w = w.clone()`. Fusion no longer applies the runtime twins' lambda preparation at all — its unconditional `let x = x.clone()` for a multi-use param cloned each element even when both uses were borrows; the clone pass runs after fusion and places what is needed.
- **Certifier**: only an iteration whose body consumes its elements justifies an owned param (C4). The certifier corrected the rule twice during this change (a consuming use cloned anyway; a `for` binder's field, which never moves).
- **Fallible twins** are self-hosted loops, so the `for` rule covers them.

## Emitted shapes

```rust
pub fn mapped(ns: &[i64]) -> Vec<i64> { (ns).iter().cloned().map(|n| n.wrapping_mul(2)).collect() }
pub fn twice(xs: &[i64]) -> i64 { let a = mapped(xs); let b = mapped(xs); … }      // was xs.clone()
pub fn lens(ws: &[String]) -> Vec<i64> { (ws).iter().map(|w| almide_rt_string_len(w.as_str())).collect() }
pub fn kept(ws: &[String]) -> i64 { (ws).iter().filter(|w| { let w: _ = (*w); … }).count() as i64 }
pub fn longest(ropes: &[Rope]) -> i64 { (ropes).iter().fold(0, |max, r| if measure(r) > max { measure(r) } else { max }) }  // was let r: Rope = r.clone()
pub fn shouted(ws: Vec<String>) -> Vec<String> { (ws).into_iter().map(|w| concat(w, "!")).collect() }  // element leaves: still owned
pub fn loop_len(ws: &[String]) -> i64 { for w in ws.iter() { … } }
```

## Reference

The Zig rewrite of Roc keeps `List.map`'s list slot owned for in-place reuse at refcount 1 (`src/base/LowLevel.zig:884-886`, `src/builtins/list.zig:334-336`), but decides the **element** per callback: the generic path reads through `list_get_unsafe`, `RcEffect.retainsResultBorrowingArgs` (`LowLevel.zig:923-926`), retained only when an occurrence demands it (`src/lir/arc_solve.zig:1630-1647`). On the native leg a borrowed source can only clone, so the per-element demand is lifted to the source's mode. Lean 4 (`InferBorrow.lean`) and Koka (`Parc.hs`) likewise decide a param's mode from its occurrences, never from the callee's declaration.

## Measurements

| gate | result |
|---|---|
| `almide test --target rust` spec/lang, spec/stdlib, spec/wasm_cross, spec/integration | 219 / 162 / 12 / 52 files, all pass |
| `scripts/check-ownership-certifier.sh` | 0 violations over 960 files |
| `tools/almide-gates` under `ALMIDE_CERTIFY_OWNERSHIP=fail` | 198 tests / 14 files; env-switches-doc OK |
| allocation ledger (`native_borrow_oracle_test`, new `strs` row + `fold`/`mapped` shapes) | list 587 → 572, strs 952 → 907 (develop → this), other rows unchanged |
| corpus signature diff (1185 files, develop vs this) | 75 `pub fn` lines gain a borrow (66 `&[T]` params), none lose one; `.clone()` 83995 → 83812; `move \|` / `__cap_` unchanged; 4 emit failures identical on both |
| `cargo test --release -p almide --no-fail-fast` | 1667 passed, 0 failed |
| `proofs/output-parity.sh` | OK, 409 baseline files byte-match |
| clippy (file, message) set diff, same toolchain | 1468 → 1468 after the last commit (the one new `to_string` line fixed); same toolchain, rustc 1.96.1 |
| `bench.py --legs native --quick --runs 5`, interleaved A/B ×2 on an idle machine | 17 benches, every ratio within 0.93–1.04 across both rounds, no consistent direction |

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>


O6lvl4:
Perf ratchet (spelling-ratio, #2098) went red on the first run: spectralnorm's indexed spelling measured 2.74x its imperative twin on the x86 runner (1.33x locally on ARM, documented 1.43x). Cause: `list.range(0, n)` as a chain source is a temporary, and the element rule made it iterate as `(temp).iter().cloned()` instead of `(temp).into_iter()` — a borrow of a value that dies at the same point, which the default opt-level compiles worse in the inner fold. b6564731d restricts the borrowed verdict to a PLACE source (a variable or a projection of one); a temporary stays consumed as before. Local re-measure: rust indexed 1.35x, enumerate 0.98x; wasm 1.31x / 1.05x. All pinned tests and spec/lang unchanged.

Issue #2296, opened 2026-09-18
The Critical profile checker skips pipe stages, constructor payloads, tuple indexes and six other expression kinds


O6lvl4:
## Reproduction

```almide
fn spin(n: Int) -> Int = n |> ((x) => {
  var i = 0
  while i < x { i = i + 1 }
  i
})

fn main() -> Unit = println(int.to_string(spin(3)))
```

`almide check --profile critical` prints `No errors`. The same loop in a plain body is E070, and a let-bound lambda is E074 + E070.

## Why

The bounded-profile walk (`crates/almide-frontend/src/check/bounded.rs`) recurses into containers through `walk_children`, whose match ended in `_ => {}`. Every expression kind it did not list hid its contents from every ALS §B rule:

| kind | shape that passed `--profile critical` |
|---|---|
| `Pipe` | `n \|> ((m) => { while … })`, and any stage call (`xs \|> list.map(f)`) |
| `Some` / `Ok` / `Err` | `ok({ var i = 0 while i < n { … } i })` |
| `TupleIndex` | `({ while … (i, 0) }).0` |
| `IfLet` | `if let v = o { v + { while … } } else { 0 }` |
| `SpreadRecord` | `{ ...p, x: { while … } }` |
| `OptionalChain` | `(some({ x: { while … }, y: 0 }))?.x` |
| `TypeAscription` | `int.to_string({ while … }: Int)` |
| `InterpolatedString` holes | `"${x * 2.0}"` (E077) — the rule arm flagged allocation in loops but never walked the parts |
| `Compose` | `let g = double >> double` (E074, a function reference per ALS-B7) |

The profile's claim ("every fn is checked against E070–E078", PARTITIONED-RUNTIME.md's bounded-stack row) did not hold for any program that spells a violation inside one of these kinds. Measured over `spec/wasm_cross`: 341 files passed `--profile critical`; with a complete walk, 300 do. The 41 that flip are all real violations (closures and higher-order calls in pipe stages, Float arithmetic in interpolations and constructor payloads, run-time-length construction).

## Fix

- `walk_children` becomes exhaustive with no wildcard, so a new `ExprKind` variant fails to compile until someone decides how the profile walks it.
[...]

PR #2299, opened 2026-09-18
Release v0.63.0-rc1

O6lvl4:
develop → main for v0.63.0-rc1 (Cargo.toml already 0.63.0, #2294). The RC tag goes on this PR's merge commit; the final v0.63.0 follows the soak.

- Interface diff v0.62.0 → develop: additive (1 added, 0 removed: `string.byte_slice`).
- Release blockers: 0.
- Aboard since v0.62.0: 111 PRs through #2292, plus the version bump (#2294), the rustls advisory update (#2295, RUSTSEC-2026-0285) and the Critical profile walk fix (#2298, #2296).
- Curated notes are applied after the tag workflow creates the prerelease.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


O6lvl4:
**Ratchet-commit separation is red on exactly one commit, and that commit predates the rule it fails.**

- Flagged: `afc6031e9` (#2191, "Report the effect fn tail, match arm, if branch and guard else that still propagated a Result implicitly"). It changes checker source and `crates/almide-syntax/tests/golden/spec-ast-manifest.txt` together.
- The golden manifests became ratchet artifacts, and the check started judging every commit of a range, in `5b56db6c6` (#2199, 2026-09-14 14:20 JST).
- On develop's history `afc6031e9` is an ancestor of `5b56db6c6`: it landed before the rule existed, so the rule in force when it merged did not classify the manifest as a ratchet.
- Every other commit in `main..develop` passes. This release PR's range is the first to span the rule's introduction; once v0.63.0 is on main, no later release range will contain this commit.

The check is not a required status on main. Whether to merge with this one documented red, or to change the gate first, is left to the maintainer: changing the gate as part of the release would be the releaser moving the bar that judges the release (the F5 concern the gate exists for).

Issue #2311, opened 2026-09-18
The interpreter reads a callback's Int result as a heap address when the value collides with a live block


O6lvl4:
## Summary

The interpreter leg of the three-way oracle abstains on the fallible `list.map` once the input list has 16 or more elements:

```
out-of-interp-scope capability: return sync: a block under an unspellable container return type (Applied)
```

```almide
effect fn bump(x: Int) -> Int = x + 1

effect fn main() -> Unit = {
  let xs = list.range(0, 16)
  let m = (xs |> list.map((x) => bump(x)!))!
  println("${list.len(m)}")
}
```

| `list.range(0, n)` | interp |
|---|---|
| 3, 8 | evaluates, prints n |
| 16, 17, 32, 64, 65, 100, 200 | abstains with the message above |

Identical on develop 595a4765d and on the #2291 / #2306 branch, so it predates that change. Driven through the `run_interp_capture` helper the ledger test uses. The other six fallible carriers (`filter`, `flat_map`, `filter_map`, `fold`, `find`, `each`) evaluate at 200 elements.

The message comes from `dispatch_sync.rs` `sync_addr`: a returned block whose declared type is a container the pool tier cannot spell. The 16-element step suggests the list crosses into a block representation at that size and `__fallible_map`'s `Result[List[B], E]` read-back has no rule for it.

## Effect

`spec/wasm_cross/fallible_hof_large_input.almd` (C-284, 200,000 elements) is recorded in `interp-abstain-ledger.txt` for this reason, so the interpreter does not cross-check any of its cells. When this is fixed, the ledger test reports the stale abstention and the fixture joins the evaluated set.

O6lvl4:
Correction to the "Effect" section: the final version of the #2291 / #2306 change does not ledger an abstention. The large-input `map` cells moved out of `spec/wasm_cross/fallible_hof_large_input.almd` into `spec/lang/fallible_hof_test.almd` (native and wasm lanes), so the fixture keeps the other six carriers under the three-way oracle at 200,000 elements. The cost of this gap is that the interpreter does not cross-check the fallible `map` at size. When this is fixed, move those two cells back into the fixture.

O6lvl4:
The diagnosis in this issue is wrong, and the title is wrong with it. This has nothing to do with the list's length. It is an integer whose VALUE collides with a live heap block address.

### What actually decides it

The abstain follows the callback's result values, not the input size. Same program, same carrier, only the arithmetic changed:

| callback | first failing `n` |
[...]

Issue #2310, opened 2026-09-18
Structural wasm leg copies the whole list on every acc + [x] when the element is a heap type


O6lvl4:
## Summary

On the structural wasm leg (the default), growing a list by `acc + [x]` copies the whole list at every step when the element is a heap type. The same loop over `Int` is linear, and `list.push` is linear for both. So the accumulator idiom is quadratic exactly where the data is strings.

Measured on develop 595a4765d, macOS arm64, wasmtime, 100,000 appends:

| shape | `Int` element | `String` element |
|---|---|---|
| tail recursion, `go(f(x), f, n - 1, acc + [x])` | 11 ms | 10,353 ms |
| `var out` + `for`, `out = out + [x]` | 11 ms | 10,349 ms |
| `var out` + `for`, `list.push(out, x)` | 11 ms | 12 ms |

```almide
fn it[T](seed: T, f: (T) -> T, n: Int) -> List[T] = {
  var out: List[T] = []
  var x = seed
  for _ in 0..<n {
    out = out + [x]
    x = f(x)
  }
  out
}
fn main() -> Unit = println("${list.len(it("a", (s) => if string.len(s) > 3 then "a" else s + "a", 100000))}")
```

Native is linear for all six cells.

## Context

The incumbent leg's `concat_to_append` rewrite (`crates/almide-mir/src/concat_to_append.rs`) exists for exactly this window (`x = x + [e]` becoming an in-place `__list_append1` when `rc == 1`). The structural leg appears to apply the in-place form only to scalar elements.

This matters beyond user code. The #2291 / #2306 fix had to write `list.iterate` and the fallible list carriers with `list.push` rather than the accumulator form, because the accumulator form was quadratic for `String` elements on the default leg. A 200,000-element `list.iterate` over strings ran out of memory that way.

## Gate

A `spec/wasm_cross` fixture that builds 1,000,000 `String` elements by `acc + [x]` in both a tail-recursive and a `var` + `for` loop, plus a perf-ratio row (`scripts/check-perf-ratio.sh`) against native, so the quadratic form cannot come back.

Issue #2309, opened 2026-09-18
The wasm size ledger records a size that depends on the fixtures compiled before it in the same process


O6lvl4:
## Summary

`crates/almide-wasm/tests/size_ratchet.rs` measures every manifest fixture in one process (`lower_to_ir` + `emit_program_with_ops` + `to_wasi` in a loop). The size it records for a fixture depends on what the fixtures before it compiled, so the ledger does not pin what `almide build --target wasm` ships.

## Evidence

On develop 595a4765d:

| fixture | `size-baseline-wasi.txt` | `almide build <f> --target wasm` |
|---|---|---|
| `spec/wasm_cross/list_zip_with_typed.almd` | 5,204 | 6,869 |

The #2291 / #2306 PR changes seven self-hosted stdlib bodies (the codepoint counters and the fallible list carriers). Re-ratifying the ledger then moved **35** rows. For **26** of them the CLI output is byte-identical before and after the change (`cmp` of the two binaries' `.wasm`), including:

- `list_zip_with_typed` (5,204 → 6,869, now equal to the CLI)
- the seven `codec_*` fixtures (+95 each)
- `env_args`, `env_get`, `env_set_overlay`, `args_surface`, `bundled_pure_modules`
- `fs_write_errno` (+338)

None of those fixtures calls a changed function. Their ledger rows moved because an earlier fixture in the corpus order linked a different stdlib body.

## Why it matters

- A PR that touches one stdlib body re-ratifies rows it did not change, so the size diff a reviewer reads is noise mixed with signal. In that PR, 26 of 35 moved rows were noise.
- A real regression in a late fixture can be masked or faked by state an earlier fixture left behind.
- Whatever the shared state is, it makes a later module's emission depend on the previous program. That is harmless for the one-program-per-process CLI, but it is the class of state `JsHostGuard` / the heap-cap knob scope with guards for exactly this reason.

## Fix direction

Find the process-global state the structural emitter keeps across `emit_program_with_ops` calls (a linked-helper / self-host cache, an interner-keyed table) and scope it per program. Gate: the size ratchet measures each fixture twice, once in corpus order and once alone, and refuses a difference.

O6lvl4:
Taking this (session almide-5f). #2306 stays with PR #2313.

O6lvl4:
Fixed by #2321 (merged as bbbcce3b7).

[...]

Issue #2308, opened 2026-09-18
Incumbent wasm leg double-frees in list.filter_map when an effect fn callback answers an Option


O6lvl4:
## Summary

On the incumbent wasm leg, `list.filter_map` with a user `effect fn` callback that answers `Int?` traps in `rc_dec` (`wasm unreachable instruction executed`) on a three-element list. Native and the structural leg print the right list. The same callback shape through `list.map` / `list.filter` / `list.flat_map` runs.

```almide
effect fn third(x: Int) -> Int? = if x % 3 == 0 then some(x) else none

effect fn main() -> Unit = {
  let fl = ([0, 1, 2, 3, 4, 5, 6] |> list.filter_map((x) => third(x)!))!
  println("filter_map=${fl}")
}
```

| build | output |
|---|---|
| native | `filter_map=[0, 3, 6]` |
| wasm, structural leg (default) | `filter_map=[0, 3, 6]` |
| wasm, `ALMIDE_WASM_INCUMBENT=1` | trap in `rc_dec` called from `almide_rt_list___fallible_filter_map__Int_Int_String` |

Present on develop 595a4765d (the carrier's recursive body) and with the loop body from the #2291 / #2306 fix, so the carrier's shape is not the cause. A plain `fn third(x: Int) -> Result[Int?, String] = ok(opt(x))` callback runs on the incumbent, which points at the effect-fn carrier strip / re-wrap around an `Option` payload (the C-284 area) rather than at the list code.

## Reach

A program only reaches the incumbent when the structural leg walls on something else in it, or under `ALMIDE_WASM_INCUMBENT=1`. The `rc_dec` trap is the RC underflow guard firing, so the incumbent drops the `Option` block once too often. Without the guard that would be a use after free.

## Gate

The large-input fixture for C-284 in the #2291 / #2306 PR drives `filter_map` with a plain fn for this reason. When this is fixed, switch that cell back to an `effect fn` so the fixture covers the shape on both legs.

O6lvl4:
Taking this (session almide-5f). #2306 stays with PR #2313.

PR #2313, opened 2026-09-18
Run the self-hosted stdlib in constant stack: codepoint counters, list.iterate and the fallible list carriers, with a gate on input-length recursion

### LINKED ISSUE #2291: wasm: string.index_of traps with 'call stack exhausted' in __cp_count when the match is ~20 KiB or further into the string (native is fine)
`string.index_of` traps on `--target wasm` when the match position is far enough into the string. Native returns the right answer for the same input.

The recursion is in `__cp_count` (the same address repeats all the way down the backtrace), so it looks like the byte offset → codepoint offset conversion recurses once per codepoint of the prefix.

**It is the match position that matters, not the string size** — a match at offset 0 in a 160 KB string is fine. That is what makes it hard to notice: it depends on where the needle happens to sit.

## Repro

`almide init`, then:

```almide
import fs

effect fn main() -> Unit = {
  for name in ["early.txt", "late.txt"] {
    let s = fs.read_text(name) ?? ""
    println(name + ": len=" + int.to_string(s.len()))
    match string.index_of(s, "NEEDLE") {
      some(i) => println("  found at " + int.to_string(i)),
      none => println("  not found"),
    }
  }
}
```

```toml
[permissions]
allow = ["FS.read", "IO"]
```

Two files of **exactly the same size** (160,007 bytes), 2000 lines of `xxxx…` (79 chars + newline).
`early.txt` has `NEEDLE` on the first line, `late.txt` has it on the last line.

```
$ almide build && ./idxbug
early.txt: len=160007
  found at 0
late.txt: len=160007
  found at 160000          # native: correct

$ almide build src/main.almd --target wasm -o idxbug.wasm && wasmtime --dir=. idxbug.wasm
early.txt: len=160007
  found at 0
late.txt: len=160007
Error: failed to run main module `idxbug.wasm`
Caused by:
    1: error while executing at wasm backtrace:
    0: <unknown>!__cp_count
    1:   0x36eb - <unknown>!__cp_count
    ... (20 frames elided, all the same address)
    2: wasm trap: call stack exhausted
```

## Threshold

ASCII-only input, needle at the end:

| match offset | wasm |
|---|---|
| 5,120 | ok |
| 10,240 | ok |
| 20,480 | **trap** |
| 40,960 | trap |
| 160,000 | trap |

So the ceiling is somewhere between 10 KiB and 20 KiB of prefix. That is low enough that ordinary source files hit it.

## `contains` is not affected

The same 8 files, `s.contains("NEEDLE")` instead — all print `yes`, including the 160 KB one.
That fits the theory: `contains` does not have to report a position, so no offset conversion is needed.

## Version / legs

- `almide 0.62.0` (from `tools/install.sh`), macOS arm64, `wasmtime` 
- The build line says `incumbent v1 leg, verified` for both `almide build --target wasm` and with `ALMIDE_WASM_INCUMBENT=0` (this program is small; both legs produced the same 15,757-byte artifact)

## How we hit it

[teastia](https://github.com/Aid-On/teastia) reads JS/TS sources and checks `native == wasm` byte-for-byte across a corpus of real repositories. One `string.index_of(src, "://")` on a 143 KB `App.tsx` (the `://` sits at line 2598 of 2989) made 8 of 16 subcommands trap on wasm while native was fine. Every fixture in the test suite is a few KB, so nothing in CI caught it — only the corpus run did.

We worked around it with `contains`, and we are avoiding `string.index_of` on whole-file strings from now on. Filing it because the threshold is low and the position-dependence makes it easy to ship without noticing.

### LINKED ISSUE #2306: list.iterate overflows the stack on both targets, at about 10,000 elements on wasm and 50,000 on native
## Summary

`list.iterate(seed, f, n)` overflows the stack on **both** targets at sizes an ordinary program reaches, and at different sizes on each, so between the two thresholds the legs disagree on output and exit code.

```almide
fn main() -> Unit = {
  let xs = list.iterate(0, (x) => x + 1, 10000)
  println("${list.len(xs)}")
}
```

| n | wasm (wasmtime) | native |
|---|---|---|
| 7,000 | 7000 | 7000 |
| 10,000 | `wasm trap: call stack exhausted` | 10000 |
| 20,000 | `wasm trap: call stack exhausted` | 20000 |
| 50,000 | `wasm trap: call stack exhausted` | `fatal runtime error: stack overflow, aborting` |
| 100,000 | `wasm trap: call stack exhausted` | `fatal runtime error: stack overflow, aborting` |

Measured on develop e73b33846, macOS arm64.

## Cause

`stdlib/list.almd`:

```almide
fn iterate[T](seed: T, f: (T) -> T, n: Int) -> List[T] = if n <= 0 then []
else [seed] + iterate(f(seed), f, n - 1)
```

The self-call is not in tail position (`[seed] + …` runs after it), so the depth is `n` on both legs. Each level also builds a fresh list from `[seed]` and the tail, so the body is quadratic in `n` as well as deep.

## Fix

Tail-recursive with an accumulator that the tail-call loop lowers to a loop on both legs: depth O(1), and appending to a uniquely owned accumulator stays linear. The same PR as #2291 adds the large-`n` cross-target fixture and a static gate on non-tail self-recursion whose depth grows with the input across the self-hosted stdlib, so the class is closed and not just these two instances.

O6lvl4:
Closes #2291. Closes #2306.

## What was wrong

Seven self-hosted stdlib bodies recursed once per byte or element in non-tail position, so an ordinary input overflowed the stack, at a size that differs per target:

| call | wasm (wasmtime) | native |
|---|---|---|
| `string.index_of` / `last_index_of`, match past ~16 KB | `call stack exhausted` | ok |
| `string.count(s, "")` over ~16 KB | `call stack exhausted` | ok |
| `list.iterate(seed, f, n)` | traps at ~10,000 | aborts at ~50,000 |
| `xs \|> list.map((x) => f(x)!)!` (and filter / flat_map / filter_map) | traps at ~10,000 | aborts at ~50,000 |

The fourth row is the fallible form the language reference recommends. The carriers behind it also copied the rest of the list at every element (`list.drop(xs, 1)`), so they were quadratic even below the overflow.

## The fix

- `__cp_count` (string_search) and its twin `__pad_cp_count` (string_pad) carry the running count in an accumulator, so the self-call is a tail call.
- `list.iterate` and all seven `__fallible_*` carriers are loops that grow the result with `list.push`. The family moves together: fold / find / each were tail calls already but paid the same per-element `list.drop` copy.
  - Why `list.push` and not `acc + [x]`: on the structural wasm leg `acc + [x]` copies the whole list per step for a heap element (100,000 `String` appends: 10.3 s against 12 ms). Filed as #2310.
  - Why a `stop` flag and not `break`: the structural leg does not lower `break` yet.
  - `__fallible_find` keeps the hit as an index and rebuilds `ok(some(v))` / `ok(none)`, because the incumbent leg walls on the shorter spellings. Without that, a program that falls back to the incumbent would stop building.
- Observable semantics are unchanged: same results, the first-err short-circuit (no element after the first err reaches the callback), and `f` runs exactly `n` times in `iterate`.

## The gate

`tests/stdlib_recursion_depth_gate_test.rs` parses every `stdlib/*.almd` and finds each non-tail call inside a recursive group (a self-call, or calls between functions of one file that reach each other). Each one must be listed in `BOUNDED` with the reason its depth does not follow the input's length: digits of an Int, nesting of a document, segments of a path, and so on. A listed call that no longer exists also fails, so the list only shrinks.

- The regex engine's backtracking points are listed with the reason stated honestly. They are bounded by the pattern except under a repeated group, which is the open #2307, and they go stale when #2307 lands.
- A second test pins the detector itself on tail and non-tail shapes, including the pre-fix `__cp_count` and `iterate` verbatim.
- Against the pre-fix stdlib the gate fails naming exactly the seven functions above.

## Evidence

- `spec/wasm_cross/string_position_large_offset.almd` (C-065, C-017): `index_of` / `last_index_of` / `count(s, "")` / `pad_*` at 1,000,000 ASCII bytes and 1,600,000 mixed-width bytes.
- `spec/wasm_cross/fallible_hof_large_input.almd` (C-284): filter / flat_map / filter_map / fold / find / each over 200,000 elements, five with a user `effect fn` callback. An `each` cell prints if any callback runs after the first err.
- `spec/lang/fallible_hof_test.almd`: the fallible `map` over 200,000 elements, and its first err at element 150,000. It lives here rather than in the fixture because the interpreter leg abstains on the fallible `map` past 16 elements (#2311, predates this change), and that would take the whole fixture out of the three-way oracle.
- `spec/stdlib/list_bundled_test.almd`: `list.iterate` at 1,000,000 `Int` and 200,000 `String` elements, and zero and negative counts.

Both fixtures are byte-identical on native, the structural leg and the incumbent leg (`ALMIDE_WASM_INCUMBENT=1`), and the interpreter evaluates both. The spec tests pass on native and on the wasm lane. With the develop binary, all four fail:

| | develop 595a4765d | this branch |
|---|---|---|
| string fixture | wasm: `call stack exhausted` | identical on all legs |
| fallible fixture | native: stack overflow, wasm: out of memory | identical on all legs |
| fallible `map` tests | native: stack overflow | pass, native and wasm |
| iterate test | native: stack overflow | pass, native and wasm |

Full runs on this branch: `almide test spec/` 446/446 files on native, and 438 passed with 8 existing `wasm:skip` on wasm. `cargo test --workspace --release` is green except the host-determinism harness, which needs the `wasm32-wasip1` target that is not installed on the machine this ran on (CI has it).

## Ledgers

- Contract evidence rows added to C-065, C-017 and C-284. No statement changed and no new id was minted.
- Run, check and AST manifests: only the two new fixtures, the two edited spec tests, and the generator header moved.
- Size ledger: 35 rows moved. For 9 of them the shipped module changed (+8 to +10 bytes for the counter's extra parameter, and the fallible fixtures). For the other 26, `almide build --target wasm` output is byte-identical before and after (`cmp`). Those rows moved because the ratchet measures the corpus in one process and carries state between fixtures; filed as #2309.
- Allocation ledger: the two new fixtures plus three small decreases.

## Found on the way, filed separately

- #2307: regex repetition recurses per match on both engines. Native overflows at ~30,000 characters of `a*`; a repeated group overflows on both targets.
- #2308: the incumbent leg double-frees in `list.filter_map` when an `effect fn` callback answers an `Option`. It predates this change; the fallible fixture's filter_map cell uses a plain fn for now.
- #2311: the interpreter abstains on the fallible `map` past 16 elements. It predates this change.
- #2309: the size ledger's cross-fixture state.
- #2310: `acc + [x]` is quadratic for heap elements on the structural leg.


O6lvl4:
Heads-up from the session that landed #2326 / #2329 / #2330 / #2323 today: develop moved under this branch and it is now DIRTY. I have not touched it.

The conflicts are all in GENERATED files, no source:

```
CONFLICT (content): crates/almide-spine/tests/golden/spec-check-manifest.txt
CONFLICT (content): crates/almide-spine/tests/golden/spec-run-manifest.txt
CONFLICT (content): crates/almide-syntax/tests/golden/spec-ast-manifest.txt
CONFLICT (content): docs/contracts/conformance.md
```

(`git merge-tree --write-tree origin/develop origin/fix-2291-stdlib-recursion-depth`; the alloc, size and size-wasi baselines and contracts.toml all auto-merge.)

Resolution that worked on the four branches above: take either side to get the rebase moving, then REGENERATE rather than hand-merge —

```
ORACLE=<the CLI built from this tree> ALMIDE_BIN=<same> bash scripts/gen-ast-manifest.sh
… gen-check-manifest.sh … gen-run-manifest.sh
bash docs/contracts/generate-conformance.sh > docs/contracts/conformance.md
```

and commit the regeneration on its own so ratchet-separation stays happy. Three traps worth knowing, each of which cost a CI round today:

- `proofs/ledger-counts.toml` drifted and the nightly is red — refreshed on develop in #2334, so rebase after that lands or you will re-measure against a stale stamp.
- A stdlib source edit may not trigger a rebuild before you regenerate the size/alloc ledgers; check that `cargo build --release` actually recompiled, or the ledger records the OLD engine.
- Regenerating `crates/almide-interp/interp-abstain-ledger.txt` rewords eight unrelated rows and breaks `scripts/check-abstain-classes.sh` (#2335... filed as #2333). Add your row by hand and leave the rest.

#2306 and #2291 are the last two release blockers for the final v0.63.0 tag — everything else is closed.

O6lvl4:
Rebased onto `develop` (the branch was identical to origin before I touched it, so nothing was lost). The two conflicts were both generated files — the three parity manifests and `docs/contracts/conformance.md` — resolved by taking one side and then **regenerating** with a CLI built from this tree, never by hand-merging. The size and alloc ledgers were re-recorded the same way (`cargo build --release` needed `touch stdlib crates/almide-types/build.rs` first: `cargo:rerun-if-changed` on a directory does not see a file's content change, so the embedded stdlib would otherwise have been the old one).

The rebase brought in the new regex engine, and the depth gate in this PR caught it — twice, in both directions:

- **Undeclared**: `__rx_grow -> __rx_drive`. The engine resumes the suspended step inside the frame that owns the bigger arena, so the resume nests once per **doubling** — log₂ of the arena's final size, not the input's length. Declared as `ARENA_DOUBLING`.
- **Stale**: all five `REGEX_BACKTRACK` entries (`__rx_alts`, `__rx_rep_group` ×2, `__rx_rep_lazy`, `__rx_repi_try`). That is exactly what the entry's own comment predicted — "when #2307 replaces these with an explicit backtrack stack the entries go stale and this gate makes the fix delete them" — so they are gone, and the surviving comment records why.

Full `cargo test --release --workspace` on the rebased tree: green, 0 failed.

Issue #2307, opened 2026-09-18
Regex repetition recurses per match on both engines: native overflows at about 30,000 characters of a*, and a repeated group overflows on both targets at different counts


O6lvl4:
## Summary

Both regex engines recurse once per repetition, so a flat input overflows the stack, and the two engines do it at different sizes. Between the thresholds the targets disagree on output and exit code.

Measured on develop 595a4765d, macOS arm64, `wasmtime` for the wasm leg:

| pattern | input | wasm | native |
|---|---|---|---|
| `^a*b$` | `"a"` × 10,000 + `"b"` | true | true |
| `^a*b$` | `"a"` × 30,000 + `"b"` | true | `fatal runtime error: stack overflow, aborting` |
| `^a*b$` | `"a"` × 200,000 + `"b"` | true | stack overflow |
| `^[a-z]+$` | `"x"` × 200,000 | true | stack overflow |
| `^a*?b$` | `"a"` × 200,000 + `"b"` | true | stack overflow |
| `^(ab)*$` | `"ab"` × 1,000 | true | true |
| `^(ab)*$` | `"ab"` × 5,000 | `call stack exhausted` | true |
| `^(ab)*$` | `"ab"` × 10,000 | `call stack exhausted` | stack overflow |

```almide
import regex
fn main() -> Unit = {
  let s = string.repeat("a", 30000) + "b"
  println("${regex.is_match("^a*b$", s)}")
}
```

## Cause

- **Native** (`runtime/rs/src/regex.rs`): the matcher is continuation-passing backtracking. `rx_rep` → `rx_rep_more` → `rx_one` → the continuation `k`, which re-enters `rx_rep` for the next repetition, so every matched repetition of `*` / `+` / `{n,}` is one more Rust frame chain. A single-char atom repeated 30,000 times exhausts the main thread's 8 MB stack.
- **Wasm** (`stdlib/regex_engine.almd`): #1524 made the single-atom repetition iterative (`__rx_repi_*`), which is why `a*` and `[a-z]+` survive there. A repeated **group** still recurses per repetition: `__rx_rep_group` calls `__rx_seq` / `__rx_group` in non-tail position.

So the legs split in both directions: single-atom repetition is safe on wasm and dies on native, and group repetition dies on both at different counts.

## Expected

Repetition depth should not be bounded by the call stack on either leg: a flat input of any length the program can hold matches or fails identically on both targets. `^a*b$` against a 1 MB line is an ordinary log-scanning call.

## Fix direction

[...]

Issue #1524, opened 2026-08-18
wasm: stdlib regex recurses per matched character (__rx_rep_bounded) and exhausts the default wasmtime stack on ~1,000-char matches


atzurody:
## Repro (almide 0.57.1, wasmtime 47.0.3)

```almide
import regex
effect fn main() -> Unit = {
  let line = "const s = \"" + string.repeat("a", 1200) + "\";"
  println(int.to_string(regex.find_all("\"[^\"]{20,}\"", line).len()))
}
```

Native prints `1`. On wasm (`almide build --target wasm`, `wasmtime x.wasm`):

```
error while executing at wasm backtrace:
    0: <unknown>!__rx_char_match
    1: <unknown>!__rx_one
    2: <unknown>!__mutual_tco_0___rx_rep_bounded
    3: <unknown>!__mutual_tco_0___rx_rep_bounded
    ... (one frame per matched character)
wasm trap: call stack exhausted
```

Threshold with wasmtime's default 512 KiB stack is between 1,000 and 1,200 matched characters; `wasmtime -W max-wasm-stack=8388608` makes 5,000-char matches pass, so it is recursion depth proportional to the **match length** (bounded/greedy repetition), not input length — `regex.find_all` over a 54 KB file with short matches is fine.

Real-world trigger: minified/JSX lines in a Next.js app (`components/ui/sheet.tsx` has a 1,232-char line; an icons file has 2,579). Found running teastia (aid-on/teastia 1.5) under wasm on vercel/ai-chatbot; worked around by scanning long lines in 800-char windows. An iterative `rep_bounded` (or a documented match-length limit) would let wasm match native here.

O6lvl4:
Fixed in f2626e8bc. The recursive `__rx_rep`/`__rx_rep_bounded` (and their capture-threading `_c` twins) inspect the recursive result to decide the fallback, so the call was never a tail call — one real wasm frame per matched character. A NON-GROUP atom consumes exactly one char per repetition and never writes a capture slot, so the same greedy semantics now flatten into two mutually TAIL-recursive walks the renderer's TCO turns into loops: extend forward to the deepest count, then try the continuation at each position DESCENDING (`__rx_back`, the UTF-8 inverse of the `t + __rx_w(t)` forward step) until `min`. Group atoms keep the recursive path — their depth is group-repetition count, not match length.

[...]

PR #2323, opened 2026-09-19
Walk both regex backtrackers on the heap so repetition depth is not the call stack's

O6lvl4:
Draft for #2307. The two engines are fixed and agree; the branch is blocked on one thing, stated at the bottom.

## What was wrong

Both regex engines recursed once per matched repetition, so a flat input overflowed the stack at a size that differed per target, and between the thresholds the targets disagreed on output and exit code.

## The change

Both engines now keep the backtracking walk's state on the heap instead of in call frames, with the search order they always had: leftmost alternative first, greedy longest-first, lazy shortest-first, an empty repetition instance ends its loop, and a group's capture is recorded at the group's end and restored when the walk backtracks past it.

- **Native** (`runtime/rs/src/regex.rs`): the continuation-passing backtracker becomes three vectors. `choices` holds untried alternatives, each recording how long the other two vectors were when it was pushed; `frames` holds what to do when a group instance ends, naming its successor by index so the continuation is a shared linked list; `trail` holds the capture slots a group end overwrote. A single-atom repetition needs no frame at all: greedy takes as many as it can in a loop and leaves one choice that gives them back one at a time.
- **Self-hosted** (`stdlib/regex_engine.almd`): the same shape, so the two legs agree step for step. #1524 had already made the single-atom case iterative there; this extends it to repeated groups.

## Evidence

The issue's table, run on this branch's binary and on v0.63.0-rc1:

| pattern, input | rc1 native | rc1 wasm | this branch, both |
|---|---|---|---|
| `^a*b$`, 30,000 a | stack overflow | true | true |
| `^a*b$`, 200,000 a | stack overflow | true | true |
| `^[a-z]+$`, 200,000 x | stack overflow | true | true |
| `^a*?b$`, 200,000 a | stack overflow | true | true |
| `^(ab)*$`, 5,000 ab | true | call stack exhausted | true |
| `^(ab)*$`, 10,000 ab | stack overflow | call stack exhausted | true |

`spec/wasm_cross/regex_repetition_depth.almd` covers single atom, class, lazy, counted, repeated group, a group repetition that must give repetitions back, and the capture a long group repetition leaves, at 200,000 characters and 100,000 repetitions, plus the same rules at a short length. Native and the structural wasm leg are byte-identical, and it runs in 0.6 s. `spec/stdlib/regex_test.almd` passes on both legs, 31 tests.

## What blocks it

**Every regex program now walls on the incumbent wasm leg**, which is a regression: v0.63.0-rc1 runs them there.

The engine's dispatch is a 15-function mutual recursion group (`__rx_seq`, `__rx_fail`, `__rx_resume`, `__rx_repi_try`, …). `run_mutual_tco` in almide-optimize fuses such a group into one `__mutual_tco_<n>_<entry>` function with thin wrappers. The structural leg emits it; the incumbent leg has no knowledge of that name at all (`git grep mutual_tco crates/almide-mir` is empty), so it refuses with `wall: unlinked stdlib/runtime call(s) with no wasm definition: __mutual_tco_0___rx_repi_try`. No stdlib module had reached that path before, because none had a mutual recursion group in its self-hosted code.

Inlining a single-caller member (`__rx_back_resume`, done in this branch) only moves which member the pass picks as the entry.

The blocker is filed separately as #2325 (it is a general limitation: any self-hosted mutual recursion group hits it). Two ways forward, neither small:
1. Restructure the self-hosted engine's dispatch into one self-recursive driver over the heap state it already keeps, so there is no mutual group for the pass to fuse.
2. Teach the incumbent leg to link the fused helper, which is a general fix: any future self-hosted mutual recursion would hit the same wall.

The native half stands on its own and could land first if the self-hosted half is held back.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


O6lvl4:
Rebased onto develop now that #2326 has landed. #2325 was the blocker: the incumbent leg walled on the fused `__mutual_tco_0___rx_repi_try` because a `match` in a branch arm's tail had no real-branch route, and `__rx_resume`'s dispatch is exactly that shape.

Verified on the rebased branch, `spec/wasm_cross/regex_repetition_depth.almd`:

| leg | exit | rows |
|---|---|---|
| native | 0 | 12, md5 `67e7c4f2` |
| structural wasm | 0 | 12, md5 `67e7c4f2` |
| incumbent wasm | 0 | 12, md5 `67e7c4f2` |

Byte-identical across all three. Out of draft.

O6lvl4:
Rebase fallout cleared, three items, all local-gate verified:

- `spec_corpus_check_matches_oracle_hashes` — the fixture's rows were missing from all three parity manifests (ast / check / run). Regenerated with `ORACLE=` the CLI built from this tree.
- `fmt --check` — the fixture kept a blank line before an in-block comment, which the formatter removes. Formatted, then the manifests re-recorded because the ast hash moved with it.
- `interp_abstain_ledger` — the interpreter cannot evaluate it: 200,000 characters of `a*` exhausts its fuel budget, the same class as `range_bind_huge` already in the ledger. That length IS the fixture's premise (native aborted from ~30,000 on, and a repeated group exhausted the wasm stack at ~5,000), so shrinking it would delete the thing being pinned. Ledgered with its reason; native, the structural leg and the incumbent all vote, and the run/size ledgers pin the emission.

`cargo test --release --workspace` is otherwise green on this branch (256 suites).

PR #2356, opened 2026-09-20
Exit on the same verdict under check --json as the plain path does

### LINKED ISSUE #2350: almide check --json always exits 0, even when the program has errors
`almide check --json` exits 0 whether or not the program has errors. Plain `almide check` exits 1 correctly. So the machine-readable mode — the one an editor integration, a CI step or an LLM harness actually consumes — is the one that discards its own verdict.

## Measured

```
$ almide check broken.almd           ; echo $?      # errors present
1
$ almide check --json broken.almd    ; echo $?      # same program
0
$ almide check --json clean.almd     ; echo $?
0
```

Pre-existing, not a regression: released `almide 0.62.0` behaves identically (`plain exit=1`, `json exit=0`). Reproduced on `develop` @ `e322037e3` too.

`broken.almd` can be anything the checker rejects, e.g.

```almide
fn main() -> Unit = println(string.to_upper("a", "b"))
```

which reports `E004: string.to_upper() expects 1 argument(s) but got 2` on both paths — and exits 0 under `--json`.

## Why it matters more than an ordinary exit-code slip

The JSON diagnostics are emitted on stdout and the exit code is the only *summary* the mode produces. A consumer that does the normal thing — run the command, check the status, parse the payload only on failure — concludes every program checks clean and never looks at the diagnostics it was given. The failure is silent and looks like success, which is the property that makes it worth more than its size.

It is also the mode furthest from a human: a person reading the terminal sees the errors regardless. Nobody watching is the point of `--json`.

## How it was found

Writing a test for #2345 that asserted a program must NOT check clean. The assertion failed while the JSON in the same output plainly contained the `E004`. The diagnostic was right; the status was not.

That also means **an existing test may be asserting nothing**: `tests/exit_literal_check_test.rs::computed_codes_and_local_functions_remain_checkable` asserts `output.status.success()` over programs that should be accepted, but it invokes plain `check` (not `--json`), so it is fine. Any test that pairs `--json` with `status.success()` would be vacuous — worth a grep before trusting one.

## Suggested shape

`--json` should exit non-zero when it emitted at least one `"level":"error"` record, matching the plain path. Warnings alone stay 0, as they do today.

A test should pin both modes agreeing on the exit code for the same program, so the two paths cannot drift apart again — the plain path being right is what hid this.

O6lvl4:
Closes #2350.

`almide check --json` exited 0 for every program that **parsed**, so a type error was written to stdout as a `"level":"error"` row and then discarded in the status. Plain `almide check` exited 1 for the same program.

The mode that threw its verdict away is the one furthest from a human. A person reading the terminal sees the errors either way; `--json` exists for an editor, a CI step or a model's harness, and that consumer does the normal thing — run the command, gate on the status, parse the payload only on failure. It therefore concluded that **every rejected program checked clean** and never read the diagnostics it had been handed. Silent, and shaped exactly like success.

## The fix

`check_json_one` now returns the same predicate the plain path exits on: any parse error, or any diagnostic at `Level::Error`. Warnings alone stay green, on both paths.

It is deliberately the *same expression* that already gated the unused-variable pass, rather than a second copy of the rule — the status cannot drift from what was emitted, because there is only one test of it.

`fmt --json` has always returned `--check`'s exit code. `check --json` was the one exception.

## The suite was holding the defect in place

`tests/check_package_entries_test.rs::json_judges_every_entry_and_each_row_names_its_file` asserted `ok` on a package with **two** type errors, annotated `"type errors keep the json exit code at 0"`. The test had written the bug down as the rule. That assertion is flipped; everything the case is actually about (one row per diagnostic across every entry, in walk order, each naming its `file`) is unchanged and still asserted.

Worth stating plainly: the reason this survived is not that nobody tested it. It is that what was tested was the behaviour rather than the agreement.

## What the new test asserts

`tests/check_json_exit_code_test.rs` compares **the two modes against each other** over one program, rather than pinning `--json` to expected constants:

- `the_two_modes_agree_on_every_verdict` — four programs (clean, type error, parse error, warning-only), plain vs `--json`, asserting equality first and the expected value second. A test that only pinned `--json` could be made green again by an equal and opposite drift on the plain side; comparing them cannot.
- `a_red_json_check_is_exactly_one_that_emitted_an_error_row` — the status/payload equivalence in both directions. Emitting nothing and exiting 1, or emitting errors and exiting 0 (the bug), each fail exactly one half.
- `a_type_error_under_json_does_not_report_success` — the regression in the issue's own spelling, so a failure names the bug rather than a matrix cell.
- `a_clean_json_check_writes_nothing` — the control: the fix changes the status, not the stream.

## Proof the test catches the bug

Run against the **released 0.62.0 binary**, which has the defect:

```
test a_clean_json_check_writes_nothing ... ok
test a_type_error_under_json_does_not_report_success ... FAILED
test the_two_modes_agree_on_every_verdict ... FAILED
test a_red_json_check_is_exactly_one_that_emitted_an_error_row ... FAILED
test result: FAILED. 1 passed; 3 failed
```

with `the modes disagree on a type error: plain accepted=false, --json accepted=true`. The one test that passes on the old binary is the control, which is what a control is for. A green test is not evidence that it catches anything until it has been run on the broken binary.

## Blast radius

No script, workflow or `tools/` entry gates on `check --json`'s exit code — grepped before changing it. The MCP `almide_check` tool maps to this command; `tests/mcp_test.rs` is in the sweep below and is unaffected.

## Evidence

- New and corrected tests: 10 passed, 0 failed.
- Every test binary that speaks `--json` re-run together: **13 binaries completed, 87 passed, 0 failed** (counted completed binaries, not a quiet tail).
- `almide test`: **446/446 files, 4108 tests, 0 failed** (438 via WASM, 8 via native fallback).
- `check-ratchet-separation`: 4 commits, implementation and verification kept apart.
- `check-llm-surface`: 20 fences, green after the `llms.txt` line.

`docs/specs/cli.md` is corrected in both places that documented the old exit code, and `llms.txt` carries the promise, since that is the surface models read.


O6lvl4:
The `Almide gates` red was mine and is fixed in `Re-record the check manifest exit column the json verdict now moves`.

**What it was.** `scripts/gen-check-manifest.sh` records `sha256<TAB>exit<TAB>path` for `almide check <file> --json` over every `spec/**/*.almd`, so the exit code is a **column of a committed golden**. Changing the verdict moves that column, and the goldens gate regenerates and diffs.

**What moved.** Eight rows, `0 -> 1`, **every sha256 unchanged** — stdout is byte-identical and only the recorded exit code differs. The eight are exactly the negative fixtures:

```
spec/gauntlet/cells/s3_module_qualified_protocol
spec/gauntlet/cells/s4_bare_mut_self.almd
spec/gauntlet/cells/s6_generic_protocol.almd
spec/gauntlet/cells/s7_protocol_as_type.almd
spec/integration/modules/deep_phantom_test.almd
spec/integration/modules/phantom_dep_error_test.almd
spec/integration/modules/vis_local_error_test.almd
spec/integration/modules/vis_mod_error_test.almd
```

Predicted before regenerating: at most 78 rows (the ones emitting any diagnostic). Actual 8 — the other 70 emit warnings only, which stay green on both paths as intended.

The `spec-run-manifest.txt` and `spec-ast-manifest.txt` diffs were the informational `# oracle:` header line alone, with no row changed, so they are restored rather than committed — the gate ignores that line (`-I '^# oracle: '`) and committing header churn would only conflict with the other PRs in flight.

**Worth recording, because it is the same defect seen from the other side.** That exit column was **constant 0 across all 1342 rows** before this change, and `crates/almide-spine/tests/check_parity.rs` reads past it (`let _rc = it.next()`), comparing only the stdout hash. So the golden had been faithfully recording a verdict nothing consumed and nothing could contradict. After this it says something.

**How I missed it in the blast-radius check.** I grepped `scripts/`, `.github/`, `tools/` and the `Makefile` for `check --json` and for the literal `"--json"`, and reported zero consumers. The generator spells it `"$ORACLE" check "$2" --json`, which matches neither. I searched for spellings of the command instead of for the command.

Gates re-run locally after the commit: `parity-goldens OK: ast, check and run manifests equal the CLI's output`, and `ratchet-separation OK: 5 commits`.

O6lvl4:
Rebased over #2347, which re-recorded the parity manifests, and **regenerated rather than merged**.

The conflict was the informational `# oracle:` header line alone. All eight of my data rows applied cleanly — which is exactly the shape that should not be trusted. A clean apply of measured rows says nothing about whether those rows are still true: a textual merge of a generated ledger succeeds with stale numbers, where a conflict would at least have stopped and asked.

So the branch was rebased, the release binary rebuilt from the rebased tree, and the goldens regenerated:

```
parity-goldens OK: ast, check and run manifests equal the CLI's output at cb618ffe7
manifest: 722 files, exclusions: 2
```

The regeneration produced **no diff**, so the eight `0 -> 1` exit-column rows are confirmed unchanged by what landed — by re-measurement, not by the merge having been quiet.

The `# oracle:` header is left at develop's value deliberately. The gate ignores that line (`-I '^# oracle: '`) precisely because it is informational and expected to differ across branches; rewriting it here would guarantee a conflict with every other in-flight PR that touches this file, which is the problem this repo keeps hitting rather than a fix for it.

Issue #2381, opened 2026-09-20
CI's critical path is 55 minutes and five test targets are 64% of it


O6lvl4:
> **状態** — 計測器・ゲート・着手可
> **一行で** — CI の critical path は 55 分で、5 本の cross-target / parity oracle スイートが全テスト時間の 64% を占める。最初の一手は本文の 4→8 shard **ではなく**、コメントの packer emulation が改訂した「`wasm_runtime_interp_ledger` をテスト関数単位で分割してから 4→8」。
>
> **やること**
> 1. `wasm_runtime_interp_ledger` を test 関数単位で 2 target に分ける。`interp_abstain_ledger`（`tests/wasm_runtime_interp_ledger.rs:53`）と `interp_bridge_fallback_ledger`（`:249`）は既に独立した `#[test]` なので、corpus は触らず機械的に分けられる。
> 2. 測定済みの `scripts/ci-test-weights.txt` 更新を入れる（リリース切り出し中は意図的に保留、その直後に投入可）。
> 3. 4 → 8 shard を、各段が単独で merge できる 3 段で: 8 本を 4 本の**横に足す** → required status contexts を `/4` 名から `/8` 名へ切替（open PR が 1 本も無いときに）→ 4 本を外す。branch protection が `Test Rust (shard N/4)` を逐語で名指ししているので、単一 PR ではどちら向きにも deadlock する。
> 4. 2230s より下は `wasm_opt_parity_spec`（644 ケースの corpus を 1 テストが歩く）の分割で、別の種類の作業 — 別 issue に。
>
> **完了条件** — critical path が 4212s → 2230s（−47%）になり、balance skew 1.07 で balance gate と `scripts/refresh-ci-test-weights.sh --check` が自身の条件で緑。
> **根拠** — 本文の実測（green run 35506988245、16,842 test-seconds / 98 target、上位 5 本で 64%、shard balance 4644s/3614s = 1.29）とコメントの packer emulation（refresh 済み weights で n=4 は skew 1.00 / 4212s、n=8 as-is は skew 1.69 で 1.4 ゲート超過、heaviest を 2 分割すると n=8 で 2230s / skew 1.07）。ツリー側は未着手 — `.github/workflows/ci.yml:351` が `shard: [0, 1, 2, 3]`、`tests/wasm_runtime_interp_ledger.rs` は 1 target のまま（2026-09-21 確認）。
[...]

PR #2386, opened 2026-09-20
Re-measure the shard weights from the run that carries the in-place append

O6lvl4:
The weekly balance ratchet is red on the post-#2352 tree, and the refresh is the documented fix. Measured, not projected.

## What the ratchet says

`bash scripts/refresh-ci-test-weights.sh --check 35520781152` against develop `e39133dc5` (run completed, success):

```
shard 106105119807: 4445s    shard 106105119810: 4741s
shard 106105119844: 2614s    shard 106105119904: 4444s
shard skew 1.81 exceeds 1.4 — the weights went stale
```

Three shards cluster at 4.4–4.7 ks and one finishes at 2.6 ks, so a quarter of the matrix idles while the critical path waits on the other three.

## What this PR does

`bash scripts/refresh-ci-test-weights.sh 35520781152` — re-records every target at or above the 5 s floor from that run's own four shard logs. **94 targets written (was 103); measured total 17,694 s → 15,964 s.** The five heaviest wasm targets:

| target | recorded | measured now |
|---|---:|---:|
| `wasm_runtime_opt_parity` | 2,176 | **1,318 (−39 %)** |
| `wasm_runtime_interp_ledger` | 3,129 | 3,315 (+6 %) |
| `wasm_runtime_cross_target` | 2,111 | 2,021 (−4 %) |
| `wasm_runtime_interp_oracle` | 1,920 | 2,004 (+4 %) |
| `run_parity` | 1,961 | 1,849 (−6 %) |

## Two things stated as unknown rather than dressed up

**`opt_parity` at −39 % is unexplained.** A smaller module after #2368's pool dedup would give the wasm-opt parity pass less to do, but that is a hypothesis with no measurement behind it. It is left in the table as an unexplained mover; if it repeats on the next two develop runs it deserves a cause, and if it does not it was variance.

**Whether these weights fix the skew cannot be known from this PR.** The packer's arithmetic can be re-run over the new numbers and will of course produce a balanced packing — that is a simulation of the packer, not a measurement of CI. The prior refresh (#2357) projected `[4694, 4694, 4693, 4693]`, skew 1.00, and the measurement above came back 1.81. So the only thing that settles it is the four shard totals of the next develop run after this lands, and those will be reported then.

## The finding behind the finding

The useful conclusion is not a shard count. **The packer's arithmetic is correct and its inputs move faster than a refresh cycle** — one target swung 39 % between two develop runs, and the weights file is a snapshot with no expiry. That says the durable fix is either a packer that tolerates stale inputs or a suite whose per-target costs do not swing that far, not "refresh more carefully". Recorded for #2381's lever list.

One correction to an earlier claim about this file, since it was repeated in two places: these weights were **not** inflated by the oversized fixture that #2352 shrank. That fixture only ever existed on the PR branch — it is in develop's history but was never in a tree a develop CI run built, because the merge queue runs the final head. A commit being an ancestor of develop does not mean CI ever measured that tree.


O6lvl4:
## Outcome, measured after landing: this did not shorten CI. Negative result.

This PR's body argued from a measured 1.81 skew that re-measuring the weights would flatten the four shards. **It did not.** Per-shard job wall times (`started_at`/`completed_at` from the Actions API), four develop pushes on the old weights against the two runs that carry the new ones:

| run | head | shard 0 | 1 | 2 | 3 | max | max/min |
|---|---|---|---|---|---|---|---|
| push, old | e39133dc5 | **57.6** | 24.8 | 44.1 | 36.9 | 57.6 | 2.32 |
| push, old | 636a9f771 | **57.0** | 41.4 | 46.2 | 36.5 | 57.0 | 1.56 |
| push, old | e8c95e285 | **59.0** | 24.1 | 45.8 | 31.6 | 59.0 | 2.45 |
| push, old | 824ba091f | **58.3** | 42.4 | 33.0 | 36.6 | 58.3 | 1.77 |
| merge_group carrying this PR | 281f23135 | **60.0** | 42.0 | 57.4 | 41.1 | 60.0 | 1.46 |
| push, new weights | 281f23135 | **60.0** | 25.6 | 41.1 | 27.9 | 60.0 | 2.34 |

Shard 0 is the critical path in every run, 57.0–59.0 before and 60.0 twice after. The spread does not narrow: 1.46 and 2.34 after, against 1.56–2.45 before. The effect of re-weighting is inside the run-to-run noise, and the critical path moved ~3% the wrong way.

### Why re-weighting cannot work here

Shard 0's hour is **one step**: `download-artifact` 0.6 min + `Cargo tests` **58.9 min**. Not a timeout cap — the step completed and the job succeeded. Weights assign whole test targets to shards, so once one shard's targets sum to a single 59-minute step, no assignment shortens the critical path. Shortening it means splitting targets or reducing what the slowest ones do, which is #2381, not this.

What this PR still buys: the weights file now reflects measured per-target costs rather than stale ones, which is the input #2381 needs and must **regenerate** whenever it changes the target list or the shard count. It is not a CI speedup and should not be cited as one.

### Caveat on the method, because it nearly produced the opposite conclusion

My first "before" candidate was the develop push for `ac10929aa`, whose shard jobs ran **3–4 seconds**: a docs-only change, where the `changes` job no-ops the matrix by design. Comparing that against a 60-minute shard would have shown a ~1000× "regression". Any before/after on this matrix must first confirm both sides actually ran the tests.

Unmeasured, stated as such: the per-target breakdown inside the 58.9-minute step. `actions/jobs/<id>/logs` returned 153 KB ending at 06:48 for a job that ran to 07:46 — setup steps only.

PR #2399, opened 2026-09-21
Guard the wasm value.keys walk on the Value tag

### LINKED ISSUE #2395: value.keys reads the Value payload slot without checking the tag on wasm, so a non-object Value walks arbitrary memory
> **状態** — BLOCKER・PR #2399 在飛行（merge queue position 1、AWAITING_CHECKS）
> **一行で** — wasm レッグの `value.keys` が Value の payload slot を **tag を見ずに** pairs-list handle として読むので、非 object の Value に対して任意のメモリを歩く（trap するか黙って誤った list を返すかはメモリの中身で決まる）。native は非 object すべてに `[]` を返す。
>
> **やること**
> 1. PR **#2399**「Guard the wasm value.keys walk on the Value tag」をレビューして着地させる（queue position 1）。
> 2. 着地後、本文の 2 行の再現を両レッグで走らせる（`value.str(48 バイト)` が両方 `len=0`）。単独では `value.int(4000000000)` も trap するので、その升も見る。`value.keys(value.null())` が通るのは heap の状態次第なので、clean を証拠として読まない（本文の「Correction to the original diagnosis」）。
> 3. 修正に `spec/wasm_cross/` の fixture を 1 本付け、`// @contract: C-NNN` を宣言する（新 ID なら almide/als に先行着地、既存 ID の再利用なら als PR は不要）。
> 4. 台帳 6 枚を再生成する（マニフェスト → size/alloc の順。マニフェストが corpus なので逆順だと新 fixture が不可視のまま緑になる）。
> 5. 兄弟 3 本は台帳が衝突するので着地順を守る: #2399 → #2404（#2398）→ #2409（#2397）。
>
> **完了条件** — 2 行の再現が両レッグでバイト一致し、`spec/wasm_cross` の新 fixture が `wasm_cross_target_spec` で equal に数えられ、`bash scripts/count-release-blockers.sh` がこの issue を数えなくなる。
> **根拠** — 修正 PR **#2399** は OPEN / merge queue position 1 / AWAITING_CHECKS、自身の check は 27 SUCCESS・4 SKIPPED で赤なし（2026-09-21 確認）。機構は `crates/almide-wasm/src/value_helpers.rs:303` の `emit_value_keys_helper` — このファイルで `m_tag` を bind せずに `m_pay` だけを読む唯一の helper — と `crates/almide-wasm/src/value.rs:371` の `emit_value_box`（tag だけ書いて SUM_FIELD を書かない）。**0.62.0 でも再現**（公開 macOS asset、sha256 照合済）なので最近の PR 起因ではない。
> **触る場所** — `crates/almide-wasm/src/value_helpers.rs`、`crates/almide-wasm/src/value.rs`、`spec/wasm_cross/`、`crates/almide-wasm/tests/golden/`

---
On the wasm leg `value.keys` reads the Value's payload slot as a pairs-list
handle **whatever the tag says**. For every non-object Value that slot holds
something else, and the helper walks it as a block: an `Int`'s payload as an
address, a `Str`'s payload as a pairs list, and — worst — `Null`'s slot, which
is never written at all, so it holds whatever the reused block last contained.

Native answers `[]` for every non-object. The wasm leg traps, or silently
returns a wrong list, depending on what is in memory.

```almide
fn main() -> Unit = {
  let v = value.str("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
  println("len=${list.len(value.keys(v))}")
}
```

```
native:  len=0
wasm:    wasm trap: out of bounds memory access     exit 134
```

No `json`, no `Result`, no `match`, no `??`. Two lines.

## Divergence matrix — `value.*` readers × Value tag

Each cell is the Value produced in the err arm of a `match` on a failing
`Result[Value, String]` (a construction that leaves the heap dirty, so the
stale-slot cell is reached deterministically rather than by luck).

| reader | null | bool | int | float | str | array |
|---|---|---|---|---|---|---|
| **keys** | **✗ trap** | = | = | = | = | = |
| field | = | = | = | = | = | = |
| merge | = | = | = | = | = | = |
| pick | = | = | = | = | = | = |
| omit | = | = | = | = | = | = |
| to_camel_case | = | = | = | = | = | = |
| stringify | = | = | = | = | = | = |

`keys` is the only unguarded reader, and the null cell is only the cell that
traps *in that program*. Standalone, with no dirty heap, `value.int(4000000000)`
and a 48-byte `value.str` both trap and `value.null()` does not. Whether a given
cell traps or silently lies is a property of the heap, not of the program.

## Mechanism

`crates/almide-wasm/src/value.rs:3` — a wasm Value is a 16-byte block
`[rc][len][cap][tag @SUM_TAG][pad][payload:8B @SUM_FIELD]`.

`emit_value_box(tag, None)` (value.rs:371) allocates the 16 bytes, stores the
tag, and **returns without writing SUM_FIELD** — that is the documented
"tag-only Null".

`emit_value_keys_helper` (value_helpers.rs:303) opened with
`local_get(v).i32_load(m_pay).local_set(v)  // pairs list` and no tag test.
It is the ONLY helper in that file that binds `m_pay` without also binding
`m_tag`; `$value_eq`, `$value_merge`, `$vfield`, `$vjson` and `$vjson_pretty`
all read the tag first.

The fault address is derived, not random:

```
err = "expected null"        first 4 bytes 'expe'  LE u32 = 0x65707865   fault 0x65707871
err = "unexpected char…"     first 4 bytes 'unex'  LE u32 = 0x78656e75   fault 0x78656e81
                                                        diff = 12 = almide_layout::PAYLOAD
```

The helper loaded a word out of a block it had no business reading, used it as a
block address, and added `PAYLOAD` to reach the data.

## Age

Not a 0.63.0 regression. Reproduced on the **published v0.62.0 macOS asset**
(sha256 `f39379ea934f…` verified against `almide-checksums.sha256` before
running): `value.str(48 bytes)` and `value.int(4000000000)` both trap on its
wasm leg, both answer `len=0` on its native leg. It is `I-divergence` and
`I-unsound` regardless of age, so it gates the final tag.

## Correction to the original diagnosis

This issue first read *"it is not the index guard and **not `value.keys`** — the
value produced by the fallback is already wrong when it reaches the first
operation"*, and named `??` in its title. Both were wrong. The Value the
fallback produces is correct — `json.stringify` prints `null` on both legs — and
the row that exonerated `value.keys`, `value.keys(value.null())` returning `1`
on both legs, was clean **by accident**: in that program the stale slot happened
to hold 0. A control that passes because of the heap's state was read as
evidence about the compiler.

## Provenance

Found by the release-gate fuzz run against `v0.63.0-rc3` (run `35540551476`),
`seed=568648823617 index=4643`, a mutation of `spec/wasm_cross/json_many_keys.almd`.
That run delivered 60.1 of 480 planned minutes and published `findings=0`
because 7 of its 8 shards were preempted before uploading; this finding exists
only in the shard logs (#2390).

The reduction was run without the fuzzer in the loop — `almide build` on both
legs, artifacts executed directly, the wasm one under stock `wasmtime`.

O6lvl4:
Closes #2395. A release blocker (`I-unsound`, `I-divergence`) — the final
v0.63.0 tag is held on it.

On the wasm leg `value.keys` loaded the Value's payload slot as a pairs-list
handle **whatever the tag said**, and walked it as a block.

```almide
fn main() -> Unit = {
  let v = value.str("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
  println("len=${list.len(value.keys(v))}")
}
```
```
native:  len=0
wasm:    wasm trap: out of bounds memory access      exit 134
```

Two lines. No `json`, no `Result`, no `match`, no `??` — the shape the issue
was first filed as was incidental, and the original diagnosis in it (which
said in so many words "it is not `value.keys`") was wrong. That is corrected
on the issue, along with why: the control row that exonerated `value.keys`
passed **because of what was in the heap at that moment**, and a control that
is clean by accident was read as evidence about the compiler.

## Mechanism

`crates/almide-wasm/src/value.rs:3` — a wasm Value is a fixed 16-byte block
`[rc][len][cap][tag @SUM_TAG][pad][payload:8B @SUM_FIELD]`. The payload slot
is a pairs-list handle **only** under tag Object:

- under `Int`, an i64 used as a block address
- under `Str`, a string handle walked as a pair array
- under `Null`, nothing at all — `emit_value_box(tag, None)` (value.rs:371)
  allocates the 16 bytes, stores the tag and returns **without writing the
  slot**, so it holds whatever the reused block last contained

Native answers `[]` for every non-object (`value_core`'s `value_keys` halves
the slot-count word, 0 for all of them). The wasm leg either trapped or
returned a wrong list, and **which of the two is a property of the heap, not
of the program**: standalone, the wide Int and the long Str trap while
`value.null()` does not; through a failing `Result[Value, String]` match,
Null's unwritten slot aliases the dead error string and the fault address is
exactly `u32_le(the error text's first four bytes) + PAYLOAD`.

`$vkeys` now tests the tag and returns an empty list for a non-object. The
object walk is byte-unchanged.

## Not a 0.63.0 regression

Reproduced on the **published v0.62.0 macOS asset**, sha256
`f39379ea934f6b989ce6e008536979a638929d7d969b52c5b1c8c79d10cd2494` verified
against `almide-checksums.sha256` before running. (A `--version` string does
not identify a build; the asset does.) It gates the final tag regardless of
age.

## Evidence, in two halves that cover each other's blind spot

**Behavioural** — `spec/wasm_cross/value_keys_non_object.almd` (`@contract: C-122`):
every `value.*` op that walks an object's pairs, over every tag, standalone
AND through a dirty-heap construction that makes the stale-slot cell
deterministic rather than lucky. Run against the v0.62.0 asset it dies at the
`int-wide` row; on this branch both legs are byte-identical.

**Source-shape** — `tests/value_helper_reads_its_tag_test.rs`: a function that
LOADS `SUM_FIELD` must also read `SUM_TAG`, crate-wide. `CALLER_GUARANTEED` is
empty.

Three ways that gate could have been built wrong, each found by almide-ee's
independent sweep of the crate and each now a test rather than a comment:

1. Keying on `i32_load(slot_memarg(..SUM_FIELD))` finds **one** site and it is
   not this bug — the real code binds the memarg to a local first. The gate
   follows the binding, and `the_gate_fires_on_the_pre_fix_body` holds it to
   the pre-fix source.
2. `SUM_TAG` is `0`, and `slot_memarg(0)` is all over the crate for slot-0
   reads that are **not** tags — including `emit_value_keys_helper`'s own
   pair-key read. Accepting a bare `0` would have exonerated the very function
   the gate exists for, so only the named constant counts, and the two real
   tag reads spelled `slot_memarg(0)` (`lower_json_to_map`) were normalised to
   `slot_memarg(almide_layout::SUM_TAG)` — `SUM_TAG == 0`, so identical bytes,
   and a guard a tool can see.
3. A regex segmenter sweeps the NEXT function's doc comment into the previous
   body and invents a member. Segmentation is brace depth from the `fn` line,
   with comments and string literals stripped.

The A/B was run on the real file, not only the embedded copy: guard removed,
crate-wide test names exactly `value_helpers.rs::emit_value_keys_helper`,
guard restored.

## Stated omission

The family matrix covers six of the eight `value.*` ops that walk pairs.
`value.to_camel_case` / `value.to_snake_case` are **deliberately** absent: the
als reference evaluator has no arm for either, so reaching them from the
corpus raises its shrink-only abstain ceiling, and writing arms into the
normative judge under release pressure is a worse risk than not covering them.
Both are already guarded by the same tag test in `value_utils` — what is
missing is third-oracle coverage, not a guard. Recorded in the fixture's
`COMPLETENESS RULE` comment, in the C-122 statement, and as almide/als#77 with
the byte-exact port source.

## Ledger

- C-122 EXTENDED to the non-object tag space — **landed in almide/als#76
  first** (merged), pin advanced to `316dc89` in its own commit
- allocation ledger, both size baselines, and the run/check/ast manifests
  ratified in their own commit. The size move is +16 B on exactly the three
  fixtures that emit `$vkeys` (`codec_variant_roundtrip`, `json_many_keys`,
  `value_object_ops_share`) — one guard, once per module, which is also a
  locality check on the change

Local: `almide test` 446/446 files, 4108 tests; `alloc_ledger`, `size_ratchet`,
`gauntlet`, `backend_parity`, `surface_matrix`, `witness_floor`,
`check-contracts.sh`, `check-als-pin.sh`, `check-ratchet-separation.sh` over
the range — all pass.


O6lvl4:
**Stdlib sweep for the same shape, and why it did not become a second gate.**

The Rust-side gate in this PR asserts that a function loading `SUM_FIELD` also
reads `SUM_TAG`. The stdlib has the same hazard in its own spelling — a walk
driven by the slot-count word at `+8`, which only counts pairs under tag Object
at `+4` — so I swept it by hand. Fourteen `+8` reads across
`stdlib/value_core.almd` and `stdlib/value_utils.almd`:

| site | guarded by |
|---|---|
| `value_keys:238` | **was unguarded — this PR** |
| `value_eq:517` | `if ta == 6` above it |
| `value_merge:714-715` | `if load32(ah+4) == 6 and load32(bh+4) == 6` |
| `value_field:751` | `if tag != 6 then err(...)` |
| `__vstr:998-999`, `__vdrop:1028-1029`, `__vstr_pretty:1284/1288` | inside `tag == 5` / `tag == 6` arms |
| `value_utils:47/153/164` | the callers `value_pick` / `value_omit` / `value_to_camel_case` / `value_to_snake_case` each test `+4 == 6` first |

One instance, and it is the one this PR fixes.

**I did not land a second gate for it, and that is a decision rather than an
omission.** Three of the fourteen sites (`__vu_sel`, `__vu_ren_c`,
`__vu_ren_s`) are caller-guaranteed, so a rule of the form "a fn reading `+8`
also reads `+4`" needs an allowlist on day one. The Rust gate avoided exactly
that by keying on *loads through the payload memarg* rather than on mentions,
which let `CALLER_GUARANTEED` stay empty — there is no equivalent structural
discriminator in the `.almd` source, where the guard genuinely lives in a
different function.

A gate that ships with a hand-maintained exception list for a
single-instance family is the shape that drifts. If a second stdlib instance
ever appears, that is the evidence for building one — with the population
problem solved first.

PR #2409, opened 2026-09-21
Give the fused fold's callback result the share guard its staged twin has, so fold over a map or filter chain no longer returns a freed element on wasm

### LINKED ISSUE #2397: string.chars through list.filter into list.fold returns freed memory on wasm: a three-line program prints the wrong byte
> **状態** — BLOCKER・**PR #2409 在飛行**（2026-09-21 時点 draft、#2404 の着地待ちで意図的に merge queue 外）
> **一行で** — `string.chars` の結果を `list.filter` 経由で `list.fold` に渡すと、wasm レッグが解放済みメモリを返す（trap も診断もなく、もっともらしい誤バイトで継続する）。融合された fold のコールバック結果に、staged 版が持っている share guard が無い。
>
> **やること**
> 1. **PR #2409 をレビューする** — `crates/almide-wasm/src/list_fuse.rs` に share guard を与え、`spec/wasm_cross/list_fold_fused_callback_result_ownership.almd` + 差分テストで固定している。
> 2. **#2404（#2398 の修正）が着地した後に #2409 を queue 投入する。** 順序は必須: #2409 の C-334 statement は judge の全文を写しており、その全文には #2404 の修正を記述する #2398 節が含まれるので、あの木が存在する前に投入すると als-pin ゲートが落ちる。
> 3. 残り 1 本の in-progress チェック（Commissioned mutation gate）が緑になったら draft を解除する。
> 4. 台帳 6 枚は #2409 が既に運んでいる（3 マニフェスト + alloc + size ×2）。rebase 後に再生成が要るなら、マニフェスト → size/alloc の順で回す（マニフェストが corpus なので逆順だと新 fixture が不可視のまま緑になる）。
>
> **完了条件** — 3 行の再現が両レッグでバイト一致し、#2409 の fixture が `wasm_cross_target_spec` で equal に数えられ、`bash scripts/count-release-blockers.sh` がこの issue を数えなくなる。
> **根拠** — 2026-09-21 に A/B 実測: 現ツリー（#2310 込み）と **0.62.0 の両方**で wasm が誤り、native は `r = B` ⇒ #2310 起因ではなく少なくとも 0.62.0 から存在する既存バグ。PR #2409 の checks は同日実測で 26 SUCCESS / 4 SKIPPED / 1 IN_PROGRESS。
> **触る場所** — `crates/almide-wasm/src/list_fuse.rs`、`spec/wasm_cross/list_fold_fused_callback_result_ownership.almd`、`docs/contracts/contracts.toml` + `proofs/als-pin.txt`、`crates/almide-wasm/tests/golden/`

---
A `String` produced by `string.chars`, passed through `list.filter` and consumed by `list.fold`, comes back on the wasm leg as **freed memory**. No trap, no diagnostic — a plausible-looking wrong byte, and the program continues.

## Reproduce — three lines, no large integers, no multi-byte characters

```almide
fn main() -> Unit = {
  let r: String = list.fold(list.filter(string.chars("AB"), ((a) => true)), "X", ((n, y) => y))
  println("r = ${r}")
}
```

```
native:  r = B          'B' = 0x42
wasm:    r =            0x20
```

Deterministic — identical across runs for a given program, so this is a fixed wrong offset or a premature free, not a heap race. The byte changes with the surrounding program (`0x20` here; `0x30` with a different predicate; `0x10` and `0x03 0x01 0x00` in the two fuzz programs that found it), which is what reading whatever now occupies a released block looks like.

## All three combinators are required

Remove any one and it is clean:

| shape | result |
|---|---|
| `fold(filter(chars))` | **DIVERGES** |
| `fold(chars)` — no `filter` | ok |
| `fold(literal)` — no `chars` | ok |
| `fold(filter(literal))` — no `chars` | ok |
| `filter(chars)` then `list.join` — no `fold` | ok |
| `fold(filter(list.map(... int.to_string ...)))` — strings, but not from `chars` | ok |
| `fold(filter(chars))` with a non-trivial predicate | **DIVERGES** (wasm gives `0`) |

So it is specifically a list of heap `String`s **produced by `string.chars`**, reshaped by `filter`, then read by `fold`. A list of strings built by `list.map` through the same pipeline is fine, which points at how `chars` allocates its elements rather than at `filter`/`fold` in general.

## Severity

`I-unsound` as well as `I-divergence`: this is a use-after-free that **returns data** instead of faulting. A trap would at least be loud. This returns a byte that is a perfectly legal `String`, so it flows onward into comparisons, keys, output and files, and the first sign of trouble is arbitrarily far from the cause.

The source program contains nothing unusual — no large integers, no emoji, no deep nesting. `xs |> list.filter(...) |> list.fold(...)` over the characters of a string is ordinary code, and it is close to the idiom `docs/CHEATSHEET.md` teaches for character-level processing.

## Not a 0.63.0 regression

A pre-0.63 development build (2026-09-13) produces the same wrong byte, independently confirmed by a second session building both legs from source and running the artifacts under stock `wasmtime`. The defect is live and predates this release, so it is not a `regression` — it is a blocker because it is unsound, not because it is new.

## Provenance

Found by the release-gate fuzz run against `v0.63.0-rc3` — re-dispatched run `35540551476`, shard 6:

```
** FINDING [OutputDivergence] seed=568648823622 index=4085
   stdout: native="r6 = some(\"😀\")" wasm="r6 = some(\"<mojibake>\")"
```

The original program reached it through an emoji, which made it look like a UTF-8 handling bug. It is not — the emoji was incidental, and the reduction above uses `"AB"`.

That shard was preempted before uploading, so this finding **never reached the run's published count**; the verdict line reads `correctness=0`. Recovered from the job log — see #2390.

**Replaying from the seed needs the tree**, not just `(seed, index)`: the generator draws its corpus from `--repo`, so the same pair yields a different program against a different commit. The three-line source above has no such dependency and is the artifact to use.

## Starting point for whoever takes it

Fuzzing found a **closure-argument use-after-free on the structural leg** on 2026-09-02, fixed by PR #1788. If that is the same mechanism, the fix was narrower than the bug, and that PR is a much better place to start than a cold read of the emitter.

Filed alongside a second, distinct `String`-corruption divergence from the same run (a `String` captured in a closure, corrupted through `map.values`/`dedup`/`update` with **no `string.chars` involved**). They may share a root cause — whoever fixes this should check — but the narrowing above does not demonstrate it, which is why they are two issues.

O6lvl4:
Closes #2397.

## What was wrong

`list.fold` has two lowerings on the structural wasm leg. The staged `lower_list_fold` (`list.rs`) has owned its accumulator's credit since 10fed0494 — `rc_share_guard` on the callback body, dec of the previous accumulator before the rebind. The fused `lower_list_fold_fused` (`list_fuse.rs`), taken when the source is a `list.map` / `list.filter` chain, was written before that rule and never received it. A callback that returns the element — `(n, y) => y` — therefore rebound the accumulator to a handle the list's own release then freed: a wrong byte, no trap, exit 0.

That is why the report needed all three combinators (no chain → staged copy → correct) and why binding the chain to a `let` cured it. It is not a `string.chars` property (`[[1], [2]]` printed `[65888]` for `[2]`), and `fold(map(chars(s), (c) => c), …)` with no `filter` diverged too — a cell the issue's table did not have.

## The fix

The fused copy now applies the staged copy's two lines verbatim (`crates/almide-wasm/src/list_fuse.rs`).

## Evidence

- `spec/wasm_cross/list_fold_fused_callback_result_ownership.almd` (C-334): two-axis matrix — element runtime shape (String / List[Int] / Result / Int) × callback tail form (borrowed / fresh) — both fused stage kinds, the staged-`let` twin, a 3585-element chain, and the `err("boom")` row that printed correctly standalone and **trapped** inside the fixture on 0.62.0. Incumbent leg renders it, no wall.
- `tests/list_fold_fused_callback_result_ownership_test.rs`: builds and runs both legs under stock `wasmtime`, 4/4.
- **A/B on the published v0.62.0 asset**: 4 of 14 cells diverge; on this tree all 14 agree; the fresh, scalar and staged rows are byte-identical on both binaries — the guard is conditional, not an unconditional inc.

## Ledger

C-334 `EXTENDED (#2397, 0.63.0)` — landed in the judge first (almide/als#85, pin advanced to `264fd9c`), statement mirrored byte-for-byte. It records two corrections to the #2398 clause it follows: `fold` was listed clean, and the family was declared closed by a sweep whose predicate ("stores into a heap slot") excluded an accumulator rebind into a local. Re-swept under "lower a callback body, then store its result" (44 sites): `set.fold` / `map.fold` are unguarded but return `Lowered::view`, so a borrowed result is never released by the caller — leak class, clean on both legs, recorded not fixed.

Six spec-fixture ledgers re-recorded in their own commit; ratchet separation clean.

## Precondition on landing order

The mirrored C-334 statement carries the `EXTENDED (#2398)` clause as well, which describes the `list.update` guard that lives in #2404, not here. **This PR must not enter the merge queue until #2404 has merged**, so that every claim in the statement is backed by the tree it lands on. (#2404 deliberately carries the older statement for the mirror-image reason — see its body's last section before "Not a regression".)

An existing fixture moved: `availability_pure_batch.almd` allocates 32 bytes less on the wasm leg and its module is 4 bytes larger — the guard's bytes, and a previous accumulator that is now released each step instead of retained. Its legs are identical before and after.


O6lvl4:

Issue #2128, opened 2026-09-12
Hold the gate language boundary in proofs/gate-verification.toml: a structured gate written in bash must fail CI


O6lvl4:
Follow-up to the roadmap decision recorded in [the gate language boundary](https://github.com/almide/almide/blob/develop/docs/roadmap/active/zero-committed-shell.md) (rewritten 2026-09-12).

## The decision this enforces

> Gates that read structured data (TOML / markdown / ledgers / generated-output comparison) are written in Almide. Gates that grep and exit stay bash.

The reason is a cost asymmetry, not taste. A bash gate touches 3 places (file, CI step, ledger row) and reuses nothing — the 60 `scripts/check-*.sh` each re-derive their own `grep`/`awk`/`sed`. The Almide side already paid for `tools/almide-gates/src/toml/mod.almd` (291 lines) and `src/mdmeta/mod.almd` (68 lines), so the next structured gate gets a typed parser free. All 8 defects Unit 0.46 found while porting were in that structured/decision-rule half.

## Why a rule alone will not hold

It already failed once. The previous goal ("no committed `.sh`") shipped with a done-criterion reading "a gate fails CI when the count goes up". It was never implemented, and in six weeks committed shell went 50 → 124 files, **50 of the 81 additions being `scripts/check-*.sh`**. Nobody noticed because nothing measured it.

## What to build

`proofs/gate-verification.toml` already carries one row per gate with `path` / `class` / `evidence`, enforced by `scripts/check-gate-verification.sh`, with a shrink-only ceiling on `UNVERIFIED`. Add a second axis to the same ledger:

- A field declaring what the gate reads — structured vs. scalar.
- The check fails when a row declares structured and the `path` is a `.sh`.
- Initial value `UNCLASSIFIED`, ceiling 79, **shrink-only** — the same honest-debt idiom the ledger already uses for `UNVERIFIED`, so classifying 79 existing rows does not block the mechanism from landing.
- New gates must declare.

## Also needed

`scripts/check-gate-verification.sh` enumerates `scripts/check-*.sh` only, so **Almide gates do not exist in the ledger at all**. Extend the enumeration to `tools/almide-gates/` subcommands, or an Almide gate can never carry a verification class.

## Explicitly not wanted

A ratchet on the raw `.sh` count. Under this boundary a grep-and-exit gate in bash is the correct choice, so the count rises legitimately. It is no longer a signal.

PR #2413, opened 2026-09-21
Pin the implementation count of every stdlib hole and refuse a copy that grows it without a ledger row

O6lvl4:
Step 1 of the revised plan in #2412 (measure → ratchet → fold), landable before the 0.63.0 final because it changes no behaviour.

A stdlib hole (`fn f(...) = _`) is one operation whose body a backend supplies. Three of the four defects the rc3 soak found (#2395, #2397, #2398) were the same shape: one operation implemented several times — native intrinsic, hand-written structural-wasm lowerings, shape-monomorphic Almide copies on the prim floor — and a rule landing in one copy never reaching the others. A fixture cannot catch the copy that does not take its shape; a sweep keyed on a guard's spelling cannot see the copy that predates the spelling. The count can.

`scripts/lib/impl-count-enumerate.py` re-derives, per hole, `(native, emitter, selfhost)` from source; `proofs/impl-count-ledger.toml` pins the triple; `scripts/check-impl-count.sh` fails on a total that grew (a new copy must add its row in the same change, so it is a diff line and not a blocker nine weeks later) and on a total that shrank without `UPDATE=1` (the ledger never sits above reality in the flattering direction).

Measured today: **702 holes, 1,698 implementations**. Worst: `list.sort` 16, `list.take` 14, `map.get` 14, `list.drop` 13, `map.from_list` 12, `list.get` 12, `list.fold` 11. Distribution of totals: 1×168, 2×341, 3×102, 4×38, 5×13, 6×17, 7×7, 8×2, 9×4, 10×3, 11×1, 12×2, 13×1, 14×2, 16×1.

Proved to fire: appending a `fn lower_list_fold_extra()` to `list_fuse.rs` fails the gate with `list.fold: total 11 -> 12 GREW (emitter 3->4)`. Wired into the CI `checks` job and lefthook (globbed over the three sources it counts).

Column definitions are heuristic by construction (a `fn lower_<mod>_<fn>[_suffix]` definition, a registry call name `<mod>.<fn>[_shape]`, an `@intrinsic` within three lines above the hole) and the script header says so; the ledger is a pinned measurement of those definitions, and every fold in the plan lowers a row.

## Revision: ported to Almide

CI's gate-verification ledger refused the first cut — a `.sh` that reads a TOML ledger must be Almide, where the typed readers live (#2128). So the gate is now `tools/almide-gates impl-count`: `scripts/lib/impl-count-enumerate.py` only counts (one TSV row per hole), the Almide half reads `proofs/impl-count-ledger.toml` with the existing `toml` module, compares, and is the ledger's only writer (`--update`). Wired in the almide-gates CI job and lefthook. `proofs/gate-verification.toml` classifies it NEGATIVE_TESTED with four forged inputs replayed through the Almide gate (a grown total, a new hole, a ledger above reality, a blinded enumerator — the last one prints 702 rows of zeros and reads as 99 drifts, and an enumerator printing under 100 rows is refused outright). Dogfood tests 14/14.


O6lvl4:
Folded commit for commit into the fourth A-ci batch, #2508, which is judged as one tree (the `-x` trailers there name these commits). The previous owner of this PR confirmed it is unclaimed. Closing here; the issue closes with the batch.

PR #2425, opened 2026-09-21
Add the stdlib-linking wasm size ladder: fourteen in-tree programs from hello to regex with a shrink-only ceiling per rung

### LINKED ISSUE #2141: Emitted wasm size is measured by five programs that all link almost nothing, and nothing ratchets it
> **状態** — 計測器・ゲート・着手可
> **一行で** — 吐いた wasm のサイズには ratchet が一つも無く、corpus の 5 本はどれもほぼ何もリンクしないので、公開されている唯一の行（hello）は集合の中で最も有利な 1 点である。
>
> **やること**
> 1. `scripts/check-wasm-size.sh` + `wasm-size-baseline.txt` を、既存の `scripts/check-embedded-size.sh` + `scripts/embedded-size-baseline.txt` と同じ形（1 プログラム 1 行、shrink-only、上げるなら理由をコミットメッセージに）で作る。
> 2. corpus に段階的に stdlib を引く tier を足す: T3 = String 構築・補間・`string.*`、T4 = `Map` / `Set`、T5 = JSON parse + serialize。現状は T0 が 1 本、T1 が 3 本、T2 が 1 本、T3–T5 が 0 本。
> 3. baseline の数値は初回実測で刻印する（本文の指定: T3–T5 は未測定なので、いま書く数は推測になる）。
> 4. `docs/benchmarks/wasm-size.txt` を hello 1 行から tier ごとの行に広げ、`scripts/gen-readme-stats.sh` → `scripts/check-readme-numbers.sh` の既存経路で README のテーブルも tier 行にする（コメントの後半）。
> 5. #2098 と噛み合わせる: perf corpus の少なくとも 1 本に、命令型と docs 慣用の両方の綴りを置く。
>
> **完了条件** — `scripts/check-wasm-size.sh` が CI から走り、`wasm-size-baseline.txt` に T0–T5 各 tier 1 行以上が刻印済みで、行を上げる変更が理由なしには通らない。README の `wasm-size:generated` ブロックが 1 行ではなく tier ごとの行を描く。
> **根拠** — ツリー確認(2026-09-21): `scripts/check-wasm-size.sh` は存在しない。`research/benchmark/perf/wasm-size/` は `hello/fizzbuzz/fibonacci/closure/variant` の 5 本と `measure.sh` だけで、`measure.sh` を参照する `scripts/check-*.sh` も workflow も 0 件。`docs/benchmarks/wasm-size.txt` は 12 行中データ 1 行（`structural_bytes = 1330` / `incumbent_bytes = 1096`）で、README.md:145–151 がそこから 1 行のテーブルを描いている。本文の実測(2026-09-13, develop @ 51b9e80d7): hello 1,330 / fib 2,033 / binarytrees 2,833 / spectralnorm 11,650。
> **触る場所** — `scripts/check-wasm-size.sh`（新規）、`research/benchmark/perf/wasm-size/`、`docs/benchmarks/wasm-size.txt`、`scripts/gen-readme-stats.sh`

---
Measured 2026-09-13 against `develop` @ 51b9e80d7, `--target wasm`.

## What the size corpus currently covers

`research/benchmark/perf/wasm-size/measure.sh` builds five programs — `hello`, `fizzbuzz`, `fibonacci`, `closure`, `variant` — and prints a table that `scripts/gen-readme-stats.sh` turns into the README's size numbers. **All five link almost nothing.** Their whole spread is the cost of the prelude plus a handful of functions.

Probes that show where the real spread is (verified / `--wasm-opt`):

| probe | verified | opt | over hello |
|---|---:|---:|---:|
| `hello` | 1,330 | 407 | — |
| print one **Int** | 1,970 | 1,067 | +640 |
| print one **Float** | **5,501** | **3,016** | **+4,171** |
| `list.repeat(1.5, 8)` + `list.len` | 2,383 | 1,285 | +1,053 |
| `math.sqrt(2.0)` | 2,008 | 1,067 | +678 |

And the four benchmark programs:

| program | verified | opt |
|---|---:|---:|
| hello | 1,330 | 407 |
| fib(35) | 2,033 | 1,144 |
| binarytrees | 2,833 | 1,924 |
| **spectralnorm** | **11,650** | **8,585** |

Three of those four are within 1.5 KB of hello. The fourth is 8.8× hello, and the reason is not that it does float *math* — `math.sqrt` is free (it lowers to the `f64.sqrt` instruction) and a float list is a sixth of printing one float. spectralnorm is large because it is **the only program in the corpus that prints a Float** (#2099).

So the corpus has **three observations of the unlinked floor and one of the linked regime**, and every size number the README quotes comes from the floor.

## What is not gated

| what | ratchet |
|---|---|
| the compiler's own embedded stdlib payload | `scripts/check-embedded-size.sh` + `embedded-size-baseline.txt` |
| native runtime vs a handwritten reference | `scripts/check-perf-ratio.sh` + `perf-ratio-baseline.txt` |
| **the size of the wasm we emit** | **nothing** |

`measure.sh` has no baseline file and is referenced by no `check-*.sh` and no workflow. It is a reporting script, not a watch. The embedded-payload ratchet exists because "the 6× growth over four months went unnoticed because nothing watched it"; emitted artifact size is in exactly that position today.

## Proposal — tiers, then a stamped ceiling per tier

Add programs that pull the stdlib in **by degrees**, so the corpus measures the slope and not just the intercept:

| tier | what it links | status |
|---|---|---|
| T0 | nothing — `hello` | covered |
| T1 | Int formatting — `fib`, `fizzbuzz`, `binarytrees` | covered (3 programs, all here) |
| T2 | Float formatting — `spectralnorm` | 1 program |
| T3 | String building, interpolation, `string.*` | **missing** |
| T4 | `Map` / `Set` | **missing** |
| T5 | JSON parse + serialize | **missing** |

Then gate it the way the other two are gated: `scripts/check-wasm-size.sh` + a `wasm-size-baseline.txt` carrying one number per program, **measured once and shrink-only** thereafter. Same shape as `embedded-size-baseline.txt`, same rule — a row may only go down, and raising one is a deliberate commit with a reason in the message.

**Target values are stamped from the first run, not invented here.** T3–T5 have never been measured, so any number written now would be a guess; the deliverable is the mechanism plus the first measurement, and the ceiling is whatever that measurement says.

## The one target that can be stated now

Printing a Float costs **+3,531 B verified / +1,949 B opt** over the byte-identical Int program — **5.5× what printing an Int costs**. Shortest-roundtrip float printing is inherently larger than integer printing, so the floor is not `+640`, but 5.5× is the gap to argue about, and nothing has read the formatter to say what the floor actually is (#2099).

Suggested first target: **halve it** — Float print under **+1,800 B verified**. That alone takes spectralnorm from 11,650 to roughly 9,900 verified and would move every program that prints a computed number.

## Why this ordering

Fixing #2099 improves one number. Adding T3–T5 tells us whether there are three more #2099s waiting in string handling, `Map`, and JSON — which is worth knowing *before* deciding how much the float formatter matters relative to them. **The corpus change is the cheaper of the two and it is the one that makes the other decidable.**

Gating it also closes the loop with #2098: the perf corpus should carry both the imperative and the documented-idiom spelling of at least one program, so a size *or* speed tax on the documented style cannot return unnoticed.

O6lvl4:
Closes #2141 for the ratchet half (the README/docs per-tier rows that need the CLI's `--measure` are left as noted below).

Emitted wasm size was measured by five programs that link almost nothing, and nothing ratcheted it. `crates/almide-wasm/tests/size_ladder.rs` (in-process emitter, release-only, same style as `size_ratchet.rs`) pins the SHIPPED `to_wasi` bytes of fourteen pre-existing programs on a T0..T7 ladder (hello → Int print → Float print → String → List → Map → JSON → regex) plus a total, in `crates/almide-wasm/tests/golden/size-ladder.txt`; above the row is a regression, below is a stale row; the table is printed on every run so a reader sees which rung grew. Seeded from this tree: 191,606 B over fourteen rungs; the two `spec/wasm_cross` rows match `size-baseline-wasi.txt` byte for byte and hello matches `docs/benchmarks/wasm-size.txt`. Both spellings of the #2098 pairs (wordfreq imperative vs `group_by`, listbuild append vs combinator) are rungs. Runs in the wasm gates job right after the carried gates.

Negative control: forging the regex row 20120→20119 turns the test red naming the rung and the +1 B; restored, green. `check-release-only-tests.sh` and `check-ratchet-separation.sh` OK (implementation and ledger in separate commits).

Left: the shell-side `docs/benchmarks/wasm-size.txt` / README per-tier rows via `gen-readme-stats.sh --measure` (needs the CLI), and `--wasm-opt` sizes.

Pre-push hook skipped (LEFTHOOK=0): a new release-only test, its golden, and a ci.yml step; run_parity and the clippy ratchet are unaffected.


O6lvl4:

Issue #2028, opened 2026-09-07
Sequential allocation on the structural leg got 51% slower in the drop-glue arc (82→124 ms), and neither gate covers that shape


O6lvl4:
## Measured regression

A **sequential** allocate-tree-then-drop program on the structural wasm leg, same artifact command, same wasmtime 47.0.2, 15 runs each — the ranges do not overlap:

| tree | median | min | max | size |
|---|---:|---:|---:|---:|
| `1dd722361` (09-07, before the drop-glue arc) | **82 ms** | 78 | 88 | 2,732 B |
| `538f3c177` (09-07 22:59, after it) | **124 ms** | 117 | 132 | 2,854 B |

**+51 % wall clock, +122 bytes.** Native is unchanged in the same session (234 → 238 ms, noise), so this is the wasm leg specifically.

The program:

```almide
type Tree = Leaf | Node(Tree, Tree)
fn make(depth: Int) -> Tree =
  if depth == 0 then Leaf else Node(make(depth - 1), make(depth - 1))
fn check(t: Tree) -> Int = match t { Leaf => 1, Node(l, r) => check(l) + check(r) + 1 }
fn main() -> Unit = {
  var total = 0
  var i = 0
  while i < 12 { total = total + check(make(19)); i = i + 1 }
  println("${total}")
}
```

Output is correct on both trees (`12582900`), and `ALMIDE_VERIFIED_DEBUG=1` confirms the structural leg still emits it — so this is a slowdown, not a routing change.

The suspect is the #2010 arc: "Records and variants drop and copy through typed shape glue" (3086388a3) makes a `Node(Tree, Tree)` variant droppable, which is exactly this program's hot path. That arc is correct work — the graveyard needed emptying — but the region window (#1961) is supposed to make this shape's drops vanish wholesale, and something now runs per node inside or around it.

## Why no gate caught it

Both paths are open by design, and each reason is defensible on its own:

1. **CI never runs the ratio verdict.** `check-wasm-runtime-ratio.sh` sets `RATIO_VERDICT=0` when `GITHUB_ACTIONS` is set (header: cross-engine ratios do not cancel hardware). On CI it enforces only the *status taxonomy* — a measured row still benches, a walled row still walls. A row that benches 51 % slower still benches.
[...]

PR #2427, opened 2026-09-21
Judge the wasm runtime leg in CI by a same-runner A/B against the latest release binary

### LINKED ISSUE #2143: The wasm runtime ratchet does not judge in CI: RATIO_VERDICT=0 under GITHUB_ACTIONS, so nothing automated watches wasm runtime regressions
> **状態** — 設計判断・判断待ち
> **一行で** — `scripts/check-wasm-runtime-ratio.sh` は CI では比を測って印字するだけ（`GITHUB_ACTIONS` 下で `RATIO_VERDICT=0`）なので、wasm の実行時退行を自動で見る watch が 1 つも無い。
>
> **やること**
> 1. 本文の 4 案 — ① 同一 run 内の native を基準に取る ② 3× 程度の広い絶対帯 ③ このジョブだけ固定クラス runner ④ per-PR ではなく nightly — から 1 つを選び、選んだ理由を `scripts/check-wasm-runtime-ratio.sh` のヘッダに書き足す。ヘッダの既存の理由（cross-engine 比は機械を打ち消さない）は正しいので、それを前提にした選択になる。
> 2. 選んだ形を実装し、`docs/benchmarks/wasm-runtime.txt` の 17 行に対して CI で判定が出るようにする。
> 3. 選んだ判定が #2028 の退行（structural レッグ 82 → 124 ms、+51 %）を捕まえるかを後追いで確認する。本文はこれを「verdict が off であることが load-bearing だった既知の 1 件」として挙げている。
>
> **完了条件** — `.github/workflows/ci.yml:1772` の実行が wasm 実行時の退行に対して非ゼロで落ちる（④ を採るなら毎日人が読む場所に判定が届く）ことと、#2028 の +51 % をその判定が捕まえると示せること。どの案も採らないと決めるなら、その理由をスクリプトのヘッダに記録して close するのも完了。
> **根拠** — ツリー確認(2026-09-21): `scripts/check-wasm-runtime-ratio.sh:111-113` が `WASM_RUNTIME_RATIO_VERDICT` 未設定かつ `GITHUB_ACTIONS` で `RATIO_VERDICT=0`、145-147 行で `informational on this machine class` と印字して `continue`。`docs/benchmarks/wasm-runtime.txt` は 17 行。呼び出しは `.github/workflows/ci.yml:1772` の 1 箇所のみ。無効化の理由はヘッダ 10-16 行に実測つきで書かれている（2026-08-31、nbody wasm/native が刻印機 2.09 に対し 2 コア runner 19.95）。
> **触る場所** — `scripts/check-wasm-runtime-ratio.sh`、`docs/benchmarks/wasm-runtime.txt`、`.github/workflows/ci.yml`

---
Surveyed 2026-09-13 against `develop` @ 51b9e80d7.

## What the gate does in CI

`scripts/check-wasm-runtime-ratio.sh` is headed **"WASM RUNTIME RATIO RATCHET (#1701)"** and gates the 17 rows of `docs/benchmarks/wasm-runtime.txt`. Line 111-112:

```bash
if   [ -n "${WASM_RUNTIME_RATIO_VERDICT:-}" ]; then RATIO_VERDICT=1
elif [ -n "${GITHUB_ACTIONS:-}" ];            then RATIO_VERDICT=0
```

**In CI the ratio is measured and printed, and nothing fails on it.** What CI enforces is that each row still *benches* — that the program builds and runs on both legs. The ratchet half of the name is a local-only behaviour.

## Why it is that way, and why it is still a hole

The header states the reason and the reason is sound: a cross-engine ratio does not cancel the machine the way the native/Rust one does. Measured 2026-08-31, `nbody` wasm/native was **2.09 on the stamping M-series and 19.95 on a 2-core ubuntu runner**. Turning the verdict on in CI would produce a gate that fails on runner class rather than on regressions.

So this is not a gate that was quietly disarmed — it is honestly disarmed, with the measurement that justifies it in the header. The hole is what follows from that: **there is currently no automated watch on wasm runtime regressions at all**, and the verdict being off is load-bearing in at least one known case. #2028 — the structural leg going 82 → 124 ms, +51 % — went through, and its own postmortem names this as one of the two reasons nothing caught it.

That regression was found by someone re-measuring by hand, twelve days later.

## Directions, none of them obviously right

1. **Ratio against a same-run native baseline instead of a stamped one.** Both legs run on the same runner in the same job, so the machine cancels within the pair even if the absolute numbers move 10×. This is what the native gate already does against handwritten Rust (`perf-ratio-baseline.txt` + the `ablation/` twins from #1466); the wasm equivalent would be wasm-vs-native measured together rather than wasm-vs-stamped-wasm.
2. **A wide absolute band** — fail only past, say, 3× the stamped row. Catches a 51 % regression on a fast runner, silent on a slow one. Crude, cheap, strictly better than no verdict.
3. **A dedicated self-hosted or pinned-class runner for this one job**, so the stamped comparison is valid. Most faithful, most infrastructure.
4. **Nightly rather than per-PR**, where a slow runner matters less and a human reads the delta. Turns a twelve-day gap into a one-day gap without pretending a cross-engine ratio is machine-independent.

I have no basis for choosing; the header's author has thought about this more than this issue has. The ask is that the choice gets made and recorded, because the current state reads as a ratchet in the filename and in the CI job list while enforcing nothing on the axis it names.

## What this is not

- **Not about which programs are watched** — that is #2142 for the native ratchet.
- **Not about emitted artifact size** — that is #2141.
- **Not a claim that the header is wrong.** It is right about cross-engine ratios. This is about what to do given that it is right.

O6lvl4:
Closes #2143.

The wasm runtime ratchet set `RATIO_VERDICT=0` under `GITHUB_ACTIONS` because the stamped wasm/native ratio does not transfer between runner classes (nbody 2.09 stamped vs 19.95 on a runner), so nothing automated watched wasm runtime regressions. The 設計判断 is taken as the shape #2028's postmortem named: gate the relative change between two builds on the SAME runner. The step downloads the latest release binary (`releases/latest`, so rc prereleases are excluded), runs two interleaved rounds (release, tree, release, tree) of `almide bench --target wasm --runs 5` per ledger row, takes the min over runs per binary (shared-runner load only adds time), and fails when tree_min/release_min exceeds 1 + band. The band is a policy line in `docs/benchmarks/wasm-runtime.txt` (`ab_band = 30`, fan rows override on the row, `binarytrees … ab_band=100`); a faster tree is printed, never failed; absolute numbers and the stamped cross-engine ratio stay informational. It cannot quietly disarm: under `GITHUB_ACTIONS` the script exits 2 without a baseline unless the step sets `WASM_RUNTIME_AB=off` on record, and a failed download fails the step. The four options and the reason for this one are in the script header, the ledger header and the README block.

Measured locally on a loaded machine (load 29–38) against the 0.62.0 binary: ten rows judged, ratios 0.61–1.22, exit 0; a second run put the 1.22 (listbuild_append) at 1.00, which is the noise the min statistic is for. Negative controls: band forced to −50 → every 30%-band row HIGH and exit 1 (binarytrees stays OK under its row band); no baseline in CI → exit 2; non-executable baseline → exit 2. `check-workflows.sh` green.

Left: a measured `binarytrees_seq` row so #2028's exact shape is watched (stamping needs a quiet machine); the band (30%) is under #2028's +51%, so a same-shape regression on any current row is caught.

Pre-push hook skipped (LEFTHOOK=0): a script, a ledger text file, ci.yml and the README block.


O6lvl4:
Rebased onto develop after #2444 (the commissioned job now consumes the build artifact); the A/B step keeps ALMIDE_BIN=/tmp/almide-bin/almide. Pushed with LEFTHOOK=0: workflow, script and docs only, no compiler source.

O6lvl4:

PR #2428, opened 2026-09-21
Classify every target-availability row by closure class and gate the classes

O6lvl4:
Advances #1423 (the header's remainder: the class vocabulary and the two bucket-C rulings; the pending-self-host burn-down needs emitter work and is not in this PR).

The generator, the both-direction gate, E081 at check time and the CI/ledger registration already existed. `proofs/target-availability.toml` is now schema 3: every one of the 122 `[[unavailable]]` rows carries `class` ∈ {pending-self-host, pending-port, native-only-forever}, forever rows carry a justification, and the rulings are reflected (`testing.assert_throws` and the nine-fn `process.exec*/spawn/…` family are native-only-forever with reasons that no longer say "wait for the wasi-p3 port"; `process.env` / `process.stdin_lines` stay pending-port; `env.set` has no row because the overlay shim serves it). The gate refuses an unknown or missing class, a class/reason-tag disagreement, a forever row without a justification, a pending row without an issue, and ratchets both pending counts against their ceilings (pending-self-host 56, pending-port 56, forever 10 — the pending-self-host ceiling went 34 → 56 by the class basis, every row that closes by work in this repo, not by regression; the header says so). `AVAIL_TSV_DIR` reuses one probe sweep so the declaration side can be forged without re-probing ~1000 fns per leg.

Measured with the tree's binary: structural ok 874 / wall 112, stock-p1 ok 922 / wall 59; declared and measured agree in both directions on both swept legs. Four forged matrices (class removed; justification removed; a pending-port row silently reclassified forever; a pending-self-host reason tag on another class) each red with the intended message, restored green. `docs/diagnostics/E081.md` describes the classes; build.rs and the fuzzer read keys by name, so the new keys are transparent.

Pre-push hook skipped (LEFTHOOK=0): a ledger, a script, docs and a comment; no compiler change.


O6lvl4:
Shard 1 was red because the process matrix test matched the availability row's old reason text ("no wasm host binding"), which this PR rewrites. It now matches E081's header line ("is not available on --target wasm"), the part the row does not own.

O6lvl4:

PR #2429, opened 2026-09-21
Make the nightly fuzz report say which shards its count came from, cap the listing by findings, and score nights at the 75% delivered-minutes bar

### LINKED ISSUE #2390: The nightly fuzz report under-counts silently: a preempted shard's findings leave the total, and the issue body is capped at exactly 20
> **状態** — 計測器・ゲート・着手可
> **一行で** — nightly fuzz の報告が 2 箇所で静かに過少申告する — preempt された shard の finding が夜の総数から落ち（`exit 143` は runner の通常停止なので稀な経路ではない）、issue body の一覧は両方の報告経路で `head -60` = ちょうど 20 件に切られる。どちらも verdict 行が既に持っている `shards=k/N` / `minutes_delivered` を issue に運べば解決する。
>
> **やること**
> 1. `reporting != FUZZ_SHARDS` の夜を fail にするか、少なくとも issue body に明示する。本文の文言案: 「8 shard のうち 6 から集めた 23 件 — 2 shard は報告せず、その finding は（あれば）この数に含まれない」。部分証拠で fail する前例は既にある（`:352` の "no shard uploaded output" guard）。
> 2. `.github/workflows/fuzz-nightly.yml:386` と `:437` の `head -60` を finding 単位の cap にし、cap が効いたときに body がそう言う（「23 件のうち 20 件を表示、残りは artifact と run log に」）。同じ cap が perf-class 経路（#2302）にもかかっている。
> 3. body の `across ${{ env.FUZZ_SHARDS }} shard(s)`（`:389`）を、実際に報告した shard 数に直す — 今は無条件に 8 と印字する。
> 4. 任意（本文の第 3 案）: 夜の count を upload された directory ではなく shard log から組む。log は artifact の 30 日保持より長生きし、seed は `run_id * 16 + shard` の導出値なので replay できる。
>
> **完了条件** — 部分的な夜の report が「N 件 / 8 shard 中 k shard から」と印字し、20 件を超えた夜は body が切られたことを自分で言い、`reporting != FUZZ_SHARDS` の夜が赤か明示つきになる。
> **根拠** — `.github/workflows/fuzz-nightly.yml`（562 行）に `head -60` が `:386` と `:437` の 2 箇所、body は `:389` で `FUZZ_SHARDS` をそのまま印字（2026-09-21 の grep）。本文の実測: run 35524384103 の verdict は `shards=1/8 ... findings=1 correctness=0` だが Slow は 3 件記録されていた（1 件は実在の perf ギャップ #2393）; run 33848741367 は upload した 6 shard の 8+4+4+3+2+2 = 23 で、shard 8 の 24 件目が欠落。
> **触る場所** — `.github/workflows/fuzz-nightly.yml`（`:352` の guard、`:384-389` の correctness 経路、`:435-437` の perf-class 経路）

---
The nightly fuzz report under-counts in two independent places, and both are silent. Found while triaging the backlog on #924, where they together turned **24 findings into a headline of 23 and a list of 20**.

Neither loses the finding itself — everything is recoverable from the run log (see the last section) — but both make the report claim a number it has not justified, and a fuzz report's whole job is to be the number someone acts on.

**Defect 1 is not a rare path.** `exit 143` is the *normal* shutdown signal on a GitHub runner, so a preempted shard is routine rather than exotic. That means the night's total has been quietly under-reporting on an unknown fraction of nights, and the one number that would reveal it — `shards=k/N` — never leaves the step log. This is the part to fix first.

**Measured, on the release-gate run itself.** Run `35524384103` against `v0.63.0-rc3` finished with **six shards failed, one cancelled, one succeeded**. Its own verdict line:

```
fuzz-night: shards=1/8 reporting=2 minutes_planned=480 minutes_delivered=119.9
            generated=20143 throughput=167.9prog/min findings=1 correctness=0 slow=1
```

Three separate facts sit in that one line:

- **`findings=1` is short by two.** Three `Slow` findings were recorded — shards 5, 6 and 7 — and two of those shards were preempted before uploading. The published count is a third of what the campaign actually found. One of the two lost findings turned out to be a **real native/wasm perf gap** (#2393), recovered only from the job log.
- **`minutes_delivered=119.9` against `minutes_planned=480` is 25% of the campaign.** A preempted shard does not only under-*count*, it under-*works*, and this is the measured form of that: `findings=N` is published unqualified while the night ran a quarter of its budget. A quiet night and a quarter-executed night are indistinguishable in the number that travels.
- **The instrument already knows.** It computes `shards=1/8` *and* `minutes_delivered` in the same line as the count it publishes. The qualifier exists, fully formed, one field away from the number that reaches the issue title. Nothing needs to be measured to fix this — only carried.

**One qualification, so this issue does not overclaim either.** `correctness=0` is trustworthy in the direction that matters: a preempted shard can only *hide* a correctness finding, never manufacture one. So "zero correctness findings across 25% delivered" is a **weaker assurance** than zero across a full campaign, not a false one. The defect is that the published number does not say which of those two it is.

Note also that the run-level log for this run can no longer be fetched at all — `gh run view --log` fails with `log not found: 106141459449` for the shard-1 job, killed hard enough to leave no log behind. Per-job logs for the other shards remain readable via `gh api repos/<o>/<r>/actions/jobs/<id>/logs`, which is how the three findings above were recovered.

The workflow is explicit that the kill is undetectable from inside the job, and treats the absence as the signal:

> A shard the runner RECLAIMED uploads nothing at all — the job is gone before this step — and that absence is precisely how the verdict job counts `shards=k/N`. Nothing here needs to detect the kill; not reporting IS the report.

That design is sound. The defect is only that `shards=k/N` stops at the step log while `findings=N` goes to the issue title — the report was built and then not delivered.

## Defect 1 — a preempted shard's findings vanish from the total

In run `33848741367`, shards 3 and 8 died with:

```
##[error]Process completed with exit code 143.
##[error]The runner has received a shutdown signal. This can happen when the runner
         service is stopped, or a manually started runner is canceled.
```

Preemption, not a fuzzer fault. **Shard 8 had already recorded a finding before it was killed** — `** FINDING [OutputDivergence] seed=541579861880 index=463` is in its log — and a preempted shard uploads nothing.

The night's total is then computed from what was collected, not from what was reported:

```bash
for d in shards/*/tools/xtarget-fuzz/findings/*/; do
  cp -R "$d" "night-findings/$(basename "$d")" 2>/dev/null || true
done
COUNT=$(find night-findings -mindepth 1 -maxdepth 1 -type d | wc -l | tr -d ' ')
```

The arithmetic is exact. The six shards that uploaded contributed 8+4+4+3+2+2 = **23**; shard 8's finding is the 24th and is absent from `COUNT`. So `findings=23` never meant "the night found 23" — it meant "23 findings survived collection".

The loss *is* printed, on the same line as the count:

```
fuzz-night: shards=6/8 reporting=6 minutes_planned=40 minutes_delivered=30.7
            generated=9980 throughput=324.9prog/min findings=23 correctness=23 slow=0
```

but only `correctness` reaches the issue (`TITLE="Nightly fuzz: $COUNT cross-target finding(s)"`). `shards=6/8` stays in the step log. A number qualified by "of the shards that came back" is published unqualified.

Note also the `2>/dev/null || true` on the `cp`: if two shards ever produce the same finding basename, the second is merged into the first and `COUNT` drops again, with no error. That has not been observed, but it is the same class and the same silence.

**Suggested fix.** Fail, or at minimum say so loudly in the issue body, when `reporting != FUZZ_SHARDS`. The phrasing that would have prevented this triage costing a day: *"23 findings collected from 6 of 8 shards — 2 shards did not report and their findings, if any, are not in this count."* The existing "no shard uploaded output" guard (line 352) already establishes that partial evidence is worth failing over; this is the same rule one step in.

All line numbers here are against `origin/develop` (562 lines).

## Defect 2 — the issue body's list is capped at exactly 20, in both report paths

`.github/workflows/fuzz-nightly.yml:384-386`, the correctness path:

```bash
SUMMARY=$(find night-findings -mindepth 1 -maxdepth 1 -type d ! -name 'Slow__*' \
  -exec cat {}/meta.txt \; \
  | grep -E '^(kind|summary|reproduce)' | head -60 || true)
```

and the identical cap again at `:435-437`, the perf-class path, differing only in `-name 'Slow__*'` instead of `! -name`:

```bash
SUMMARY=$(find night-findings -mindepth 1 -maxdepth 1 -type d -name 'Slow__*' \
  -exec cat {}/meta.txt \; \
  | grep -E '^(kind|summary|reproduce)' | head -60 || true)
```

Worth recording how the second one was nearly missed, since it is the same defect class as the issue itself: a repeat-folding reader rendered both hits on the first hit's line number with a `×2` marker, so the second location was invisible until the search was re-run with plain `grep -n`. **The cap that hides duplicate findings was found with a reader that hides duplicate matches.** When the question is *how many* or *where else*, a summarising reader is the wrong instrument.

So this is a pattern in the file rather than a slip in one step, and the perf-class tracking issue (#2302) is subject to the same cap — it happens to carry 1 finding today, so it is not currently truncated, but it would truncate identically at 21.

`meta.txt` contributes three matching lines per finding (`kind`, `summary`, `reproduce`), so `head -60` is a hard cap of **20 findings** in the body — while the sentence directly above it says "recorded **$COUNT** unique finding(s)". On this run that printed 23 and listed 20.

Nothing compares the two, so the body can disagree with its own headline indefinitely. The three that fell off were not marked as omitted; they were simply not there, which is why they read as unaccounted for rather than as truncated.

**Suggested fix.** Cap by findings rather than by lines, and say when the cap bites — `… showing 20 of 23; the rest are in the artifact and in the run log`. Any cap is fine as long as the body says it is a cap.

## The consequence, concretely

During the #924 triage I treated the 3 unlisted findings as unclassifiable and named artifact retention (`retention-days: 30`) as the cause. That was wrong, and the two defects above are why it looked right: the headline said 23, the body enumerated 20, and the difference had no explanation visible from the issue.

## Findings are recoverable long after the artifacts expire

Worth recording because it makes both defects much less costly, and because it is not obvious. **Run logs outlive the 30-day artifact retention and contain the complete record.** Each shard prints every finding:

```
** FINDING [OutputDivergence] seed=541579861878 index=547 — both targets disagree with reference interpreter (...)
```

and the seed is derived, not random — the workflow says so itself:

```bash
# replayable from the run alone: `run_id * 16 + shard` is unique per
# (run, shard) and well inside u64.
if [ -z "$BASE" ]; then BASE=$(( <run_id> * 16 )); fi
SEED=$(( BASE + <shard> ))
```

Recovering the full enumeration for `33848741367` is one grep over `gh run view <id> --log`, and all 24 replay under `xtarget-fuzz replay --seed S --index I` (the nightly passes no `--family`, so the default `all` reproduces). All 24 are CLEAN on `5d8568272`.

Since the log is the more durable record and already carries everything, a third option for defect 1 is to build the night's count from the shard logs rather than from uploaded directories — then a preempted shard costs nothing at all.

O6lvl4:
Closes #2390; implements the #924 ruling of 2026-09-21 (a streak night delivers ≥75% of its planned fuzz-minutes).

The verdict heading now reads `findings=N of k/N shards` and names the shards that did not report ("their findings, if any, are NOT in this count"), with delivered vs planned fuzz-minutes as a percentage; the `fuzz-night:` record line gains `reporting=`, `missing=`, `minutes_planned=`, `minutes_delivered=`, `delivered_pct=`, `budget=full|partial`, and a `::warning::` when a shard is missing. The tracking-issue body is rendered by `scripts/fuzz-night-issue-body.sh`: it caps by FINDINGS (20) and says "showing 20 of N; the remaining M are in the artifact and the run log"; the unconditional "across N shards" and both `head -60` are gone; if the verdict outputs are absent it says "an unknown number of the N planned shards", never "N of N". `fuzz-track-record.sh` and `fuzz-green-streak.sh` score a night from that line at the 75% bar, and a missing record is `NO RECORD` (unknown, not full) and breaks both streaks; pre-#2390 lines still score from their minutes fields (run 33848741367 → 76% full; the rc3 run → 24% partial).

Evidence: `tests/fuzz_night_report_test.rs` (7 tests) forges a preempted shard (missing=3,8, 60% partial), a full night, 7-of-8, zero shards, a 23-finding body (exactly 20 reproduce lines, "showing 20 of 23"), the unknown-reporting body, and streak scoring on forged logs including cancelled and no-record; plus a guard that the workflow calls the renderers and carries no `head -60`. `check-workflows.sh` green.

Left: building the count from shard logs so a preempted shard costs nothing (the issue's optional item); the Almide twin `fuzz-track-record` (unwired) not updated — note for #2130.

Pre-push hook skipped (LEFTHOOK=0): scripts, a workflow and a Rust test that runs the scripts.


O6lvl4:

PR #2430, opened 2026-09-21
Give map.fold and set.fold the accumulator ownership the staged list.fold has

### LINKED ISSUE #2408: set.fold and map.fold rebind the accumulator with no ownership guard — a leak, not a freed read, because they return the accumulator as a view
> **状態** — 実装・着手可
> **一行で** — `set.fold`（`collections_set.rs:577`-`578`）と `map.fold`（`collections.rs:464`-`465`）は callback body を lower して accumulator を rebind するのに ownership guard も前の accumulator の release も持たず、`Lowered::view` を返すため #2397 と違って freed read ではなく**解放されないメモリ**になる。コメント 2 本が「suggested shape」を自己訂正しており、修正は guard・前 acc の dec・`Lowered::owned` の**三点セット**（半分だけ適用すると現状より悪い）。
>
> **やること**
> 1. まず計測する: fresh-tail の `set.fold` / `map.fold` を alloc ledger で A/B する。本文と 1 本目のコメントの読みは `arm.rs:46`-`:52` の契約からの導出で、まだ実測ではない。
> 2. 三点セットを**まとめて**適用する — `rc_share_guard`、前の accumulator の dec、そして `Lowered::owned(b)`。staged `list.fold`（`list.rs:647`-`651`、`:657`）の規約を丸ごと採る。片方だけは駄目で、guard だけなら確保済みブロックに `View` を宣言して行が伸び、return flip だけなら borrow に `Owned` を宣言して #2397 と同じ cross-target divergence になる。
> 3. 反証可能な予測として測る: fresh-tail の行（`edge_map_order.almd:18`、`map_fold_heap_acc_skv.almd:12` / `:14` / `:28`、`map_fold_heap_acc.almd:16`）は **下がる**こと、scalar accumulator の行（`empty_collection_fold.almd:18` / `:19` / `:28`、`map_set_index_threshold` の Int fold）と borrowed-tail の fold は**バイト一致**であること（後者が negative control）。fresh 行が動かなければ読みが誤りなので止める。
> 4. `Lowered::view` の return を load-bearing として残す判断をするなら、lowering のその場所に「return mode が理由で guard は不要」と書く — 次の掃引が omission ではなく decision を見つけられるように。
> 5. これは #2397 の 3 例目**ではない**（観測が逆: #2397 は freed read、これは retained memory）。shape で再 triage されないよう C-334 の記録を保つ。
>
> **完了条件** — 三点セット適用後、`crates/almide-wasm/tests/golden/alloc-baseline.txt` の fresh-tail fold 行が下がり、scalar accumulator 行と borrowed-tail fold 行はバイト一致のまま、両レッグの出力が変わらないこと。あるいは fresh 行が動かず読みが反証されたことが記録されること。
> **根拠** — 現ツリー(develop)で行番号を確認: `collections_set.rs:577`-`578` と `collections.rs:464`-`465` は guard も dec も無く、`:590` / `:477` で `Lowered::view(b)` を返す。`list.rs:647`-`651` は `rc_share_guard` + 前 acc の dec を持ち `:657` で `Lowered::owned(b)` を返す。コメントの読み: init は 3 者とも `ArgMode::Retain` で lower される（`list.rs:637`、`collections.rs:437`）ので、retain した credit が解放されない。fresh-tail の leak 自体は **未測定** — `arm.rs` の契約からの導出で、alloc ledger は equality pin なので leak を含んだ値を pin している（本文の「借用 tail で probe して clean」は leak しない側の列を引いていた）。
> **触る場所** — `crates/almide-wasm/src/collections_set.rs`、`crates/almide-wasm/src/collections.rs`、`crates/almide-wasm/tests/golden/alloc-baseline.txt`、`docs/contracts/contracts.toml`（C-334）

---
Found by the re-sweep that #2397 forced, and recorded in C-334 rather than fixed
with it. **Not a blocker** — the distinction in the title is the whole point and
is the first thing to lose if this gets triaged by shape.

Line numbers are at develop `824ba091f`.

## The shape

```rust
// crates/almide-wasm/src/collections_set.rs:577   (set.fold)
self.lower(body, Some(b))?;
self.f.instructions().local_set(acc_p);

// crates/almide-wasm/src/collections.rs:464       (map.fold)
self.lower(body, Some(b))?;
self.f.instructions().local_set(acc_p);
```

Both lower a callback body and rebind the accumulator with **no ownership
guard and no release of the previous accumulator** — the same two lines
`10fed0494` gave the staged `list.fold` and `list.scan`
(`list.rs:647`, `list.rs:556`), and the same two lines #2397 restores to the
fused `list.fold`.

## Why it is a leak and not a freed read

`set.fold` and `map.fold` return their accumulator as a **`Lowered::view`**.
A view is not owned by the caller, so the caller never releases it — the
missing `+1` therefore has no matching `-1` to expose it. The block is never
freed under a live binding; it is simply never freed.

Probed with borrowed tails on both legs: clean. That is the reason this is
`P-low` and was recorded rather than fixed alongside #2397 — and the reason it
must not be re-triaged later as "the third instance of the #2397 bug", because
the observable is the opposite one. #2397 is a freed read (wrong bytes, silent,
exit 0). This is retained memory.

## Why fix it anyway

1. **The asymmetry is the defect.** Four sibling folds, two conventions. The
   next person to change how accumulators are lowered has to rediscover that two
   of them are safe for a reason that lives in their *return mode* rather than
   in their lowering.
2. **A view is a local decision that can change.** The day either fold returns
   an owned value, this becomes #2397 verbatim, and nothing in the emitter
   records that the return mode is load-bearing.
3. The alloc ledger is an equality pin, so the leak is **measurable** — a fix
   should move those rows, which is the cheap way to prove it was real.

## Suggested shape

Add the guard + previous-accumulator dec at both sites, matching `list.rs:647`
verbatim, and let the alloc ledger rows move. If the `Lowered::view` return is
to stay load-bearing instead, say so **at the lowering** — a comment naming the
return mode as the reason no guard is needed — so the next sweep finds a
decision rather than an omission.

## Provenance

The re-sweep that found these used the predicate **"lower a callback body, then
store its result"** — 44 `self.lower(<body>, …)` sites. The earlier #2398 sweep
used "stores into a **heap slot**" and counted guard *spellings*, which is why
it reported the family closed: an accumulator rebind into a local is not a heap
slot, and a file carrying no guard vocabulary at all (`list_fuse.rs`) registers
as nothing-to-count rather than as a gap. See C-334's `TWO CORRECTIONS` clause
and #2397.

O6lvl4:
Closes #2408.

`map.fold` and `set.fold` lowered the callback body straight into the accumulator local with **no ownership guard, no release of the previous accumulator, and a `view` return** — while the staged `list.fold` (`list.rs:637-657`) does all three. `init` arrives `ArgMode::Retain` in all of them, so the accumulator owns a credit from the first instruction; without the dec, that credit and every intermediate one were abandoned.

**Not a divergence — a leak. This does not gate the release.** Observable behaviour is unchanged on both legs, so there is no cross-target promise to add and no new contract: the ledger movement is the evidence.

## The guard and the return mode are ONE change

`arm.rs:25-28` names both failure modes, and doing either half alone is worse than doing neither:

- **guard without the return flip** — the accumulator now owns a credit and the arm still declares `View`: *"an arm declaring `View` for a block it allocated shows as a growing row"*.
- **return flip without the guard** — *"an arm declaring `Owned` for a view shows as a cross-target divergence"*: the caller releases a block it does not own, which is #2397 again and strictly worse than the leak being fixed.

So both sites take the staged fold's three lines verbatim: `rc_share_guard(body, b)`, dec of the previous accumulator, `Lowered::owned(b)`.

## What the measurement actually showed, including where it failed

Prediction recorded before looking: four fresh-tail rows drop, scalar rows byte-identical. **Two dropped.**

```
alloc   map_fold_heap_acc       66528 -> 66496   (-32)
        map_fold_heap_acc_skv   66560 -> 66416   (-144)
        2 of 724 rows changed; every scalar-accumulator control unmoved
```

`edge_map_order` and `map_fold_heap_accumulator` did not move. Rather than assume "the leak fits under the watermark", the **size** baseline gives a discriminator — instructions were added, so a fixture whose emitted code changed must have taken the arm:

```
size    edge_map_order             5108 -> 5104
        map_fold                   8805 -> 8801
        map_fold_heap_acc_skv      4099 -> 4111
        map_fold_heap_accumulator  3826 -> 3822
```

Both silent fixtures **did** take the arm. Their leak is real and simply sits below `heap_end`, which is a high-water mark rather than a total — the alloc ledger cannot see a leak that fits beneath the peak. Read from the alloc axis alone, those two rows would have said "clean".

## Why three sizes SHRANK when only code was added

`calls_modules.rs:46`:

```rust
Some(l) if l.own == Own::View && em.borrowed_temps.len() > depth => {
    Some(em.promote_escaping_view(l))
}
```

A `View` escaping a scope that holds borrowed temps is promoted — **an inc**. The old arms declared `view` for a block they actually owned, so the caller was emitting a compensating retain to make that declaration survivable. Declaring `owned` removes the compensation (−4 where it fired); the guard and dec add where the accumulator is a handle (+12).

That reframes the defect: not "a missing dec" but **a wrong ownership declaration the caller was silently paying to survive** — which is why a borrowed-tail probe reported these sites clean, and why the old path leaked in two different ways depending on whether the result escaped.

## Verification

- `run_parity` green over the whole corpus, and every fold-touching fixture byte-identical native == wasm: `edge_map_order`, `map_fold`, `map_fold_heap_acc`, `map_fold_heap_acc_skv`, `map_fold_heap_accumulator`, `empty_collection_fold`, `map_set_index_threshold`, `nested_map_list_values`, `map_entries_reclaim`.
- No alloc row anywhere increased.
- Checked before the leak measurement, because a wrong dec is a use-after-free and would matter far more than the leak.

Commits are separated for `check-ratchet-separation.sh`: implementation, then the re-recorded baselines.


O6lvl4:
Folded commit for commit into the fourth A-ci batch, #2508, which is judged as one tree (the `-x` trailers there name these commits). The previous owner of this PR confirmed it is unclaimed. Closing here; the issue closes with the batch.

Issue #2384, opened 2026-09-20
A develop build reports the previous release's version, and the package version pin believes it


O6lvl4:
A build from `develop` reports the **previous release's** version number, with nothing to distinguish it. `make install` between a release tag and the next version bump produces a binary that says `almide 0.62.0` and is not 0.62.0.

## Measured

`~/.local/bin/almide`, built 2026-09-13:

```
$ ~/.local/bin/almide --version
almide 0.62.0

$ cat m.almd
fn third(s: String) -> String = string.byte_slice(s, 2, 3) ?? "?"

$ ~/.local/bin/almide check m.almd
No errors found

$ <binary built from tag v0.62.0> check m.almd
error[E002]: undefined function 'string.byte_slice'
```

`string.byte_slice` is the **one** signature `scripts/check-interface-diff.sh v0.62.0 HEAD` reports as added. The binary reporting `0.62.0` has it; released 0.62.0 does not.

The window, from the log:

| | |
|---|---|
| `v0.62.0` tagged | 2026-09-08 (`f32cedf53`) |
| `string.byte_slice` lands | 2026-09-10 (`ec3613914`) |
| Cargo.toml → `0.63.0` | 2026-09-18 (`4b833bb15`) |

**Ten days in which every `make install` from `develop` produced a binary wearing `0.62.0`.** The window is not a slip; it is structural. The bump is one commit and everything before it in the cycle is mislabelled by construction.

## What it has already cost

A downstream session ran every one of its A/B package sweeps for this release with that binary as the "0.62.0" column. The rows were not stale — they were *mislabelled*, which no re-run would have caught, because the binary answers `--version` consistently and correctly-looking every time. Found only by building the tag and comparing, on a probe chosen from the interface diff.

It is also why the engine-consuming packages in this repo's neighbourhood have been on develop builds "without anyone deciding to": they call `string.byte_slice`, so the last release cannot build them, and nothing said so.

## Why it is not only a measurement nuisance

`src/project.rs:242` compares the same string against `almide.toml`'s `almide_min`:

```rust
let installed = env!("CARGO_PKG_VERSION");
```

So a develop build that *has* a feature is refused by a package requiring the release that first shipped it, and accepted by a package pinning the release it is not. The pin is doing the opposite of its job in both directions.

## The class

[...]

PR #2431, opened 2026-09-21
Refuse to record parity goldens from a stale tree, and read the oracle header every parity test skipped

### LINKED ISSUE #2405: The parity manifest generators write happily from a stale tree, and the one field that would say so is the one the gate skips
> **状態** — 計測器・ゲート・着手可
> **一行で** — 3 つの parity manifest 生成器は worktree が upstream より後ろでも黙って書き、それを唯一示す `# oracle:` provenance 行はゲートが `git diff -I '^# oracle: '` で意図的に除外している唯一の行なので、どのブランチにも存在しないビルドが作った golden が恒久的に比較対象になりうる。
>
> **やること**
> 1. `scripts/gen-run-manifest.sh` / `gen-check-manifest.sh` / `gen-ast-manifest.sh` が、worktree が追跡 upstream より後ろのときは**書き込みを拒否**するようにする（意図的な場合の override フラグ付き）。
> 2. 失敗メッセージに、間にある commit のうち emitter パスに触ったものを名指しさせる — それが pedantic と urgent を分ける情報。
> 3. 範囲を確認する: 同じ論法はコンパイラ挙動に紐づく golden を書く生成器すべてに当てはまる。alloc / size baseline は corpus を run manifest から取るので、古い run manifest は台帳 6 枚に伝播する。
> 4. `# oracle:` 行はゲートから除外されたまま・tree ごとに differ するままにする（= conflict も残る）。隠せなくなった conflict は単なる merge。
>
> **完了条件** — 追跡 upstream より後ろの worktree で 3 つの生成器が非ゼロ終了して何も書かず、override フラグを付けたときだけ書くこと。`# oracle:` 行が `scripts/check-parity-goldens.sh` から除外されたままでも、古いツリー由来の golden が commit できないこと。
> **根拠** — 本文の実測: #2404 の準備中に 3 枚を再生成したところ全行がもっともらしく、唯一動いたのは除外されたヘッダ（`# oracle: almide 0.63.0 at f0d819b2e` → `at 5d8568272`）。その worktree は `develop` の 22 commits 後ろで、うち 2 つが `crates/almide-wasm/src/` に触っていた（1 つは RC 隣接）。現ツリーで確認: 3 つの生成器に `upstream` / `behind` / `rev-list` の検査は無く、除外は `scripts/check-parity-goldens.sh:47`、ヘッダの共有ヘルパは `scripts/lib/oracle-header.sh`。
> **触る場所** — `scripts/gen-run-manifest.sh`、`scripts/gen-check-manifest.sh`、`scripts/gen-ast-manifest.sh`、`scripts/lib/oracle-header.sh`

---
Split out of #2403, where it was filed as a comment. Same class as the
statement drift there — *a field that exists, is correct, and is read by
nothing* — but the remedy is different, so it is its own issue.

## The hole

The three parity manifests carry a provenance header naming the binary that
produced them:

```
# oracle: almide 0.63.0 (dev) at e8c95e285 — the CLI built from this tree; informational
```

The parity gate compares with `git diff --exit-code -I '^# oracle: '`, so that
line is **deliberately excluded**. Nothing else reads it. It says
"informational" and it means it.

It is excluded for a good reason: it legitimately differs per generating tree,
so comparing it directly would fail every honest regeneration. It is also, for
that same reason, the only line guaranteed to conflict between two branches that
both regenerate — it was the sole `als-pin`-adjacent conflict between #2399 and
#2404 while the rows themselves were disjoint.

So the field that would tell you the rows are untrustworthy is unreadable
precisely because it is the field that always differs.

## What it nearly cost

While preparing #2404 I regenerated all three manifests. Every row looked
plausible. The only thing that revealed a problem was the excluded header:

```
- # oracle: almide 0.63.0 at f0d819b2e
+ # oracle: almide 0.63.0 at 5d8568272
```

The worktree was 22 commits behind `develop`, and two of those commits touched
`crates/almide-wasm/src/` — one of them RC-adjacent. Every hash I had just
generated was fixture output from a compiler **missing two emitter changes**.

Committed, that manifest would have been a golden produced by a build that
exists on no branch, and the gate reading it would have compared against that
forever, silently. I noticed only because a reviewer told me to check the diff
for churn. Nothing in CI would have.

## Proposed fix: refuse at write time, not at review time

**Have the manifest generator refuse to write when the worktree is behind its
upstream**, with an override flag for the deliberate case.

Why write-time rather than a gate:

- It fails at the point of error, in front of the one person who can cheaply
  fix it (`git pull`, rebuild, regenerate). A reviewer looking at a 300-row
  manifest diff cannot act on a stale-tree warning and will not spot one.
- It leaves the conflict **intact but harmless**. The header still differs per
  tree, still conflicts, still needs resolving — but a conflict that can no
  longer hide a stale build is just a merge.
- A warning printed at review time would be read by someone who cannot act on
  it, which is the same reason the nightly's red was not a delivery mechanism
  (#2390, #2391).

The check is a compare against the tracked upstream; the failure message should
name the commits in between that touched emitter paths, since that is what makes
it urgent rather than pedantic.

## Scope

`gen-run-manifest` / `gen-check-manifest` / `gen-ast-manifest` — and the same
argument applies to any generator whose output is a golden keyed to compiler
behaviour. The alloc and size baselines take their corpus from the run manifest,
so a stale run manifest propagates to all six ledgers.

O6lvl4:
Closes #2405.

The three `gen-*-manifest.sh` generators wrote happily from a stale tree, and the `# oracle:` header line that would say so was the field the gates skipped. `scripts/lib/oracle-header.sh` now refuses BEFORE truncating the manifest when the ORACLE binary is not this tree's CLI (layered, using #2384's stamp: `--version` must be `<Cargo.toml version> (dev…)`; a stamped sha must equal HEAD; an unstamped binary is accepted only from this tree's `target/` and only if no compiler source is newer), when `spec/` holds an untracked `.almd` (they were silently excluded), or when the worktree is behind its upstream by commits that touch compiler paths. Modes via the registered `ALMIDE_MANIFEST_TREE_CHECK`: `strict` (default), `gate` (CI's detached, unstamped artifact: untracked check only), `off` (prints a warning; the `# oracle:` line stays as the trace). `check-parity-goldens.sh` verifies each committed header before regenerating. `crates/almide-corpus` gained `oracle_header()` / `verify_oracle_header()`, and `ast_parity`, `check_parity` and `run_parity` now read the header: version must equal the root `[package]` version and kind must be `dev`; the sha is parsed but informational because a rebase moves it.

Measured on this tree: the foreign wt-docs-emit oracle (unstamped, other tree) → refused with the exact reason; the PATH 0.62.0 release binary → refused as "a released or other-tree binary"; under `off` the ast manifest regenerates with only the header sha moved. `crates/almide-corpus/tests/manifest_generators_refuse_stale_tree.rs` forges each stale tree (12 tests: release binary, wrong version, wrong sha, no sha, older than sources, untracked fixture also in gate mode, behind by a compiler commit vs a prose commit, header shapes accepted/refused, `[package]` vs `[workspace.package]` version). The committed manifests did not need regenerating.

Left: the spine parity tests were not built locally (disk); their edit is the identical three-line call as `ast_parity`, which was run green.

Pre-push hook skipped (LEFTHOOK=0): scripts, corpus-crate helpers and tests; run_parity's own change is the header read, exercised by the ast twin.


O6lvl4:

Issue #2396, opened 2026-09-20
json path index: an i32 atoi accumulator truncates it on wasm, so an out-of-range index removes or overwrites a real element


O6lvl4:
A JSON path index is truncated to 32 bits on the wasm leg. An index far outside any array's range — which native correctly treats as out of range and ignores — **silently removes a real element on wasm**.

**Correction (mechanism):** an earlier version of this issue attributed this to `i32.wrap_i64`, as in #2385. **That is wrong.** `crates/almide-wasm/src/json_path_helpers.rs` contains **zero** occurrences of `i32_wrap_i64`. See "Mechanism" below for what actually happens — the observable is identical, the code shape is not, and the difference decides where a gate for this family has to look.

## Reproduce — source only, six lines

```almide
import json

fn main() -> Unit = {
  let arr = json.parse("[10, 20, 30]") ?? value.null()
  let r = json.remove_path(arr, json.index(json.root(), -9223372036854775807))
  println("r:" + json.stringify(r))
}
```

```
native:  r:[10,20,30]      the index is out of range, nothing is removed
wasm:    r:[10,30]         element 1 is gone
```

## The index is taken mod 2^32

Every divergent row is exactly "remove the element at `wrap32`":

| index | `i32.wrap_i64` | native | wasm | |
|---|---:|---|---|---|
| `1` | 1 | `[10,30]` | `[10,30]` | agree — in range on both |
| `-1` | -1 | `[10,20]` | `[10,20]` | agree — in range on both |
| `4294967296` (2³²) | 0 | `[10,20,30]` | `[20,30]` | **diverges** |
| `4294967297` (2³²+1) | 1 | `[10,20,30]` | `[10,30]` | **diverges** |
| `-9223372036854775807` | 1 | `[10,20,30]` | `[10,30]` | **diverges** |
| `9223372036854775807` | -1 | `[10,20,30]` | `[10,20]` | **diverges** |
| `2147483648` (2³¹) | -2147483648 | `[10,20,30]` | `[10,20,30]` | agree — out of range on both |

The two agreeing out-of-range rows are the tell: `2147483648` wraps to `-2147483648`, which is *still* out of range for a three-element array, so wasm accidentally does the right thing. It is right by luck, not by a check.

## Mechanism — an i32 atoi accumulator, not a narrowing instruction

[...]

PR #2432, opened 2026-09-21
Widen the domain-edge matrix to the json and value surfaces, name every skipped slot, and gate the skip population both ways

### LINKED ISSUE #2402: domain-edge matrix skipped 12 of 22 json signatures, including the one that carried #2396
> **状態** — 計測器・ゲート・着手可
> **一行で** — `tools/domain_edge_matrix.py` は `VALUES` / `RENDER` に型の項が無いセルを skip して数えるだけで、その skip 母集団を誰も読み返していない。json は 22 シグネチャのうち 12 が unbuildable で、その中に #2396 を運んだ `json.index(path: JsonPath, i: Int)` が入っていた。
>
> **やること**
> 1. `VALUES` / `RENDER` に `JsonPath` と `Value` の項を入れる（本文の 4 行そのまま。`JsonPath` は不透明なので、既知のドキュメントに**適用して**描画する — そのコメントも一緒に）。json family が skip ではなく計測されるようにする。
> 2. `proofs/domain-edges.toml:8` のヘッダを、ツールの母集団（`parse_stdlib(only=None)` は全モジュールを歩く）と、ある実行が実際に計測したモジュール（2026-08-17 の 6 つ）を区別する文言に直す。
> 3. 残る `value` の skip を閉じるか、閉じないなら**名前で列挙**して残差を帰属させる（数のままにしない）。
> 4. skip 数が **増えた**ときにゲートが落ちるようにする — divergence 数が増えたときと同じ扱い。unbuildable になったシグネチャは失われたカバレッジで、今は誰も気づかない。
> 5. コメントの残差を埋める: 全モジュール掃引は `bytes` / `list` / `string` の一部までで停止したので、`string` の末尾・`datetime`・`io`・`math` は widening 下で **未計測**（clean ではない）。掃き直すステップを同梱するか、ディスクに余裕のあるマシンで回す。
>
> **完了条件** — json に対する `tools/domain_edge_matrix.py` の実行が `skipped signatures (unbuildable): 0` を報告し、`proofs/domain-edges.toml` のヘッダが母集団と実行済みモジュールを区別し、`value` の残る skip が名前で記録され、skip 数の増加が `scripts/check-domain-edges.sh` を落とすこと。
> **根拠** — 本文の実測: 変更前の `json` は 14 cells / 10 signatures・skipped 12、`VALUES` / `RENDER` 追加後（#2396 未修正のツリー）は 179 cells・DIVERGE 6（`json.index` の u32_max / two_pow_32 / i64_max_m1 / i64_max / i64_min_p1 / i64_min）・skipped 0。コメントの部分掃引の実測: 67 divergent cells のうち台帳に無いのは 9 行で、うち `bytes.new` / `bytes.pad_left` / `bytes.pad_right` の `i32_max` 3 行は #2387 の可用性依存 phantom。現ツリーで確認: `tools/domain_edge_matrix.py` に `JsonPath` の項は無く、`proofs/domain-edges.toml:8` は今も `six modules (bytes, list, string,` と書いている。
> **触る場所** — `tools/domain_edge_matrix.py`、`proofs/domain-edges.toml`、`scripts/check-domain-edges.sh`

---
`tools/domain_edge_matrix.py` skips a cell whose parameter or return type has no
entry in its `VALUES` / `RENDER` tables, and counts the skip. That is the honest
design — "the matrix reports what it could not measure rather than pretending".
But the skipped population was never read back, and it contained a release
blocker.

## Measured

On `json`, before any change:

```
domain-edge matrix: 14 cells over 10 signatures
  AGREE    14
  skipped signatures (unbuildable): 12
```

**12 of 22 json signatures were unbuildable**, and `json.index(path: JsonPath, i: Int)`
was one of them — the signature that carried #2396, where an out-of-range index
silently deleted or overwrote a real element on wasm.

Adding two `VALUES` entries and two `RENDER` entries closes it:

```python
VALUES["JsonPath"] = "json.root()"
VALUES["Value"]    = 'json.parse("[10, 20, 30]") ?? value.null()'
# A JsonPath is opaque, so it is rendered by APPLYING it to a known document —
# the only way an out-of-range step is observable at all.
RENDER["JsonPath"] = 'json.stringify(json.remove_path(json.parse("[10, 20, 30]") ?? value.null(), {e}))'
RENDER["Value"]    = 'json.stringify({e})'
```

With those, on a tree **without** #2396's fix:

```
domain-edge matrix: 179 cells over 10 signatures
  AGREE    173
  DIVERGE  6          json.index at u32_max, two_pow_32, i64_max_m1, i64_max, i64_min_p1, i64_min
  skipped signatures (unbuildable): 0
```

The widened matrix **independently rediscovers #2396**. The un-widened one
could not see it at all.

`value` also improves but is not closed: 33 cells over 14 signatures, all AGREE,
still 12 skipped.

## Two corrections to how this gate's coverage has been described

Both were made today, by two different readers, from the same two places:

1. **"a JSON path index is not a public stdlib `Int` parameter that the edge
   matrix enumerates"** — it is one. The cell is skipped because the
   *receiver* type `JsonPath` has no constructor in `VALUES`, not because the
   parameter is invisible.
2. **`proofs/domain-edges.toml`'s header says "six modules (bytes, list,
   string, datetime, io, math)"**, which reads as the tool's population.
   `parse_stdlib(only=None)` walks **every** module; six is what that
   particular 2026-08-17 run measured. The header should say so.

A skip count that nobody reads back is a coverage claim with a hole of exactly
the shape of the bug it misses — the same failure as a gate whose findings go
somewhere nobody reads (#2379, #2390).

## Done when

- `JsonPath` / `Value` in `VALUES` and `RENDER`, with the "render by applying
  it" note, so the json family is measured rather than skipped
- `proofs/domain-edges.toml`'s header distinguishes the tool's population from
  the modules a given run measured
- the remaining `value` skips are either closed or **named**, so the residue is
  attributed rather than a number
- ideally: the gate fails on a skip count that GROWS, the way it already fails
  on a divergence count that grows — a signature that becomes unbuildable is
  coverage lost, and right now nothing notices

O6lvl4:
Closes #2402.

The matrix "skipped 12 of 22 json signatures" for two reasons that were not a filter: `JsonPath`/`Value` had no value/render entry, so `json.index(i)` — the only public json fn with an Int parameter, the one that carried #2396 — was unbuildable; and the other twelve were `json_path_*` self-host helpers a program cannot call (E002), which the old tool counted as AGREE because both legs failed with empty stdout. Two deeper holes on the way: the fn regex could not see generic fns (25 invisible), and `parse_type` stripped `?`, so every `T?`-returning slot failed to compile and counted as AGREE.

Now: `JsonPath` (rendered by applying it to `[10,20,30]`), `Value`, generics instantiated at Int, `T?` and `Result` printed through `match`, `effect fn` and `mut` params; the public surface read from `almide compile <mod> --json` (helpers named, not slots); one `almide check` per slot so a non-compiling probe is a named skip, never AGREE; coverage printed as `coverage: N of M signatures exercised, skipped: [slot: reason, …]`. `proofs/domain-edges.toml` records 186 named skips (`skip_count` + `[[skip]]` rows) and `check-domain-edges.sh` diffs the population both ways (an undeclared skip = coverage lost → fail; a declared skip that became buildable → fail, delete the row). The weekly workflow measures json and value too.

A/B that matters: on the published 0.62.0 binary the widened matrix reports `DIVERGE 6` on `json.index(i)` at the u32/i64 edges — it rediscovers #2396 on its own; the old tool could not see the slot. On the develop build: json `1 of 1 exercised`, value `1 of 1`; whole stdlib `334 of 520 signatures exercised, 186 skipped` (matrix 70, bytes 36, int 25, net 17, map 14, …: unrenderable returns, unbuildable params, 6 type-check).

Negative test `tests/domain_edge_skip_ledger_test.rs`: a forged public `json.index(path: Widget, …)` → the tool names it as unbuildable and the gate exits 1; a phantom `[[skip]]` row → exit 1 "now buildable — delete". Not run under cargo locally (disk); exercised by hand with a script replica — please run it with ALMIDE_BIN before merge.

Note: the widening adds cells to bytes/list/string/datetime/io/math that no run has measured; the first weekly run may surface new DIVERGE rows there, which are findings, not a reason to loosen.

Pre-push hook skipped (LEFTHOOK=0): a Python tool, a script, a ledger, a workflow and a Rust test that runs the tool.


O6lvl4:

PR #2433, opened 2026-09-21
Gate the fold-seed-consumes-its-source shape from source with a per-file census floor

### LINKED ISSUE #2380: Gate the fold-seed-consumes-its-own-source shape from source, without a compiler
> **状態** — 計測器・ゲート・着手可
> **一行で** — #2377 の形（`list.fold(xs, <xs を消費する seed>, f)`）を source だけで捕まえる scanner は下流セッションのスクラッチにしか無く、このリポジトリの corpus に対して走るゲートが存在しない（`scripts/` に該当スクリプトは無い）。
>
> **やること**
> 1. scanner をこのリポジトリに置く。`list.fold` / `foldf` / `fold_float` / `reduce` / `scan` の引数リストを top level で分割し、**第一引数が bare identifier でそれが seed の中に再登場する**呼びを報告する。出発点は `o6lvl4-ab` 側の `scratchpad/e0505/scan.py`。
> 2. 空の結果を信じる前に既知 2 形で発火することを示す（本文の 2 例をそのまま自己テストに）。
> 3. CI から走らせる。所有はこのリポジトリ側 — corpus に対して fusion pass を触る変更のときに走るのが要点で、下流の consumer が持つゲートは違う母集団を違う時刻に見てしまう（コメントの裁定）。
> 4. ゲートが出来たら、下流 22 パッケージへ向けた走行を別に依頼する（`o6lvl4-ab` が per package で行を報告する用意あり）。
>
> **完了条件** — `scripts/` の scanner が CI から走り、既知 2 形で発火することを自身のテストが pin し、このリポジトリの corpus に対して 0 hit で緑。
> **根拠** — 本文の実測（4,172 ファイル / 0 hit、既知 2 形で発火を先に確認）。#2377 は CLOSED、修正と回帰テストは PR #2376（MERGED、`crates/almide-codegen/src/pass_chain_source_borrow.rs` と `tests/chain_source_borrow_vs_consuming_seed_test.rs` を含む）。ゲートは 2026-09-21 時点で不在（`scripts/` に該当スクリプト無し）。
> **触る場所** — `scripts/`（新しい scanner）、`.github/workflows/ci.yml`、`spec/` / `stdlib/`（走行対象の corpus）

---
#2377 was a native `E0505` behind the "this is an Almide bug" banner: `list.fold(xs, <seed that consumes xs>, f)` borrowed the source and moved it in the seed. It is fixed, and four regression tests pin it. This issue is about the cheaper instrument that would have caught it, and would catch the next one.

## The shape is detectable from source alone

`o6lvl4-ab` wrote a scanner while confirming the blast radius: it splits the argument list of `list.fold` / `foldf` / `fold_float` / `reduce` / `scan` at top level and reports any call whose **first argument is a bare identifier that appears again inside the seed**. Ten lines of argument splitting, no compiler, no build.

They proved it fires on both known shapes before believing an empty result — the discipline that makes an absence a measurement:

```almide
list.fold(xs, list.is_empty(list.sort_by(xs, ((y) => 1))), ((a, b) => a))   // fires
list.fold(v, v, ((a, b) => a))                                              // fires
```

Then ran it over every `.almd` under their workspace: **4,172 files, 0 hits.** So no downstream package writes this shape today — which is a fact worth having, and also why nobody reported it before a fuzzer synthesised it.

Starting point: `scratchpad/e0505/scan.py` on the `o6lvl4-ab` session's side.

## Why it is worth a gate rather than a note

The regression tests pin the shapes we know. A source-level scanner covers the ones we do not, costs nothing to run, and fails with a diagnostic instead of a rustc error under a banner that tells the author to file a compiler bug. It is the same argument as the AST gates already in `scripts/`: cheap, total over the corpus, and independent of whether the emitter happens to be right today.

## Deliberately not in 0.63.0

Landing a new gate while cutting a release is how a release acquires a defect it did not have. The fix and its tests are in #2376; this is the follow-up.

Scanner and the 4,172-file measurement by `o6lvl4-ab`.

O6lvl4:
Closes #2380.

"Without a compiler" read as: no compiler in the loop — the scanner reads `.almd` text, so the emitter under test cannot blind it. It lives where #2128 puts a gate that reads a corpus and renders a report: `tools/almide-gates/src/fold_seed_source.almd`, subcommand `fold-seed-source`, stdlib-only so the PATH binary runs it in lefthook (1.4 s). Shape: a `list.fold` / `foldf` / `fold_float` / `reduce` / `scan` call whose first argument is a bare identifier that reappears as a whole identifier in the seed (or the closure for `reduce`), plus the pipe spelling `xs |> list.fold(seed, f)` reported as `form=pipe`; string literals blanked and `//` tails dropped first (the scanner's own header comment and a comment in `list_fold_hsca.almd` were false sites otherwise); an argument list that never closes is an error, not a silent skip.

The gate is a shrink-only FLOOR, not a ban: the shape is legal after #2376, and the point is that fixtures exercising it stay in the corpus. `proofs/fold-seed-source-baseline.txt` holds `<path> <count>`; each file must keep at least that many sites; files with sites and no row are reported as unpinned; stdlib/ sites are printed separately (currently 0, the issue's "nobody writes this shape"). Fixture `spec/lang/fold_seed_source_test.almd` carries the four shapes of PR #2376's Rust tests (passes on wasm and rust), floor seeded at 4. Census: 1,749 files, 4 sites in 1 file.

Evidence: 11 scanner unit tests in the module (run by the existing `almide test tools/almide-gates/` step), the negative end-to-end (a forged corpus with one test cut → `has 3 site(s) … floor is 4`, exit 1; fixture deleted → exit 1), gate-verification row (reads=structured, NEGATIVE_TESTED), ci.yml step in the almide-gates job, lefthook entry.

If a 0-hit ban over stdlib/ is wanted in addition, it is a one-line change in `verdict`.

Pre-push hook skipped (LEFTHOOK=0): Almide gate sources, a fixture, a baseline, ci.yml and lefthook.


O6lvl4:

Issue #1469, opened 2026-08-16
Documentation is 8x thinner than the cohort and hand-written: ship a doc generator


O6lvl4:
> **状態** — 計測器・ゲート・着手可
> **一行で** — `docs/stdlib/*.md` の機械所有部分は署名索引ブロックだけで、~1,700 の公開名に対するモジュール別リファレンス本文は手書きのまま（本文の「generator: none」という前提はコメントで訂正済み）。
>
> **やること**
> 1. 生成範囲と手書き範囲を分ける: `docs/stdlib/*.md` の `<!-- BEGIN GENERATED SIGNATURE INDEX ... -->` ブロックが機械所有、その外が手書き散文。
> 2. その生成ブロックを署名索引からモジュール別リファレンスページへ広げる（各公開名のシグネチャ・doc コメント・`@since` epoch）。ホームは `tools/gen-stdlib-doc-index.py`（`src/cli/docs_gen.rs` は llms.txt + diagnostics-registry drift 用、とコメントの指摘）。
> 3. 残る手書き散文を、生成器が保存する prose overview セクションへ再配置する。
> 4. `proofs/gate-verification.toml` に negative control 付きの行を足し、`scripts/check-gate-verification.sh` に載せる。
>
> **完了条件** — `python3 tools/gen-stdlib-doc-index.py --check` が拡張後の生成ブロック（全公開名を含む）に対して緑で、`docs/stdlib/` の手書き散文が生成器の保存範囲に入り、`proofs/gate-verification.toml` にこのゲートの negative control 行がある。
> **根拠** — 本文の survey 測定（43 pages / 5,987 lines / 42 modules / generator none）と、それを訂正したコメントの実測（2026-08-17: 41/43 ファイルが生成ブロックを持ち、1,287/5,987 行 = 21% が生成、`--check` は既にある）。2026-09-21 に tree で再確認: `docs/stdlib/*.md` は 45 ファイル・43 が生成ブロックを持ち・合計 9,502 行、CI gate は `.github/workflows/ci.yml:1328`。
> **触る場所** — `tools/gen-stdlib-doc-index.py`、`docs/stdlib/`、`.github/workflows/ci.yml`、`proofs/gate-verification.toml`

---
Found by the compiler-championship survey.

## The measurement

| project | generated API docs | per module | generator |
|---|---:|---:|---|
[...]

PR #2437, opened 2026-09-21
Derive the exported types and the clock pages in the stdlib doc generator, and gate every page's generated block

O6lvl4:
Advances #1469 (re-measured and re-scoped on the issue; the compiler-side doc-comment channel is split out as https://github.com/almide/almide/issues/2436).

Measured at 281f23135: 45 stdlib pages, generated block on 43 (compute and duration were skipped by name because they are checker surface, not modules), 986 public fns (the issue's "~1,700 names" is not the case; README says 986 too), 442 fns with a hand-written heading, 13 modules with zero per-fn prose. The "8x thinner" claim recomputed with the issue's own method is 5.2x today, and 2.6x with a consistent method on both sides.

Changes: `tools/gen-stdlib-doc-index.py` emits exported types (record/variant/alias, 6 types in 5 modules) as a `## Type index` fence, and derives the compute/duration pages from `time_units.rs`'s TIME_UNITS × TIME_MODULES table (loud failure if the table shape changes), so every page is generator-owned — a page without a block is stale, not exempt. `scripts/check-stdlib-docs.sh` wraps the regenerate-and-diff, wired into ci.yml (replacing the bare `--check` call) and lefthook, with a gate-verification row (scalar, three negative controls: a forged signature, a deleted type index, a stripped block; regenerated tree passes 45/45). `check-interface-diff.sh 281f23135 HEAD` is additive (+12 clock constructors); note it ignores `type` lines, a small follow-up if type removals should count as breaking.

The real blocker for per-fn prose is measured and filed: the interface JSON has a `doc` slot, but the by-name compile route serves the embedded source with every comment blanked (0/986 docs), and the file route is not signature-equivalent; the stdlib sources contain 0 `///` lines anyway.

Pre-push hook skipped (LEFTHOOK=0): a Python generator, seven regenerated doc pages, a script, ci.yml, lefthook and a ledger row.


O6lvl4:

PR #2441, opened 2026-09-21
Pin the domain-edges wasm leg to a declared memory budget so the i32_max verdict is the compiler's, not the machine's free memory

### LINKED ISSUE #2387: The domain-edges gate decides its i32_max rows on free memory, not on the compiler: the same binary gives AGREE and DIVERGE minutes apart
> **状態** — 計測器・ゲート・着手可
> **一行で** — `proofs/domain-edges.toml` の `i32_max` 行は、コンパイラではなく「その瞬間どれだけメモリが空いていたか」で決まる — wasm レッグの上限が `StoreLimits::default()`（`crates/almide-wasm-run/src/host.rs:663`）で構造的な境界ではないため。本文の 4 段のうち **4 番（台帳ヘッダに警告を載せる）は着地済**で、残るのは上限を構造的に固定する 1〜3。
>
> **やること**
> 1. 上限を CLI に出す（または matrix tool 用の harness entry point を作る）ので、sweep が wasm レッグを固定予算の `run_wasm_capped` 経由で走れるようにする。`run_wasm_capped` は `host.rs:626` に既にあり、呼び出しは `crates/almide-wasm/tests/rc_budget.rs` だけ、CLI フラグは無い（2026-09-21 grep で確認）。
> 2. native レッグにも同じ上限をかける。片方だけだと同じ種類の phantom が反対のレッグに移るだけ。
> 3. 予算を台帳ヘッダに記録する。行の意味が「最後にこれを走らせたマシンで乖離した」ではなく「宣言された予算で乖離する」になる。
> 4. `bytes` モジュールの 3 行（`bytes.new:len:i32_max`、`bytes.pad_left`/`pad_right:target_len:i32_max`）は上限が固定されるまで宣言しない — 台帳ヘッダ（`proofs/domain-edges.toml:47-59`）が既にそう書いていて、ゲートはそれまで bytes で赤のまま。
>
> **完了条件** — `i32_max` の升が同じツリーに対してマシンの負荷に関わらず同じ verdict を返し（負荷をかけた再走行で反転しない）、`proofs/domain-edges.toml` のヘッダがその予算を数値で持ち、`domain-edges.yml` の 6 モジュールが宣言された予算の下で緑。
> **根拠** — 本文の実測（同一バイナリ `5d8568272` / 同一プログラムで AGREE と DIVERGE が数分差、分類は `(returncode, stdout)` 比較 = `tools/domain_edge_matrix.py:263-270`）とコメントの独立再現（2026-09-21、負荷下で withdrawn 3 行が再出現）。台帳側の現況（2026-09-21 確認）: `string.pad_end`/`pad_start:n:i32_max` の 2 行は削除済で `count = 62`、削除理由が「phantom が永久宣言されていたから、何も直っていない」と `:44-45` に明記、`:54` が #2387 を名指し。
> **触る場所** — `crates/almide-wasm-run/src/host.rs`（`:626` の `run_wasm_capped`、`:661-663` の limits）、`src/cli/`（上限フラグ）、`tools/domain_edge_matrix.py`、`proofs/domain-edges.toml`

---
The `i32_max` rows in `proofs/domain-edges.toml` are decided by how much memory the machine has free at that instant, not by the compiler. The gate is bidirectional and the ledger is shrink-only, so those cells fail the job in one direction or the other more or less at random — and a phantom that lands once is then declared forever.

This is an **instrument** defect. It is not a cross-target divergence, it does not gate the release, and it is not a sibling of #2385.

## The measurement

Same binary (`5d8568272`, `almide 0.63.0`), same program, same machine, minutes apart:

```almide
fn main() -> Unit = {
  let b = bytes.new(2147483647)
  println("len=${int.to_string(bytes.len(b))}")
}
```

```
first run                     later runs (machine under pressure)
  native rc=0  len=2147483647   native rc=0  len=2147483647
  wasm   rc=0  len=2147483647   wasm   rc=1  Error: out of memory
  → AGREE                       → DIVERGE
```

The classifier compares `(returncode, stdout)` (`tools/domain_edge_matrix.py:263-270`), so those two readings are a clean AGREE and a clean DIVERGE of the same cell. `bytes.pad_left` and `bytes.pad_right` at the same edge behave identically.

Note what the wasm leg does in the second reading: it takes the **defined C-197 line**. Nothing is wrong with the compiler in either reading. Whether the cell "diverges" is entirely a question of whether a 2 GiB `memory.grow` happened to succeed.

## Mechanism

`almide run --target wasm` reaches `run_wasm_src` with `max_memory_bytes: None`:

```rust
// crates/almide-wasm-run/src/host.rs:661
let limits = match max_memory_bytes {
    Some(cap) => wasmtime::StoreLimitsBuilder::new().memory_size(cap).build(),
    None      => wasmtime::StoreLimits::default(),   // :663 — no ceiling
};
```

So the wasm leg's ceiling is **host availability, not a structural bound**. `i32_max` is 2 GiB, which sits right where a host either can or cannot satisfy the growth.

The fix site is already in the file. `crates/almide-wasm-run/src/host.rs:626`:

```rust
/// Run under a hard linear-memory budget (bytes). Growth past the cap
/// fails, which the emitted allocator turns into the DEFINED
/// "Error: out of memory" + exit 1 (C-197) ...
pub fn run_wasm_capped(bytes: &[u8], max_memory_bytes: usize) -> anyhow::Result<RunResult>
```

A capped run produces exactly the line measured above — the wasm leg is already on the C-197 path; it is just being triggered by host allocation failure instead of a declared bound. `run_wasm_capped` has no caller outside `crates/almide-wasm/tests/rc_budget.rs`, and **there is no CLI flag for it**, so the matrix tool (which drives `almide run`) cannot pin the ceiling today.

## The tool already knew

`tools/domain_edge_matrix.py:305-326`, the comment above the solo retry:

> A cell sitting just inside the 2 GiB ceiling therefore asks BOTH legs for 2 GiB, and under memory pressure one of them can lose that race while the other wins. That is a machine artifact wearing a divergence's clothes, and it is indistinguishable from the real thing in the verdict alone.
>
> Measured: a full sweep run CONCURRENTLY with another heavy job reported `string.pad_start:n:i32_max` and `pad_end:n:i32_max` as divergent; re-run on a quiet machine both agree ... Two phantom rows out of 7767 is enough to make a shrink-only ledger untrustworthy, **because a phantom that lands once is then declared forever.**

The solo retry was built for precisely this and does not survive `domain-edges.yml`: the workflow runs six modules as six parallel jobs, so the pressure lasts the whole sweep instead of being a transient race. The retry re-runs once, immediately, under the same conditions.

## It has already happened — twice declared, three nearly added

The two cells that comment names by hand are **in the committed ledger**:

```
proofs/domain-edges.toml:257   cell = "string.pad_end:n:i32_max"
proofs/domain-edges.toml:269   cell = "string.pad_start:n:i32_max"
```

They carry the C-197 "contracted divergence" reason, under a header that says the sweep was taken on a *"quiet machine, solo-retry instrument"*. The tool documented the phantom; nobody carried it into the ledger's header; the rows went in anyway.

The first ever run of the gate (`35508754819`, 2026-09-20, 3 of 6 modules red) then reported those same two rows as **stale — no longer diverging** — and simultaneously reported three new undeclared ones on the other module:

```
bytes.new:len:i32_max
bytes.pad_left:target_len:i32_max
bytes.pad_right:target_len:i32_max
```

During the v0.63.0 soak a parallel session declared all three in the ledger with the C-197 reason, count 62 → 65, on the strength of **one** reading of `bytes.new(i32_max)` taken while that machine was running a cargo test suite, two matrix sweeps and a release build. It was withdrawn before landing, once the flip above was measured. That near-miss is the point of this issue: the failure mode is not exotic, it is what a careful reader does when handed a red `i32_max` row, because the row looks exactly like the contracted divergence it sits next to.

Note also that the two string deletions are correct but **not for the reason a reader would assume** — they go because a phantom was declared forever, not because anything was fixed. "Both legs answer there now" is itself an availability claim wearing structural clothes.

## What is not affected

- **#2385 is unaffected and remains the real finding.** Its four cells (`pad_start`/`pad_end` at `i64_max`, `i64_max_m1`) are deterministic on any machine: `i64::MAX` is a size nothing can satisfy. And its `two_pow_32` diagnosis — the width taken mod 2^32, answering `"xxxab"` — is a wrong answer, not an availability artifact.
- The larger edges (`two_pow_32`, `u32_max`) are far enough past any host's ceiling that both legs fail on any machine.

The hazard is specific to edges that land near a real host's allocation limit, `i32_max` today.

## Suggested fix

Make the wasm leg's ceiling structural so the verdict is reproducible:

1. Expose the cap on the CLI (or give the matrix tool a harness entry point) so the sweep can run the wasm leg through `run_wasm_capped` at a fixed budget.
2. Cap the **native** leg the same way, or the same class of phantom just moves to the other leg.
3. Record the budget in the ledger header, so a row's meaning is "diverges at a declared budget" rather than "diverged on whatever machine last ran this".
4. Until then, the ledger header should carry the tool's warning — that is where the reader who is about to declare a row will be looking, and it is the step that was missed.

## Why the gate matters more than a hand audit

Worth stating because it is the argument for repairing this rather than dropping it. `96c2bcffd` guards the full i64 length before `i32.wrap_i64` — but only at the **allocation** sites in `crates/almide-mir/src/render_wasm_p2.rs`. #2385's defect is the same discipline missing in `lower_string_pad`, in a different crate, which that commit never touched. There are **217** `i32_wrap_i64` / `i32.wrap_i64` sites in `crates/` (165 in the Rust-API form, 52 in the wat-text form), so auditing them by hand is not a credible instrument. This gate is. It found #2385 on its first run. It needs to be trustworthy.

O6lvl4:
**Stacked on #2432; merge after it.** Closes #2387 (items 1 and 3; item 2, the native cap, is left open and said so in the header).

## What was wrong

The `i32_max` rows of `proofs/domain-edges.toml` were decided by free memory, not by the compiler: `almide run --target wasm` executes on the embedded host with `StoreLimits::default()`, so a 2 GiB request (a 1-byte-element fn at `i32_max`) either succeeded or took the C-197 abort depending on what the machine had free. The same binary read `bytes.new(i32_max)` as AGREE and DIVERGE minutes apart, two phantoms were declared for a month, and three more were nearly added.

## What this does

- **The wasm leg is pinned to a declared budget, structurally.** `tools/domain_edge_matrix.py` now builds the module (`almide build --target wasm`; a wall surfaces with its mark exactly as before) and runs it on the wasmtime CLI with `-W max-memory-size=1073741824` (`WASM_BUDGET_BYTES`). Any request past the budget takes the defined `Error: out of memory` / exit 1 on every machine. No compiler change; the repo ratified (A 2026-08-17, `check-alloc-ceilings.sh`) that plain allocations carry no chosen cap, so there was no language ceiling to bound the probe with — the budget is the instrument's, and it is declared.
- **The ledger declares the budget** (`wasm_budget_bytes = 1073741824`) and a row now means "diverges at this budget". The five 1-byte-element cells (`bytes.new:len`, `bytes.pad_left/pad_right:target_len`, `string.pad_start/pad_end:n` at `i32_max`) are DIVERGE at it and are declared with a C-197-at-budget reason (count 62 → 67). The 17 already-declared `i32_max` rows (array readers, `bytes.repeat`, `list.range`) request ≥ 4 GiB on wasm32 and do not move.
- **An availability reading cannot reach the ledger.** The ledger step is extracted to `tools/domain_edge_ledger.py` (keeping #2432's skip census, `--measured`, `--skips-only`, `--update --only` scoping) and refuses, before any diff or write, a measurement with no `wasm_budget_bytes`, a different budget, a ledger without a declared budget, or a DIVERGE cell marked `pinned: false`. `scripts/check-domain-edges.sh --selftest` forges each of those (13 cases, including through `--update`, asserting the ledger file is byte-identical afterwards) plus pinned controls diffed both ways, cells and skips. Wired into the PR `checks` job and the weekly workflow, which now installs wasmtime (the `ci-cross-target.yml` block).
- The native leg stays uncapped: RLIMIT_AS is unenforced on Darwin (setrlimit → EINVAL, measured), so a native cap would split local and CI verdicts; its 2 GiB request is the same availability claim the ledger already makes at 4 GiB for every `u32_max` row. The leg that flipped in every recorded reading was the wasm leg.

## Evidence

`almide 0.63.0 (dev)`, wasmtime 47.0.2, `--only bytes/string --edge i32_max` (272 cells), three consecutive runs and a fourth under a 6 GiB touched-memory hog:

| cell | r1 | r2 | r3 | r4 (hog) |
|---|---|---|---|---|
| bytes.new:len:i32_max | DIVERGE | DIVERGE | DIVERGE | DIVERGE |
| bytes.pad_left:target_len:i32_max | DIVERGE | DIVERGE | DIVERGE | DIVERGE |
| bytes.pad_right:target_len:i32_max | DIVERGE | DIVERGE | DIVERGE | DIVERGE |
| string.pad_start:n:i32_max | DIVERGE | DIVERGE | DIVERGE | DIVERGE |
| string.pad_end:n:i32_max | DIVERGE | DIVERGE | DIVERGE | DIVERGE |
| bytes.repeat, 15 array readers :i32_max | DIVERGE ×4 (unchanged) |
| string.repeat:n:i32_max | AGREE ×4 (REPEAT ceiling, its own line) |

Zero of 272 cells moved across the four runs. On the merged tree (after #2432's helper exclusion, 125 public bytes cells) the capped `--edge i32_max` sweep reads the same 19 DIVERGE, and `--skips-only --only json` + `--measured` (the shape of `tests/domain_edge_skip_ledger_test.rs`) passes. `check-gate-verification.sh` is green.


O6lvl4:

PR #2442, opened 2026-09-21
Refuse to enqueue a PR whose non-required gate failed

O6lvl4:
Follow-up to #2416 (stacked on it; merge after it lands). The script judged only the 17 branch-protection contexts, so a PR whose non-required gate was red — measured tonight on #2433, where "Coq proofs + axiom audit + PCC gate" failed on the walled-real ratchet while every required context was green — would have been enqueued, merged, and turned develop's non-required jobs red. It now refuses when ANY check on the PR reports FAILURE or TIMED_OUT, naming it "(not required)" so the reader knows which set tripped; pending non-required checks still do not block (the required set decides readiness, the failed set decides refusal).

Pre-push hook skipped (LEFTHOOK=0): one script; the locale-pin test was run and passes.


O6lvl4:
Rebased onto develop (the base commits landed through the queue under new shas); only the two refusal commits remain. Pushed with LEFTHOOK=0: script only.

O6lvl4:

